PR #383, opened 2023-04-13
feat: Call permissions and capabilities

oliverlaz:
### Overview

Implements the missing functionalities around call permissions and user capabilities.

#### Added:
- New APIs and React Hooks which allow easy access to user's permissions and capabilities
- New UI components for requesting and approving permissions

#### Changed:
- Own capabilities shall be updated in and read from `call.state.metadata`
- Events: a few of the default event handlers are now attached to the appropriate call instance. This allows cleaner code with fewer side effects in their implementation. The remaining event handlers will be tackled in a follow-up PR.
- Internal: The coordinator client also accepts generic type parameters for request data.
- OpenAPI code-gen: Enums are now generated in UPPER_CASE notation (SFU enum notation alignment). The schema validation is enabled again.


packages/react-dogfood/components/Stage.tsx
```diff
@@ -9,23 +9,17 @@ export const Stage = ({
...
-    );
-  }
-
   return (
-    // @ts-expect-error
-    <SelectedComponent
-      groupSize={!groupSize || groupSize > 16 ? undefined : groupSize}
-    />
+    <div className="str-video__stage">
+      <PermissionRequests />
```
arnautov-anton, line 18: Let's move this to MeetingUI above Stage instead of having it within Stage directly. We should also rename Stage to something like `ActiveCallView` or something (but that's for another PR).

packages/client/src/rtc/Call.ts
```diff
@@ -1170,12 +1255,15 @@ export class Call {
...
     };
-    return this.streamClient.patch<UpdateCallResponse>(
+    return this.streamClient.patch<UpdateCallResponse, UpdateCallRequest>(
       `${this.streamClientBasePath}`,
       payload,
     );
   };
 
+  /**
+   * Ends the call.
```
MartinCupela, line 1265:
```suggestion
   * Ends the call. Once the call is ended, it cannot be re-joined.
```

packages/react-dogfood/components/Stage.tsx
oliverlaz: done.

arnautov-anton (approved):

PR #386, opened 2023-04-14
feat(react-native): keep call alive in the background

santhoshvai:
This PR contains guide and example implementation in the dogfood app to keep call alive in the background.


packages/react-native-dogfood/src/screens/Call/CallScreen.tsx
```diff
@@ -33,6 +37,16 @@ export const CallScreen = ({ navigation }: Props) => {
...
 
+  useEffect(() => {
+    if (!activeCall) {
+      return;
+    }
+    startForegroundService();
+    return () => {
+      stopForegroundService();
+    };
+  }, [activeCall]);
```
vanGalilea, line 48: These `useEffect` should become a hook to reduce duplication and to be used by our future integrators.

Something like `useForegroundService` or `useKeepCallAlive`

packages/react-native-dogfood/src/screens/Meeting/MeetingScreen.tsx
```diff
@@ -12,6 +16,18 @@ export const MeetingScreen = ({ navigation }: Props) => {
     navigation.navigate('CallParticipantsInfoScreen');
   };
 
+  const activeCall = useActiveCall();
+
+  useEffect(() => {
```
vanGalilea, line 21: > These `useEffect` should become a hook to reduce duplication and to be used by our future integrators.
> 
> Something like `useForegroundService` or `useKeepCallAlive`

To be used here

packages/react-native-sdk/docusaurus/docs/reactnative/02-guides/12-running-calls-in-background.mdx
```diff
@@ -3,4 +3,170 @@ id: 12-running-calls-in-background
...
+Add the following entry to your `AndroidManifest.xml`
+
+```xml
+    <service
+      android:name="app.notifee.core.ForegroundService"
+      android:foregroundServiceType="camera|microphone" />
+```
+
+Add the following function in your codebase to start a foreground service and to keep the call alive.
+
```
vanGalilea, line 57: I find that lib. Specific installation steps should not be mentioned in our docs, yet we should send our integrators via a link to the relevant section in Notiffee's docs.

Otherwise, it's just to difficult to maintain.

vanGalilea:
Have a look at vale vocab please ;)

vanGalilea:
⚠️  I have discovered a **severe** bug ⚠️
- open DF app on Android and join a call. 
- join the same call on iOS
- close the Android app fully
- the bug occurs: both video and audio streams originating from the closed Android app will still be available to all other customers in the call.

I expect that when the app closes fully and a call was active, the user will disconnect fully from the call and will not publish he/hers streams anymore.

packages/react-native-sdk/docusaurus/docs/reactnative/02-guides/12-running-calls-in-background.mdx
santhoshvai: this is specific to our usecase unfortunately 

```
android:foregroundServiceType="camera|microphone"
```

packages/react-native-sdk/src/components/ActiveCall.tsx
```diff
@@ -32,8 +32,17 @@ export interface ActiveCallProps {
...
-  if (!activeCall) return null;
+  const activeCallRef = useRef(activeCall);
+  activeCallRef.current = activeCall;
+
+  useEffect(() => {
+    return () => {
+      // ensure that if this component is unmounted, the call is left.
+      activeCallRef.current?.leave();
+    };
+  }, []);
```
vanGalilea, line 43: An important step that must be incorporated in our `useForgroundService` hook

vanGalilea (approved):

packages/react-native-sdk/src/components/ActiveCall.tsx
vanGalilea: I thought about incorporating this after joining/leaving a call, explicitly calling it vs. a side effect approach 🤔

packages/react-native-dogfood/src/screens/Call/CallScreen.tsx
vanGalilea: Let's do this in the [useCallCycleEvents](https://github.com/GetStream/stream-video-js/blob/f11633d0a0f859ec579ffa670dbd2a8ba5227a44/packages/react-native-sdk/src/hooks/useCallCycleEffect.tsx#L23)

PR #395, opened 2023-04-17
feat: enable codec preferring through sdp munging

santhoshvai:
In react-native, we dont have trasceiver.setCodecPreferences support. So we have to do sdp-munging. 

But this support can be added in the future, upon which this should be removed. currently, its unknown when, so we had to add sdp-munging. 

## Some client JS changes related to web

* If red is to be disabled, we remove only RED codec from the codec preferences 
* If Dtx is disabled, we set it to `usedtx=0` explicitly
* removed the support for setting codecs through the web query params


oliverlaz (approved):

vanGalilea:
This PR in comparison with main breaks some of the func.

All video streams are visible on iOS. 
Yet video streams originating from an iOS devices are not displayed on Android (tested on physical devices with android 11 and 13)

 
Screenshot_2023-04-19-10-50-38-27_54cf48aba4dd0a548f7457bfffe8f31d

vanGalilea (changes requested):
iOS video streams are not visible on Android devices after munging, see screenshot

vanGalilea:
iOS video tracks are also not visible on web. so both not on web and android

santhoshvai:
> iOS video streams are not visible on Android devices after munging, see screenshot

This should be fixed now

santhoshvai:
Note: blank streams in some Android 33 devices is still not resolved in this PR

packages/client/src/rtc/publisher.ts
```diff
@@ -244,12 +259,19 @@ export class Publisher {
...
+    if (sdp) {
+      if (this.isDtxEnabled) {
+        sdp = enableDtx(sdp);
+      }
+      if (isReactNative()) {
+        if (this.reactNativePlatform === 'android') {
+          sdp = setPreferredCodec(sdp, 'video', 'vp8');
+        }
+        sdp = setPreferredCodec(sdp, 'audio', this.preferredAudioCodec);
+      }
```
vanGalilea, line 272: This

vanGalilea, line 272: This

vanGalilea: @santhoshvai and @oliverlaz let's make it more straightforward to our integrators and introduce `preferredVideoCodec` over `reactNativePlatform`

vanGalilea: i.e. 
```
const videoClient = useCreateStreamVideoClient({
    user: USER,
    tokenOrProvider: USER?.custom?.token,
    apiKey: STREAM_API_KEY,
   preferredVideoCodec: Platform.os === "android" && "vp8" 
});
```

oliverlaz: I think `preferredVideo/AudioCodec` was introduced for "test" purposes. Ideally, the clients should negotiate which codec to use.

In this particular example, Android would announce it only supports VP8. All other clients should switch to VP8 as part of the negotiation process.
cc: @thesyncim @su225 @szuperaz

santhoshvai: will refactor this 👍

packages/client/src/rtc/publisher.ts
```diff
@@ -244,12 +259,19 @@ export class Publisher {
...
+    if (sdp) {
+      if (this.isDtxEnabled) {
+        sdp = enableDtx(sdp);
+      }
+      if (isReactNative()) {
+        if (this.preferredVideoCodec) {
+          sdp = setPreferredCodec(sdp, 'video', this.preferredVideoCodec);
+        }
+        sdp = setPreferredCodec(sdp, 'audio', this.preferredAudioCodec);
+      }
```
vanGalilea, line 272: is this still needed?

vanGalilea, line 272: @oliverlaz do you want to turn on munging for other browsers besides chrome too?

vanGalilea (approved):

packages/client/src/rtc/publisher.ts
vanGalilea, line 272: This

packages/client/src/rtc/publisher.ts
```diff
@@ -244,12 +259,19 @@ export class Publisher {
...
-      offer.sdp = offer.sdp.replace(
-        'useinbandfec=1',
-        'useinbandfec=1;usedtx=1',
-      );
+    let sdp = offer.sdp;
+    if (sdp) {
+      if (this.isDtxEnabled) {
+        sdp = enableDtx(sdp);
+      }
+      if (isReactNative()) {
```
vanGalilea, line 267: here

packages/client/src/helpers/sdp-munging/setPreferredCodec.ts
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Returns an SDP with the preferred codec in front of the other codecs.
+ * Example: Suppose we want to prefer VP8
+ * 1. look for specified codec (VP8)  a=rtpmap:96 VP8/90000
+ * 2. extract 96 as an identifier of VP8
+ * 3. find video media specification m=video 9 UDP/TLS/RTP/SAVPF 100 101 96 97 35 36 102 125 127
+ * 4. move 96 to the front
+ * 5. now media looks like this: m=video 9 UDP/TLS/RTP/SAVPF 96 100 101 97 35 36 102 125 127
+ */
+export const setPreferredCodec = (
```
tbarbugli, line 58: do we have tests for this function?

packages/client/src/helpers/sdp-munging/setPreferredCodec.ts
```diff
@@ -0,0 +1,76 @@
...
+) => {
+  const mediaSection = getMediaSection(sdp, mediaType);
+  if (!mediaSection) {
+    return sdp;
+  }
+  const codecId = extractIdOfRtpMap(sdp, preferredCodec);
+  if (!codecId) {
+    return sdp;
+  }
+  const newCodecOrder = moveCodecToFront(mediaSection.codecOrder, codecId);
```
szuperaz, line 71: Sorry for the late comment but in case of audio codecs if the preferred codec is Opus we have to remove RED from the SDP altogether, see [web implementation for reference](https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/rtc/publisher.ts#L277), I might be missing something but I don't see that removal in this implementation

packages/client/src/helpers/sdp-munging/setPreferredCodec.ts
santhoshvai: I didnt write tests here because this is supposed to be temporary until react-native webrtc library adds support for transceiver.setCodecPreferences. But as of now we dont have any news of when the support will come. So I am going to add tests now. Consulted @vanGalilea on this. Thanks @tbarbugli

packages/client/src/helpers/sdp-munging/setPreferredCodec.ts
santhoshvai: Thanks that was an important find. I did not know this.

packages/react-native-sdk/docusaurus/docs/reactnative/06-ui-cookbook/05-participant-list.mdx
```diff
@@ -64,6 +65,9 @@ export default () => {
     user: USER,
     tokenOrProvider: USER.custom.token,
     apiKey: STREAM_API_KEY,
+    options: {
+      preferredVideoCodec: Platform.OS === 'android' ? 'VP8' : undefined,
```
vanGalilea, line 69: Can you add a comment why we do this above this line?

packages/react-native-call-starter-kit/src/components/MessengerWrapper.tsx
```diff
@@ -48,6 +49,9 @@ export const VideoWrapper = ({children}: PropsWithChildren<{}>) => {
     user,
     tokenOrProvider: token,
     apiKey: STREAM_API_KEY,
+    options: {
+      preferredVideoCodec: Platform.OS === 'android' ? 'VP8' : undefined,
```
vanGalilea, line 53: Do we want to do this everywhere?

szuperaz (approved):

PR #403, opened 2023-04-17
feat(react-native): add default sorting mechanism

santhoshvai:


packages/react-native-sdk/src/components/ActiveCall.tsx
```diff
@@ -21,6 +22,10 @@ export interface ActiveCallProps {
    * Handler called when the participants info button is pressed in the active call screen.
    */
   onOpenCallParticipantsInfoView: () => void;
+  /**
+   * When true, the default participant list sorter will be disabled.
+   */
+  disableDefaultSorter?: boolean;
```
khushal87, line 28: Rather than `disableDefaultSorter`, can we name the prop as `enableDefaultSorter` and keep the default value as `true`? This will make the code a bit more readable, IMO. Feel free to ignore my suggestion if it doesn't make sense.

packages/react-native-sdk/src/components/ActiveCall.tsx
```diff
@@ -78,6 +83,7 @@ const InnerActiveCall = (props: ActiveCallProps) => {
       <View onLayout={onLayout} style={styles.callControlsWrapper}>
         <CallControlsView onHangupCall={onHangupCall} />
       </View>
+      {!disableDefaultSorter && <ActiveCallDefaultSorter />}
```
khushal87, line 86: If a participant doesn't want to opt for default sorting, are we asking them to design the entire `ActiveCall` component? 🤔

packages/react-native-sdk/src/components/ActiveCallDefaultSorter.tsx
khushal87, line 1: This looks more like a utility than a component. Correct me if I am wrong. 🤔

packages/react-native-sdk/src/components/ActiveCall.tsx
santhoshvai: This is sort of a normal RN way... 

example: https://reactnative.dev/docs/pressable#disabled 

let me know if it makes sense

packages/react-native-sdk/src/components/ActiveCallDefaultSorter.tsx
santhoshvai: Its called as "headless" component. A way of moving state locally 

In this case the benefit is to move `useRemoteParticipants` hook locally to this component so it doesn't trigger a state change in the parent. 

Let me if there is a better way

packages/react-native-sdk/src/components/ActiveCall.tsx
vanGalilea: It's either disable or enable, both are fine. Yet I feel like It makes sense to me to do disable a default functionality over enabling it, indeed @santhoshvai

packages/react-native-sdk/src/components/ActiveCallDefaultSorter.tsx
vanGalilea: I find it very, very difficult to agree with this approach.
My approach is:  render components that are rendering RN components (might be visible or invisible at times)

packages/react-native-sdk/src/components/CallParticipantsList.tsx
```diff
@@ -57,26 +60,33 @@ export const CallParticipantsList = (props: CallParticipantsListProps) => {
...
+          : VisibilityState.INVISIBLE,
+      };
       if (viewToken.isViewable) {
         viewableParticipantSessionIds.current.add(viewToken.key);
+      } else {
+        viewableParticipantSessionIds.current.delete(viewToken.key);
       }
     });
-    if (viewableItems.length) {
+    if (changed.length) {
```
vanGalilea, line 87: 👍🏻

vanGalilea, line 87: Not using `viewableItems` causes participants video not to be rendered when available and fully visible.
This happens when a participant joins a call, see screenshot

packages/react-native-sdk/src/components/ActiveCallDefaultSorter.tsx
vanGalilea: What are our alternative ways to achieve the same behavior?

santhoshvai: Please suggest an alternative if you dont agree, team.

vanGalilea: We can have a hook for this:
export const useDefaultSorting = (disableSorting: boolean = false) => {
  const remoteParticipants = useRemoteParticipants();
  const isOneToOneCall = remoteParticipants.length === 1;
  const activeCall = useActiveCall();
  useEffect(() => {
    if (!activeCall || disableSorting) return;
    const sortingPreset = getDefaultSortingPreset(isOneToOneCall);
    activeCall.setSortParticipantsBy(sortingPreset);
  }, [activeCall, isOneToOneCall]);
};
I think this would be easier to be understood and explained

santhoshvai: I removed this whole default sorting stuff in our SDK.. due to presence of default sorter in the client JS SDK

@vanGalilea btw the downside of hook is that the state is not moved locally..

vanGalilea (approved):

PR #431, opened 2023-04-25
fix: enable source-maps in dev mode

oliverlaz:
### Overview

Enables source maps in Pronto in dev mode.
Until now, we were looking at the transpiled JS whenever we had to debug our code. Hopefully, this change would drastically improve our DX.


MartinCupela (approved):

MartinCupela (commented):
thank you

PR #433, opened 2023-04-26
feat: Provide device info for web SDKs

szuperaz:
## Aim of PR

Provide device + SDK version info to SFU

## Details

Device info only set up for web, RN part is TODO, but that can be done later
SDK info is set up for React and RN, it currently sends a dummy version as we have yet to set up proper versioning

### DIFF
--- a/packages/client/src/rtc/Call.ts
+++ b/packages/client/src/rtc/Call.ts
@@ -2,7 +2,7 @@ import { StreamSfuClient } from '../StreamSfuClient';
 import { createSubscriber } from './subscriber';
 import { Publisher } from './publisher';
 import { getGenericSdp } from './codecs';
-import { TrackType } from '../gen/video/sfu/models/models';
+import { ClientDetails, TrackType } from '../gen/video/sfu/models/models';
 import { registerEventHandlers } from './callEventHandlers';
 import {
   Dispatcher,
@@ -89,6 +89,7 @@ import {
   EventHandler,
   StreamCallEvent,
 } from '../coordinator/connection/types';
+import { UAParser } from 'ua-parser-js';
 
 /**
  * The options to pass to {@link Call} constructor.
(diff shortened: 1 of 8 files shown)


packages/react-sdk/index.ts
```diff
@@ -1,7 +1,17 @@
...
 export * from './src/hooks';
 export * from './src/types';
+
+// TODO: set valid version
+setSdkInfo({
+  type: SfuModels.SdkType.REACT,
+  major: '0',
+  minor: '0',
+  patch: '0',
+});
```
su225, line 17: Any way this can be automated? It is too easy to forget bumping up the version while cutting the release. Like - based on tag or branch name etc?

su225 (commented):
Looks good :)

oliverlaz (approved):

packages/client/src/rtc/Call.ts
```diff
@@ -597,6 +599,28 @@ export class Call {
     });
 
     try {
+      const clientDetails: ClientDetails = {};
+      if (typeof navigator === 'undefined') {
+        // TODO RN
```
vanGalilea, line 604: I'm creating an task for this, we should lean here on our own framework-specific APIs as much as possible i.e. [Platform](https://reactnative.dev/docs/platform-specific-code#platform-module) and [DeviceInfo](https://github.com/react-native-device-info/react-native-device-info)

santhoshvai: btw navigator is not undefined in RN

packages/react-sdk/index.ts
szuperaz: Yes, it'll be, we just haven't yet set it up

packages/client/src/rtc/Call.ts
szuperaz: @santhoshvai can you help me with the adequate checks for this use-case?

szuperaz: I've seen this in another PR: https://github.com/GetStream/stream-video-js/pull/395/files#diff-61771ffe92cbe5568b7ceecac9671fc8090fc07df09ad2ff8a511f842a5320d0 - should I be using this?

szuperaz: Fixed, please check if this looks ok

vanGalilea: Correct ✅

vanGalilea (approved):

packages/client/src/rtc/Call.ts
szuperaz: Thanks for checking

santhoshvai: missed reading this @szuperaz glad that you found it!

PR #441, opened 2023-04-26
fix: handle gracefully abrupt track ending

oliverlaz:
### Overview

Any `MediaStreamTrack` we publish can be ended abruptly by an external factor such as permission revoking, detaching a device, etc.
In such cases, our engine should be able to detect this and update our internal state and notify the SFU accordingly, as otherwise, remote call participants might observe a frozen stream.
With this change, these abruptions will be handled as "graceful muting" and remote participants would react to this appropriately (muting audio, rendering avatar instead of video, hiding screen share, etc...).

### DIFF
--- a/packages/client/src/rtc/publisher.ts
+++ b/packages/client/src/rtc/publisher.ts
@@ -12,10 +12,13 @@ import {
   findOptimalVideoLayers,
 } from './videoLayers';
 import { getPreferredCodecs } from './codecs';
+import { trackTypeToParticipantStreamKey } from './helpers/tracks';
+import { CallState } from '../store';
 import { PublishOptions } from './types';
 
 export type PublisherOpts = {
-  rpcClient: StreamSfuClient;
+  sfuClient: StreamSfuClient;
+  state: CallState;
   connectionConfig?: RTCConfiguration;
   isDtxEnabled: boolean;
 };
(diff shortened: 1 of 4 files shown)


packages/client/src/rtc/publisher.ts
```diff
@@ -92,64 +102,69 @@ export class Publisher {
...
-        }
-      }
-
-      if (trackType === TrackType.AUDIO) {
-        let returnOnlyMatched = opts.preferredCodec === 'opus';
-        const codecPreferences = getPreferredCodecs(
-          'audio',
-          opts.preferredCodec!,
-          returnOnlyMatched,
+      if ('setCodecPreferences' in transceiver && codecPreferences) {
```
arnautov-anton, line 143:
```suggestion
      if (Object.hasOwn(transceiver, 'setCodecPreferences') && codecPreferences) {
```

szuperaz (approved):

packages/client/src/rtc/publisher.ts
oliverlaz: `Object.hasOwn` has just been recently introduced in Safari (15.4, which is one year old). Also, this won't work due to some internal differences between `in` and `hasOwn/hasOwnProperty`:

Screenshot 2023-04-27 at 09 35 40

arnautov-anton: Oh, TIL - Transceiver object has `setCodecPreferences` somewhere deeper in the prototype. 👍

PR #445, opened 2023-04-27
feat(react-native): add permissions support

khushal87:


packages/react-native-sdk/src/components/PermissionsNotification.tsx
```diff
@@ -0,0 +1,109 @@
...
+    Alert.alert(messageApproved);
+  }, [messageApproved]);
+
+  const showRevokedNotification = useCallback(() => {
+    Alert.alert(messageRevoked);
+  }, [messageRevoked]);
+
+  const showApprovalRequestNotification = useCallback(() => {
+    Alert.alert(messageAwaitingApproval);
+  }, [messageAwaitingApproval]);
```
vanGalilea, line 61: callbacks unneeded here IMHO

packages/react-native-sdk/src/components/PermissionsNotification.tsx
```diff
@@ -0,0 +1,109 @@
...
+      await call
+        .requestPermissions({ permissions: [permission] })
+        .catch((error) => {
+          if (error instanceof AxiosError) {
+            console.log(
+              'RequestPermissions failed',
+              error.response?.data.message,
+            );
+          }
+        });
```
vanGalilea, line 100: async await with .catch 😢 please use async await with try catch

packages/react-native-sdk/src/components/PermissionsNotification.tsx
```diff
@@ -0,0 +1,109 @@
...
+export const PermissionNotification = (props: PermissionNotificationProps) => {
+  const {
+    permission,
+    messageApproved,
+    messageRevoked,
+    messageAwaitingApproval,
+    children,
+  } = props;
+  const hasPermission = useHasPermissions(permission);
+  const previousHasPermission = useRef(hasPermission);
```
vanGalilea, line 46: we have a usePrevious hook please use it

packages/react-native-sdk/src/components/PermissionsNotification.tsx
```diff
@@ -0,0 +1,109 @@
...
+              error.response?.data.message,
+            );
+          }
+        });
+    }
+  }, [call, permission]);
+
+  if (!hasPermission && !isAwaitingApproval) {
+    return <Pressable onPress={handleRequestPermission}>{children}</Pressable>;
+  }
```
vanGalilea, line 106: This will become a pressable that wraps a pressable :/ I'm confused

packages/react-native-sdk/src/components/PermissionsNotification.tsx
```diff
@@ -0,0 +1,109 @@
...
+  messageRevoked: string;
+
+  /**
+   * The message to display in the notification while
+   * the requested permission is awaiting approval.
+   */
+  messageAwaitingApproval: string;
+}>;
+
+export const PermissionNotification = (props: PermissionNotificationProps) => {
```
vanGalilea, line 37: Very difficult to read component, not sure about it's implem. let's speak about it

vanGalilea (changes requested):

vanGalilea:
* in call participants info screen:
- partic. item should be clickable, not only chevron
- me as the connected user in that list should not be clickable
- other partic.'s as not admin should not be able to click on partic. item
- when not clickable remove chevron

vanGalilea:
- close options modal after selection was made

packages/react-native-sdk/src/components/ActiveCall.tsx
```diff
@@ -73,6 +74,7 @@ const InnerActiveCall = (props: ActiveCallProps) => {
       <View style={styles.iconGroup}>
         <CallParticipantsBadge />
       </View>
+      <PermissionRequests />
```
vanGalilea, line 77: change to hook

packages/react-native-sdk/src/components/PermissionsNotification.tsx
vanGalilea: We discussed this component, and we agreed that this approach is not readable. We will refactor this one and take the 'check before action' approach, I.E 

check whether use may toggle video
if yes continue, if not send request

packages/react-native-sdk/src/components/CallParticipantsInfoView.tsx
```diff
@@ -32,16 +45,16 @@ const CallParticipantInfoItem = (props: CallParticipantInfoViewType) => {
   const isScreenSharing = publishedTracks.includes(
     SfuModels.TrackType.SCREEN_SHARE,
   );
-  const showYouLabel = participant.isLoggedInUser;
```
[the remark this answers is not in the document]
khushal87: No, it doesn't seem to be working anymore.

packages/react-native-sdk/src/components/LobbyView.tsx
```diff
@@ -57,8 +57,12 @@ export const LobbyView = () => {
   const joinCallHandler = async () => {
     try {
       await call?.join({ create: true });
-    } catch (e) {
-      console.log('Error joining call', e);
+    } catch (error) {
+      console.log(error);
+      if (error instanceof AxiosError) {
+        console.log('Error joining call', error);
+        Alert.alert(error.response?.data.message);
```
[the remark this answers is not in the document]
khushal87: The reason being that, when the user was blocked he should not be allowed to join the call with same call id in which he was blocked before, so we should show an alert/error somewhere stating this response from the backend.

packages/react-native-sdk/src/components/PermissionsNotification.tsx
khushal87: Umm, these are a dependency to the useEffect below, so not having a callback actually throws an eslint warning of:
```
The 'showGrantedNotification' function makes the dependencies of useEffect Hook (at line 78) change on every render. Move it inside the useEffect callback. Alternatively, wrap the definition of 'showGrantedNotification' in its own useCallback() Hook.
```

Should I remove the function from the useEffect dependency and add an eslint flag that would stop the warning? @vanGalilea

khushal87: Also, I see no harm in having the callback, since its dependent on prop only.

packages/react-native-sdk/src/components/CallParticipantsOptions.tsx
```diff
@@ -21,6 +26,104 @@ type CallParticipantOptionsType = {
...
+              }
+            : null,
+          participant.publishedTracks.includes(SfuModels.TrackType.AUDIO)
+            ? {
+                title: 'Mute Audio',
+                onPressHandler: muteUserAudio,
+              }
+            : null,
+        ]
+      : [];
```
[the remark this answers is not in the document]
khushal87: Tried improving it by having better variable names and extracting stuff that I could. Please have a look.

packages/react-native-sdk/src/components/PermissionsNotification.tsx
khushal87: Its not actually Pressable on Pressable, the logic works such that the upper pressable will be triggered always when the permissions are lacking for a given call control button.

khushal87: Anyways I have refactored the way it was written, hope it will make sense and be more readable now.

packages/react-native-sdk/src/components/CallParticipantsInfoView.tsx
```diff
@@ -60,11 +73,13 @@ const CallParticipantInfoItem = (props: CallParticipantInfoViewType) => {
           </View>
         )}
         {/* Disablling it until we support permissions */}
```
vanGalilea, line 75:
```suggestion
```

packages/react-native-sdk/src/components/CallParticipantsOptions.tsx
```diff
@@ -21,6 +26,125 @@ type CallParticipantOptionsType = {
...
+  const participantCanPublishAudio = participant.publishedTracks.includes(
+    SfuModels.TrackType.AUDIO,
+  );
+
+  console.log({
+    userHasBlockUserCapability,
+    userHasMuteUsersCapability,
+    userHasUpdateCallPermissionsCapability,
+  });
+
```
vanGalilea, line 82:
```suggestion
```

packages/react-native-sdk/src/components/CallParticipantsOptions.tsx
```diff
@@ -52,10 +176,20 @@ export const CallParticipantOptions = (props: CallParticipantOptionsType) => {
...
               index < options.length - 1 ? styles.borderBottom : null;
+
+            const onPressHandler = () => {
+              option?.onPressHandler();
+              onCloseParticipantOptions();
+            };
+
+            if (!option) {
+              return null;
+            }
```
vanGalilea, line 187:
```suggestion
            if (!option) {
              return null;
            }
            
            const applyBottomPadding =
              index < options.length - 1 ? styles.borderBottom : null;

            const onPressHandler = () => {
              option?.onPressHandler();
              onCloseParticipantOptions();
            };
```

packages/react-native-sdk/src/components/ToggleAudioButton.tsx
```diff
@@ -0,0 +1,100 @@
...
+          await call.requestPermissions({ permissions: [permission] });
+        } catch (error) {
+          if (error instanceof AxiosError) {
+            console.log(
+              'RequestPermissions failed',
+              error.response?.data.message,
+            );
+          }
+        }
+      }
```
vanGalilea, line 50:
```suggestion
      if (!call?.permissionsContext.canRequest(permission)) return null;
      
      setIsAwaitingApproval(true);
      try {
        await call.requestPermissions({ permissions: [permission] });
      } catch (error) {
        if (error instanceof AxiosError) {
          console.log(
            'RequestPermissions failed',
            error.response?.data.message,
          );
        }
      }
```

vanGalilea (approved):

PR #446, opened 2023-04-28
State store docs

szuperaz:


packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
```diff
@@ -3,8 +3,84 @@ id: 02-state-store
...
+    <user>,
+});
+
+const activeCall = // this should change with the refactor
+
+<StreamVideo client="{client}">
+  <StreamCallProvider call="{activeCall}">
+    <!-- Video UI -->
+  </StreamCallProvider>
+</StreamVideo>
```
oliverlaz, line 69:
```suggestion
<StreamVideo client={client}>
  <StreamCallProvider call={activeCall}>
    <!-- Video UI -->
  </StreamCallProvider>
</StreamVideo>
```

oliverlaz (approved):

packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
```diff
@@ -3,8 +3,84 @@ id: 02-state-store
...
-[//]: # 'State store (raw/hooks)'
-[//]: # '- introduction- a central source of truth, core parts working with store, exposed to use by SDKs (React/RN/Angular, or React bindings).'
-[//]: # '- Code examples- what does the integ. gain from code (with hooks/raw)'
+The `StreamVideoClient` instance manages its own state store and each `Call` instance has its own state store as well.
+
+The state stores are updated after each API call/WebSocket event. The reactive state store makes it possible to observe the state from anywhere in your application. You can access the state store using hooks. This guide will help you understand how state management work which is useful if you want to attach custom logic upon state changes or if you want to create custom components.
+
+## Create client
+
+Before we can explore the state management mechanism we need to have a `StreamVideoClient` instance, to create it we can use the [`useCreateStreamVideoClient`](./hooks-and-contexts.md#usecreatestreamvideoclient) hook.
```
MartinCupela, line 14:
```suggestion
Before we can explore the state management mechanism we need to have a `StreamVideoClient` instance. To create it we can use the [`useCreateStreamVideoClient`](./hooks-and-contexts.md#usecreatestreamvideoclient) hook.
```

packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
```diff
@@ -3,8 +3,84 @@ id: 02-state-store
...
+const client = useCreateStreamVideoClient({
+    <your API key>,
+    <user token>,
+    <user>,
+});
+```
+
+## Observe client state
+
+To observe client state you need to use the [`StreamVideo`](./hooks-and-contexts.md#streamvideo) context provider:
```
MartinCupela, line 26:
```suggestion
To observe the client state you need to use the [`StreamVideo`](./hooks-and-contexts.md#streamvideo) context provider:
```

packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
```diff
@@ -3,8 +3,84 @@ id: 02-state-store
 title: State Store
 ---
 
-Stay tuned for more information on this topic.
+Our SDK uses a design pattern that we refer to as a reactive state store.
 
-[//]: # 'State store (raw/hooks)'
-[//]: # '- introduction- a central source of truth, core parts working with store, exposed to use by SDKs (React/RN/Angular, or React bindings).'
-[//]: # '- Code examples- what does the integ. gain from code (with hooks/raw)'
+The `StreamVideoClient` instance manages its own state store and each `Call` instance has its own state store as well.
```
MartinCupela, line 8: I propose to mention, what state the client holds compared to the that of Call. It is not clearly distinguished in this article.

packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
```diff
@@ -3,8 +3,84 @@ id: 02-state-store
...
+});
+
+<StreamVideo client="{client}">
+    <!-- Video UI -->
+</StreamVideo>
+```
+
+Then you'll be able to use the [client state hooks](./hooks-and-contexts.md#client-state) to easily access the client state anywhere in your application. These hooks are updated when a relevant client state change occurs.
+
+Let's see an example:
```
MartinCupela, line 42: Missing example?

packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
szuperaz: Do you have a suggestion for this? My idea was that integrators will know what state means by looking at the list of hooks that we reference below

packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
szuperaz: The example is below

packages/client/docusaurus/docs/client/04-call-engine/02-state-store.mdx
MartinCupela: What could help here is to formulate what was our aim, why have we done things in a certain way. So we can answer the reader question "Why have we separated client and call states?". Can we say, what is the general purpose of the client state vs that of the call?

I have a feeling that if we force the reader to look at the reference list, we force them to perform this mental exercise similar to reading the source code. And then they have to make some synthesis on their own. Maybe we can provide them some anchor upfront.

szuperaz: > Why have we separated client and call states?

The client and the call are two separate entities, so I think it's expected that they have separate states

> what is the general purpose of the client state vs that of the call

I've added examples to show what kind of data are stored in each store

PR #450, opened 2023-04-28
feat(core): replace useActiveCall() with useCall()

oliverlaz:
### Overview

This PR removes the existing `useActiveCall()` in favor of `useCall()`.
As part of this PR, both `useIncomingCalls()` and `useOutgoingCalls()` are deprecated and will be removed later.

`useActiveCall()` doesn't work well with the existing SDK architecture as there can currently be multiple calls active at once (joined, watched, etc.), which often leads to edge cases and unambiguity, especially in ring-call scenarios.
The goal of this PR is to have a more streamlined API that works identically in all scenarios.

### TODOs:
- [X] Client package updated (event handlers, Call State, Client State)
- [X] Pronto works
- [X] Messenger Clone (ringing flow) works
- [X] Egress-Composite works
- [X] React Tutorial works
- [x] React Video Demo works
- [x] Zoom Clone (meeting) works
- [x] React Native sample apps work


sample-apps/react/react-video-demo/package.json
```diff
@@ -5,6 +5,7 @@
   "type": "module",
   "scripts": {
     "dev": "vite",
+    "start": "vite",
```
zwaardje, line 8: I take it start is the common one used in stream?

zwaardje (approved):

packages/client/src/events/call.ts
```diff
@@ -1,123 +1,44 @@
...
- * Event handler that watches delivery of CallRejected Websocket event.
- * Updates the state store and notifies its subscribers that
- * the given user will not be joining the call.
+ * Event handler that watches delivery of `call.rejected` Websocket event.
+ * Once the event is received, the call is left.
  */
-export const watchCallRejected = (store: StreamVideoWriteableStateStore) => {
-  return function onCallRejected(event: StreamVideoEvent) {
+export const watchCallRejected = (call: Call) => {
+  let totalRejections = 0;
```
szuperaz, line 24: I think on the long run we could handle this better, I've created a Notion task with my suggestions: https://www.notion.so/stream-wiki/8e8d9743b99f4379b24bd2415e5bfdbd?v=714cda30bb2d433fa7151a33556811c5&p=f2994a2225ea48868f69e8167356092f&pm=c - but i think it's fine for now

szuperaz (approved):

packages/client/src/events/call.ts
oliverlaz: We can definitely do this better... counting the rejections was the simplest solution now.
I like your proposal, we can implement it next/later 👍

sample-apps/react/react-video-demo/package.json
oliverlaz: Yes, typically all of our packages have a default `start` script.

packages/client/src/events/call.ts
```diff
@@ -1,123 +1,46 @@
...
-    const { call_cid } = event;
-    const acceptedIncomingCall = store.incomingCalls.find(
-      (incomingCall) => incomingCall.cid === call_cid,
-    );
-    if (acceptedIncomingCall) {
-      console.warn('Received CallAcceptedEvent for an incoming call');
-      return;
+    const { state } = call;
+    if (state.callingState === CallingState.RINGING) {
+      await call.join();
```
MartinCupela, line 14: @oliverlaz `Call.join` logs error in this case, but it seems a valid scenario. Shouldn't the condition above check, whether the call is already joined / joining?

packages/react-sdk/src/core/contexts/MediaDevicesContext.tsx
```diff
@@ -294,10 +311,10 @@ export const MediaDevicesProvider = ({
     return () => subscription.unsubscribe();
   }, [videoDevices.length]);
 
-  // useEffect(() => {
-  //   if (!call) return;
-  //   call.setAudioOutputDevice(selectedAudioOutputDeviceId);
-  // }, [call, selectedAudioOutputDeviceId]);
+  useEffect(() => {
+    if (!call || callingState !== CallingState.JOINED) return;
+    call.setAudioOutputDevice(selectedAudioOutputDeviceId);
```
MartinCupela, line 316: I think this may break lobby?

oliverlaz: I don't think it does. The device selector there just stores the user's preference. The `VideoPreview` doesn't emit any audio.

packages/client/src/events/call.ts
oliverlaz: This will execute only if the call hasn't been already accepted.

vanGalilea (approved):

PR #454, opened 2023-05-01
Voip cert

tbarbugli:


packages/client/src/gen/coordinator/index.ts
```diff
@@ -2525,7 +2493,7 @@ export interface ReactionResponse {
    * @type {{ [key: string]: any; }}
    * @memberof ReactionResponse
    */
-  custom?: { [key: string]: any };
+  custom: { [key: string]: any };
```
oliverlaz, line 2496:
```suggestion
  custom?: { [key: string]: any };
```

The newest schema marks this field as optional.

packages/client/src/gen/coordinator/index.ts
```diff
@@ -2432,7 +2400,7 @@ export interface QueryMembersRequest {
    * @type {{ [key: string]: any; }}
    * @memberof QueryMembersRequest
    */
-  filter_conditions?: { [key: string]: any };
+  filter_conditions: { [key: string]: any };
```
oliverlaz, line 2403:
```suggestion
  filter_conditions?: { [key: string]: any };
```

The newest schema marks this field as optional.

PR #465, opened 2023-05-04
fix(lobby): hide video/audio preview screen for anonymous users

oliverlaz:


szuperaz (approved):
Do we need this because

PR #467, opened 2023-05-05
feat(latency): breadth-first time-boxed latency measurement

oliverlaz:
### Overview

An attempt to make the "measure latencies" step more predictable and reliable.
This new flow will run all latency measurements in parallel, for all provided edges. Every measurement attempt is time-boxed to 1s and the whole measuring process can't take longer than 1.2 seconds.

The coordinator might receive none, or incomplete measurements as a side-effect of this new flow.

### DIFF
--- a/packages/client/src/rtc/flows/latency.ts
+++ b/packages/client/src/rtc/flows/latency.ts
@@ -1,43 +1,91 @@
+import {
+  DatacenterResponse,
+  GetCallEdgeServerRequest,
+} from '../../gen/coordinator';
+
 const toSeconds = (ms: number) => ms / 1000;
 
 /**
  * Measures the latency of the current client to the given endpoint.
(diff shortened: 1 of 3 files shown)


packages/client/src/rtc/flows/latency.ts
```diff
@@ -1,43 +1,91 @@
...
+  const controller = new AbortController();
+  const abortTimeout = setTimeout(() => {
+    controller.abort();
+  }, timeoutAfterMs);
+  try {
+    const src = new URL(endpoint);
+    src.searchParams.set('r', `js_${Math.random() * 10000000}`);
+    await fetch(src.toString(), {
+      signal: controller.signal,
+    }).then((response) => response.blob());
```
arnautov-anton, line 28:
```suggestion
    });
```

szuperaz (approved):

PR #470, opened 2023-05-08
fix: preserve last used deviceId when muting

oliverlaz:
### Overview

https://github.com/GetStream/stream-video-js/pull/441 introduced a regression in the area of persisting the last used `deviceId` of any media track.
This PR restores the missing functionality and adds a test that verifies the behavior is correct.

On the testing side, I had to mock many of the WebRTC APIs as they aren't natively available in Node nor in the "fake" DOM testing environments like `jsdom` and `happy-dom`. I'd appreciate if you can have a look at this and let me know if you have some better idea how to approach WebRTC unit testing.

### DIFF
--- a/packages/client/src/rtc/__tests__/mocks/webrtc.mocks.ts
+++ b/packages/client/src/rtc/__tests__/mocks/webrtc.mocks.ts
@@ -1,0 +1,42 @@
+import { vi } from 'vitest';
+
+const RTCPeerConnectionMock = vi.fn((): Partial<RTCPeerConnection> => {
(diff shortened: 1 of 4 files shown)


szuperaz (approved):
I don't know a better solution, apart from maybe trying to find a mock lib that comes with the necessary definitions

oliverlaz:
> I don't know a better solution, apart from maybe trying to find a mock lib that comes with the necessary definitions

Unfortunately, I couldn't find anything good that can fit our needs.

PR #472, opened 2023-05-08
feat(react-demo): copy code to s3 bucket with git workflow

zwaardje:
Copy react demo build code to an s3 bucket

### DIFF
--- a/.github/workflows/copy-react-demo-s3.yml
+++ b/.github/workflows/copy-react-demo-s3.yml
@@ -60,9 +60,9 @@ jobs:
      - name: Copy PR build artifacts to AWS S3 Bucket
        if: ${{ github.ref_name != 'main' }}
        run: |
-          aws s3 sync sample-apps/react/react-video-demo/staging/dist s3://${{secrets.S3_BUCKET_NAME_STAGING}}/video/demos --delete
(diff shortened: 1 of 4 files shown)


oliverlaz (approved):

.github/workflows/copy-react-demo-s3.yml
```diff
@@ -0,0 +1,63 @@
...
+     - name: Setup AWS credentials
+       uses: aws-actions/configure-aws-credentials@v2
+       with:
+          aws-access-key-id: ${{ secrets.AWS_ACCESS_KEY_ID }}
+          aws-secret-access-key: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+          aws-region: ${{vars.AWS_REGION}}
+
+     - name: Copy artifacts to AWS S3 Bucket
+       run: |
+          aws s3 sync sample-apps/react/react-video-demo/build s3://${{vars.S3_BUCKET_NAME}}/react-video-demo --delete
```
oliverlaz, line 62: I think you should make two different directories in S3, as from what I see, PR builds and master builds would push code to the same dir in the bucket. This might be a bad idea as broken code can be accidentally pushed to the production site.

zwaardje: Ah yes, thank you makes sense. I will change that. I might even remove PR builds

PR #474, opened 2023-05-08
fix: in call manager bluetooth state patch for android

santhoshvai:


vanGalilea:
Testing results:
- BT devices: AirPods and Sony WH-1000XM4 🎧 
- iOS(iPhone SE [16.1.1]: No regression ✅ 
- Android (OnepPlus [33] and Mi2 [31]:
  - sound is being routed via BT headset when connected before or during call ✅ 
  - when BT device is not connected or after being disconnected, audio should be routed via speaker and not earpiece ❌
 - JS: when a device is being disconnected, `Sending 'WiredHeadset with no listeners registered` warning occurs see screenshot.
 
File

vanGalilea (approved):

packages/react-native-sdk/src/components/ActiveCall.tsx
```diff
@@ -109,3 +111,33 @@ const styles = StyleSheet.create({
   callParticipantsWrapper: { flex: 1 },
   callControlsWrapper: { position: 'absolute', bottom: 0, left: 0, right: 0 },
 });
+
+const verifyAndroidPermissions = async () => {
```
vanGalilea, line 115: Let's encapsulate in a util hook or anywhere else ;)

PR #497, opened 2023-05-15
docs: Docs suggestions

szuperaz:


DaemonLoki (approved):
All good points, and perfect to keep it in mind with the notes you added. Thanks a lot @szuperaz !

PR #499, opened 2023-05-15
feat: faster join flow

oliverlaz:
### Overview

Speeds up the Join Flow by eliminating unnecessary REST API calls and latency measurements. The user's approximate location is now retrieved by peeking into the AWS Cloudfront Edge network (aws-pop).


packages/client/src/Call.ts
```diff
@@ -447,6 +447,9 @@ export class Call {
     this.state.setMetadata(response.call);
     this.state.setMembers(response.members);
 
+    this.watching = true;
```
szuperaz, line 450: We should add `await httpClient.connectionIdPromise;`, and if we add watching to the `get` call we should also add the same logic to the `gotOrCreate` method

PR #512, opened 2023-05-17
refactor(react-sdk): enumerate devices when components that need them are rendered

MartinCupela:
## Refactor
1. Move the device enumeration effects into hooks that are invoked in components that actually need access to  up-to-date array of devices of a specific kind. Remove enumerate prop from `MediaDeviceContextProps`.
2. Rename `MediaDeviceContext` API methods that toggle initial device state to reflect they are touching only initial device state.

## Bug fix

1. When publishing audio / video as a result of click on toggle buttons, set the initial video/audio state to enabled. This will then allow republishing streams, when selected device id changes. 
2. Allow ParticipantView to shrink if the container is smaller


packages/styling/src/ParticipantView/ParticipantView-layout.scss
```diff
@@ -6,6 +6,7 @@ $scope-name: 'str-video__participant-view';
   width: 100%;
   max-width: 1280px;
   max-height: 100%;
+  min-height: 0;
```
arnautov-anton, line 9: Is this a change for a specific layout?

MartinCupela: This is reaction to the component overflow in messenger clone.

packages/react-sdk/src/core/hooks/useObserveDevices.ts
```diff
@@ -0,0 +1,137 @@
...
+import {
+  getAudioDevices,
+  getAudioOutputDevices,
+  getVideoDevices,
+} from '@stream-io/video-client';
+
+/**
+ * Observes changes in connected devices and maintains an up-to-date array of connected MediaDeviceInfo objects.
+ * @param observeDevices
+ */
```
szuperaz, line 12: I think the `@category Device Management` lines are missing, so the new hooks are under "Device Management" category in the docs: https://staging.getstream.io/video/docs/react/call-engine/hooks-and-contexts/#device-management

packages/react-sdk/src/core/contexts/MediaDevicesContext.tsx
```diff
@@ -95,13 +97,14 @@ export type MediaDevicesContextAPI = {
...
+  setInitialAudioEnabled: (enabled: boolean) => void;
   setInitialVideoState: (state: DeviceState) => void;
   selectedAudioInputDeviceId?: string;
   selectedAudioOutputDeviceId?: string;
   selectedVideoDeviceId?: string;
   switchDevice: (kind: MediaDeviceKind, deviceId?: string) => void;
-  toggleAudioMuteState: () => void;
-  toggleVideoMuteState: () => void;
+  toggleInitialAudioMuteState: () => void;
+  toggleInitialVideoMuteState: () => void;
```
oliverlaz, line 107: These are public api methods. Why do they have an `initial` in their name when they are used for regular toggling?

packages/styling/src/ParticipantView/ParticipantView-layout.scss
MartinCupela: I have removed the rule in CallParticipantsView as you requested :)

packages/react-sdk/src/core/contexts/MediaDevicesContext.tsx
oliverlaz: Ok, I missed that part. I keep forgetting that both buttons (preview and in-call) live in the same file in the SDK.

PR #519, opened 2023-05-18
feat: server-side defaults for mic and camera

oliverlaz:
### Overview

Adds support for call-type provided defaults for microphone and camera.
The integrators still have the opportunity to override this behavior client-side.

### DIFF
--- a/packages/react-sdk/src/components/StreamCall/StreamCall.tsx
+++ b/packages/react-sdk/src/components/StreamCall/StreamCall.tsx
@@ -8,15 +8,33 @@ import {
 import { MediaDevicesProvider, MediaDevicesProviderProps } from '../../core';
 
 type InitWithCallCID = {
-  callId: string;
+  /**
+   * The call type.
+   */
   callType: string;
+  /**
+   * The call id.
+   */
+  callId: string;
+  /**
+   * The call instance to use.
+   */
   call?: never;
 };
 
 type InitWithCallInstance = {
+  /**
+   * The call instance to use.
+   */
   call: Call | undefined;
-  callId?: never;
+  /**
+   * The call type.
+   */
   callType?: never;
+  /**
+   * The call id.
+   */
+  callId?: never;
 };
 
 type InitStreamCall = InitWithCallCID | InitWithCallInstance;
@@ -47,7 +65,7 @@ export type StreamCallProps = InitStreamCall & {
    * </StreamCall>
    * ```
    *
-   * This property is ignored if you pass the `call` prop.
+   * This property is ignored if you pass the `call` prop or enable `autoJoin`.
    *
    * @default true.
    */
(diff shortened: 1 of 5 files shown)


packages/react-sdk/src/components/StreamCall/StreamCall.tsx
```diff
@@ -66,6 +90,18 @@ export const StreamCall = ({
     }
   }, [activeCall, callId, callType, client]);
 
+  const connectedUser = useConnectedUser();
+  useEffect(() => {
+    if (!connectedUser) return;
+    // run the effect only when the user is connected and the call
+    // is created declarative by using the `callId` and `callType` props.
+    if (activeCall && callType && callId && autoLoad) {
+      activeCall.getOrCreate(data).catch((err) => {
```
szuperaz, line 99: We should look at the `create` flag in data, and call `get` or `getOrCreate` based on it's value

packages/react-sdk/src/components/StreamCall/StreamCall.tsx
```diff
@@ -30,6 +31,28 @@ export type StreamCallProps = InitStreamCall & {
    */
   autoJoin?: boolean;
 
+  /**
+   * If true, the call will be loaded automatically.
+   *
+   * This property is useful for the scenarios where you declarative create
```
MartinCupela, line 37:
```suggestion
   * This property is useful for the scenarios where you declaratively create
```

packages/react-sdk/src/components/StreamCall/StreamCall.tsx
oliverlaz: Sure, good point!

packages/react-sdk/src/components/StreamCall/StreamCall.tsx
```diff
@@ -66,6 +90,18 @@ export const StreamCall = ({
     }
   }, [activeCall, callId, callType, client]);
 
+  const connectedUser = useConnectedUser();
+  useEffect(() => {
+    if (!connectedUser) return;
+    // run the effect only when the user is connected and the call
+    // is created declarative by using the `callId` and `callType` props.
+    if (activeCall && callType && callId && autoLoad) {
```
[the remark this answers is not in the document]
oliverlaz: yes, `getOrCreate` is redundant in this case. I'll update.

szuperaz (approved):

packages/react-sdk/src/components/StreamCall/StreamCall.tsx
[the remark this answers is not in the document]
oliverlaz: updated.

packages/react-sdk/src/components/StreamCall/StreamCall.tsx
oliverlaz: We have decided to stay with `getOrCreate`. Reason:
- `call.get()` would fail with HTTP 404 because new calls won't exist in the system and we expect this to be the majority of the scenarios.

PR #525, opened 2023-05-19
feat: floating local video view is draggable

vanGalilea:
The goal is to have the local video floating view draggable to view partic. when needed.

https://github.com/GetStream/stream-video-js/assets/25864161/ab8cb297-4dd9-4301-bd41-6100f6debfe8

Limitations: no snapping to top right/left, bottom right/left corners

### DIFF
--- a/packages/react-native-sdk/src/components/LocalVideoView.tsx
+++ b/packages/react-native-sdk/src/components/LocalVideoView.tsx
@@ -1,5 +1,12 @@
-import React from 'react';
-import { StyleProp, StyleSheet, View, ViewStyle } from 'react-native';
+import React, { useRef } from 'react';
+import {
+  Animated,
+  PanResponder,
+  StyleProp,
+  StyleSheet,
+  View,
+  ViewStyle,
+} from 'react-native';
 import { VideoRenderer } from './VideoRenderer';
 import { useLocalParticipant } from '@stream-io/video-react-bindings';
 import { SfuModels } from '@stream-io/video-client';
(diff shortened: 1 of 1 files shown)


packages/react-native-sdk/src/components/LocalVideoView.tsx
```diff
@@ -56,29 +63,59 @@ export const LocalVideoView = (props: LocalVideoViewProps) => {
   const isCameraOnFrontFacingMode = useStreamVideoStoreValue(
     (store) => store.isCameraOnFrontFacingMode,
   );
+  const pan = useRef(new Animated.ValueXY()).current;
+  const panResponder = useRef(
+    PanResponder.create({
+      onMoveShouldSetPanResponder: () => true,
+      onPanResponderMove: Animated.event([null, { dx: pan.x, dy: pan.y }], {
+        useNativeDriver: false,
```
santhoshvai, line 71: why is it false here? is true not supported for this usecase?

false is potentially slow on medium/low end android devices

packages/react-native-sdk/src/components/LocalVideoView.tsx
```diff
@@ -56,29 +63,59 @@ export const LocalVideoView = (props: LocalVideoViewProps) => {
   const isCameraOnFrontFacingMode = useStreamVideoStoreValue(
     (store) => store.isCameraOnFrontFacingMode,
   );
+  const pan = useRef(new Animated.ValueXY()).current;
```
santhoshvai, line 66: some comments of what this variable does and the animation logic would help for maintenance

packages/react-native-sdk/src/components/LocalVideoView.tsx
```diff
@@ -56,29 +63,59 @@ export const LocalVideoView = (props: LocalVideoViewProps) => {
...
+  const panResponder = useRef(
+    PanResponder.create({
+      onMoveShouldSetPanResponder: () => true,
+      onPanResponderMove: Animated.event([null, { dx: pan.x, dy: pan.y }], {
+        useNativeDriver: false,
+      }),
+      onPanResponderRelease: () => {
+        pan.extractOffset();
```
santhoshvai, line 74: what does this do? 

btw with the release, we can snap to positions... Can this be added in this PR?

packages/react-native-sdk/src/components/LocalVideoView.tsx
santhoshvai: translate animations are supported with useNativeDrive afaik

packages/react-native-sdk/src/components/LocalVideoView.tsx
vanGalilea: > Limitations: no snapping to top right/left, bottom right/left corners.

Agree, and I mentioned the limitation indeed, going to timebox the snapping feat.

packages/react-native-sdk/src/components/LocalVideoView.tsx
vanGalilea: they are supported, it just failed due to an unknow reason now it works 👍🏻

vanGalilea: https://github.com/facebook/react-native/issues/13377#issuecomment-292657541

packages/react-native-sdk/src/components/LocalVideoView.tsx
vanGalilea: This will take to much time to figure out atm, going to park this ;)

PR #527, opened 2023-05-22
feat: missing WS events and endpoints

szuperaz:
- Watch for `member_added`, `member_removed`, `member_updated_permission` and `call.member_updated` events
- Watch for `live_started` event, `useIsCallLive` hook added
- `call.ended` event handler fix: watch this for all calls, not just for ring calls
- Send client details in SFU join
- Implement `queryMembers` function -> it doesn't store the result in the state, as I don't know how that data can be merged with the members from join/create response
- Small docs updates

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -707,7 +707,12 @@ export class Call {
             this.streamClient.options.preferredVideoCodec,
           ),
         )
-        .then((sdp) => sfuClient.join({ subscriberSdp: sdp || '' }));
+        .then((sdp) =>
+          sfuClient.join({
+            subscriberSdp: sdp || '',
+            clientDetails,
+          }),
+        );
 
       // 2. in parallel, wait for the SFU to send us the "joinResponse"
       // this will throw an error if the SFU rejects the join request or
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -355,6 +355,7 @@ export class Call {
    * Leave the call and stop the media streams that were published by the call.
    */
   leave = async ({ reject = false }: CallLeaveOptions = {}) => {
+    // TODO: handle case when leave is called during JOINING
     const callingState = this.state.callingState;
     if (callingState === CallingState.LEFT) {
       throw new Error('Cannot leave call that has already been left.');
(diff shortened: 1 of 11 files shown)


packages/client/src/Call.ts
```diff
@@ -1392,6 +1399,19 @@ export class Call {
     });
   };
 
+  /**
+   * Query call members with filter query. The result won't be stored in call state.
+   * @param request
+   * @returns
+   */
+  queryMembers = (request: Omit<QueryMembersRequest, 'type' | 'id'>) => {
```
oliverlaz, line 1407: I wonder why `type` and `id` aren't part of the endpoint path...

packages/client/src/events/backstage.ts
```diff
@@ -0,0 +1,15 @@
+import { StreamVideoEvent } from '../coordinator/connection/types';
+import { CallState } from '../store';
+
+/**
+ * Watches for `call.live_started` events.
+ */
+export const watchCallLiveStarted = (state: CallState) => {
+  return function onCallRecordingStarted(event: StreamVideoEvent) {
```
oliverlaz, line 8:
```suggestion
  return function onCallLiveStarted(event: StreamVideoEvent) {
```

packages/client/src/events/members.ts
```diff
@@ -0,0 +1,62 @@
+import { StreamVideoEvent } from '../coordinator/connection/types';
+import { CallState } from '../store';
+
+/**
+ * Watches for `call.member_added` events.
+ */
+export const watchCallMemberAdded = (state: CallState) => {
+  return function onCallRecordingStarted(event: StreamVideoEvent) {
```
oliverlaz, line 8:
```suggestion
  return function onCallMemberAdded(event: StreamVideoEvent) {
```

oliverlaz (approved):

packages/client/src/Call.ts
szuperaz: I think because that endpoint is already used for call member updates: https://getstream.github.io/protocol/?urls.primaryName=Video#/Video%20Calls/UpdateCallMembers - but maybe that could be a `PATCH` endpoint

szuperaz:
@oliverlaz thanks for spotting the copy-paste mistakes 😄

packages/client/src/Call.ts
oliverlaz: yeah, it could be... but on our side, I think we are covered for any potential future changes because we exclude `id` and `type` from the `request`.

PR #530, opened 2023-05-22
feat(react-native): guest and anonymous login and refactoring of the dogfooding app

khushal87:
This PR targets introducing the Guest and Anonymous login features to the RN dogfooding app. In achieving the same, there are some architectural changes in how the client is being instantiated at the application level and how it is passed to the StreamVideo and StreamCall components. 

Note: The ringing flow is broken due to the changes of the PR. This will be fixed in a separate PR.


packages/react-bindings/src/hooks/client.ts
```diff
@@ -57,9 +57,11 @@ export const useCreateStreamVideoClient = ({
...
           });
       }
-      return client.connectUser(user, tokenOrProvider).catch((err) => {
-        console.error(`Failed to establish connection`, err);
-      });
+      if (user.id && tokenOrProvider) {
+        return client.connectUser(user, tokenOrProvider).catch((err) => {
+          console.error(`Failed to establish connection`, err);
+        });
+      }
```
santhoshvai, line 64: this check seems unnecessary. If missing we can throw an error instead. ore clear to the integrator in that way I think

khushal87: I have removed it in my latest commit.

packages/react-native-dogfood/src/components/LobbyViewComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+        <Pressable
+          style={styles.anonymousButton}
+          onPress={() => {
+            navigation.navigate('GuestModeScreen', { callId });
+          }}
+        >
+          <Text style={styles.anonymousButtonText}>
+            Join as Guest or Anonymously
+          </Text>
+        </Pressable>
```
santhoshvai, line 32: This is can be moved to a separate component and reused.. as the same style and pressable is used twice and only onPress and text changes...

simpler to maintain that way

packages/react-native-dogfood/src/components/LobbyViewComponent.tsx
```diff
@@ -0,0 +1,62 @@
...
+  callId: string;
+};
+
+export const LobbyViewComponent = ({
+  callId,
+  navigation,
+  route,
+}: LobbyViewComponentType) => {
+  return (
+    <View style={[StyleSheet.absoluteFill, styles.container]}>
```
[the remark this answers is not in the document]
khushal87: Not sure, what would be the problem if array is used. I have always seen this array pattern in projects. Can you help?

packages/react-native-dogfood/src/components/MeetingUI.tsx
```diff
@@ -0,0 +1,103 @@
...
+
+  const returnToHomeHandler = () => {
+    navigation.navigate('JoinMeetingScreen');
+  };
+
+  const backToLobbyHandler = () => {
+    setShow('lobby');
+  };
+
+  let ComponentToRender: JSX.Element | null = null;
```
santhoshvai, line 43: instead of let I think it is easier to  use early return

of course optional for you to change. 

Example:

```
else if (show === 'lobby') {
   return (
      <LobbyViewComponent callId={callId} {...{ navigation, route }} />
    );
```

packages/react-native-sdk/src/components/CallParticipantsInfoView.tsx
```diff
@@ -110,38 +124,95 @@ const InnerCallParticipantsInfoView = () => {
...
+            </Restricted>
+          </View>
+          <FlatList
+            data={participants}
+            keyExtractor={(item) => `participant-info-${item.sessionId}`}
+            renderItem={({ item: participant }) => (
+              <CallParticipantInfoItem
+                participant={participant}
+                setSelectedParticipant={setSelectedParticipant}
+              />
```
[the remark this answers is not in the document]
khushal87: Done

packages/react-native-sdk/src/components/LobbyView.tsx
```diff
@@ -81,53 +86,57 @@ export const LobbyView = () => {
...
+            >
+              {VideoIcon}
+            </CallControlsButton>
+          </View>
+        )}
         <View style={styles.info}>
           <Text style={styles.infoText}>
-            You are about to join a test call at Stream. 3 more people are in
-            the call now.
+            You are about to join a call with id {call?.id} at Stream. {count}{' '}
```
santhoshvai, line 136: Is mentioning call id necessary.. for example in google meet only how many are already in the call is shown..

And if count is 0 may be say.. you will be the first to join

packages/react-native-sdk/src/components/Avatar.tsx
```diff
@@ -35,10 +35,12 @@ export const Avatar = (props: AvatarProps) => {
...
   } = props;
-  const label = useMemo(
-    () => getInitialsOfName(name || userId) || '?',
-    [name, userId],
-  );
+
+  const label = useMemo(() => {
+    const userDetails = name || userId;
+    return userDetails ? getInitialsOfName(userDetails) : '?';
+  }, [name, userId]);
```
santhoshvai, line 42: this useMemo seems unnecessary to just compute a string

packages/react-native-sdk/src/components/ActiveCall.tsx
```diff
@@ -40,10 +40,6 @@ export const ActiveCall = (props: ActiveCallProps) => {
   useEffect(() => {
     // when the component mounts, we ask for necessary permissions.
     verifyAndroidBluetoothPermissions();
-    return () => {
-      // ensure that if this component is unmounted, the call is left.
-      activeCallRef.current?.leave();
-    };
   }, []);
```
santhoshvai, line 43: why removed though?

packages/react-native-sdk/src/components/LobbyView.tsx
khushal87: This text is tentative and according to design. I am not sure, why it is so.

packages/react-native-sdk/src/components/CallParticipantsInfoView.tsx
```diff
@@ -110,38 +125,99 @@ const InnerCallParticipantsInfoView = () => {
...
+      transparent={true}
+      visible={isCallParticipantsViewVisible}
+      onRequestClose={onCloseCallParticipantsViewVisible}
+    >
+      <View style={styles.container}>
+        <View style={styles.content}>
+          <View style={styles.header}>
+            <View style={styles.leftHeaderElement} />
+            <Text style={styles.headerText}>
+              Participants ({participants.length})
```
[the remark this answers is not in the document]
khushal87: Figma has it this way

packages/react-native-dogfood/src/screens/Meeting/GuestMeetingScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const guestCallType = 'default';
+
+  const user: UserResponse = {
+    id: `anonymous-${Math.random().toString(36).substring(2, 15)}`,
+    created_at: new Date().toISOString(),
+    updated_at: new Date().toISOString(),
+    role: 'guest',
+    teams: [],
+    custom: {},
+  };
```
santhoshvai, line 39: better to move it inside useState().. having it outside seems unuseful

packages/react-native-dogfood/src/screens/Meeting/GuestMeetingScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+    client
+      .createGuestUser({
+        user: {
+          id: guestUserId,
+          name: guestUserId,
+          role: 'guest',
+        },
+      })
+      .then((guestUser) => {
+        console.log(guestUser);
```
santhoshvai, line 93: unnecessary log

packages/react-native-dogfood/src/screens/Meeting/GuestMeetingScreen.tsx
```diff
@@ -0,0 +1,134 @@
...
+        setUserToConnect(guestUser.user);
+        setTokenToUse(guestUser.access_token);
+        setIsAnonymous(false);
+      })
+      .catch((err) => {
+        console.error('Error creating guest user', err);
+      });
```
santhoshvai, line 100: @vanGalilea asked me in a pr to not use promises

but my view is inside useEffects its fine but not to mix both, up to you

packages/react-native-sdk/src/components/ActiveCall.tsx
santhoshvai: If you do android hardware back or ios default back swipe. does it leave the call? can you check that please

If not, this code will handle that.. may be we can have a condition to check if it is already left and not to do in that case

packages/react-native-sdk/src/components/LobbyView.tsx
santhoshvai: ah okay design seems incomplete then to me..

santhoshvai: upto you to change or not, was a suggestion

packages/react-native-dogfood/src/components/LobbyViewComponent.tsx
santhoshvai: I mean a seperate component in the same file, just for reusing.. DRY principle

packages/react-native-dogfood/src/components/LobbyViewComponent.tsx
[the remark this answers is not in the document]
santhoshvai: a small performance thing that is hardly noticeable here though..

- when you do `[StyleSheet.absoluteFill, styles.container]` you are merging two styles into one always on render..
- when you do object merging in stylesheet.create, it is computed only once when app is initialised

packages/react-native-meeting-starter-kit/src/components/MeetingUI.tsx
```diff
@@ -0,0 +1,88 @@
...
+  };
+
+  const backToLobbyHandler = () => {
+    setShow('lobby');
+  };
+
+  let ComponentToRender: JSX.Element | null = null;
+
+  if (show === 'error-join' || show === 'error-leave') {
+    ComponentToRender = (
```
santhoshvai, line 37: also early return here is better in my view rather than using let

packages/react-native-dogfood/src/screens/Meeting/GuestMeetingScreen.tsx
khushal87: I feel this is clear

packages/react-native-sdk/src/components/CallParticipantsInfoView.tsx
```diff
@@ -110,38 +125,99 @@ const InnerCallParticipantsInfoView = () => {
...
-          />
-        )}
+  const onCloseCallParticipantsViewVisible = () => {
+    setIsCallParticipantsViewVisible(false);
+  };
+
+  const renderItem = useCallback(({ item }: { item: Participant }) => {
+    return (
+      <CallParticipantInfoItem
+        participant={item}
```
santhoshvai, line 135: also I would add key here and remove the keyextractor.. so none of the dependencies except data to flatlist will be changing on every render then

packages/react-native-dogfood/src/screens/Meeting/GuestMeetingScreen.tsx
santhoshvai: imagine someone looking at this code a month after.. they might be wondering why is user created on every render.. but only used at the initial state

santhoshvai: I would have suggested moving outside of the render function... but date needs to created newly at mount, that cannot be done

packages/react-native-sdk/src/components/ActiveCall.tsx
santhoshvai: no that will be buggy as the useEffect dep in calling state

better to get calling state from `activeCallRef.current?.state.callingState`

vanGalilea:
- The close X at the top right corner is not located correctly- see [figma](https://www.figma.com/file/WTMx1povWFb9A4MYwAzR4o/Video-%2F-Voice-Calling-UI?type=design&node-id=1462-92397&t=uyMUBi5336RnBiUo-4)
- After clicking on a certain partic. I get an empty modal with no actions, can we remove it, and remove chevron from partic. item?

Simulator Screen Shot - iPhone 14 - 2023-05-26 at 10 23 48

vanGalilea:
In the guest mode screen:
- can we use this [design](https://file.notion.so/f/s/d4400be2-d692-46a5-9075-2ff0ecf3a863/Screen_Shot_2023-04-07_at_10.39.45_AM.png?id=63aa91d8-273f-468d-9c25-2dfef80378f2&table=block&spaceId=eadf664a-4346-4f42-84bc-741fd336fc09&expirationTimestamp=1685176484033&signature=SIyl2Sm0Nf3PJgF0SzhNztPaHjKn9JWqTZPyXBW4fr8&downloadName=Screen+Shot+2023-04-07+at+10.39.45+AM.png) of the TextInput? (current one is to rounded.

vanGalilea:
As anon/guest you're not allowed to hang up calls, so you're stuck forever. We clearly forgot something there

simulator_screenshot_BAA77EED-283B-43E8-A9D5-DAAE960D6D90

vanGalilea (approved):

packages/react-native-dogfood/App.tsx
```diff
@@ -151,117 +143,108 @@ const Login = () => {
...
+      break;
+  }
+
   useEffect(() => {
     const subscription = prontoCallId$.subscribe((prontoCallId) => {
       if (prontoCallId) {
-        setState({
-          callId: prontoCallId,
-        });
+        meetingNavigation.navigate('MeetingScreen', { callId: prontoCallId });
```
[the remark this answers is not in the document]
santhoshvai: best to move it inside the meeting screen as a state update, and then immediately join as well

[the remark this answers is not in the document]
santhoshvai: also maybe also leave the current call if we are already in a different call

PR #544, opened 2023-05-24
refactor(react-native): use create-token auth endpoint instead of rn-pure-jwt to create token

khushal87:
Use the `https://stream-calls-dogfood.vercel.app/api/auth/create-token` endpoint to create the token and populate it within the dogfooding app. The usage of `react-native-pure-jwt` is removed as a part of the PR.


packages/react-bindings/src/hooks/client.ts
```diff
@@ -57,9 +57,11 @@ export const useCreateStreamVideoClient = ({
...
           });
       }
-      return client.connectUser(user, tokenOrProvider).catch((err) => {
-        console.error(`Failed to establish connection`, err);
-      });
+      if (user.id && tokenOrProvider) {
+        return client.connectUser(user, tokenOrProvider).catch((err) => {
+          console.error(`Failed to establish connection`, err);
+        });
+      }
```
oliverlaz, line 64: Apologies for jumping in this PR but, why are these checks necessary? `user`, `user.id` and `tokenOrProvider` are marked as required.
If missing at runtime, we should throw an error instead of silently doing nothing.

khushal87: Yes, but does it make sense to call connectUser unnecessarily, if the user hasn't provided the token and user details appropriately. `user` is still fine, but there are cases where token might not be generated and therefore not be passed at all. In such a case, connectUser shouldn't be called from what I understand.

khushal87: Please correct me if I am wrong.

vanGalilea (approved):

PR #547, opened 2023-05-24
fix: remove media devices context to fix unnecessary rerenders

santhoshvai:
while working with reconnection flow I found that there were many unnecessary rerenders due to MediaDevicesContext. 

The easy way to fix was to move the values into the video store that we provide which supports atomic state updates. Also reduces the LOC by ~35 lines ;)


packages/react-native-sdk/src/contexts/StreamVideoContext/MediaDevices.tsx
khushal87, line 1: How about making it a hook since it returns nothing and mostly has state and effect? This way, it can be used as a hook in the `StreamVideoStoreProvider` rather than a component.

vanGalilea: Good one. This looks like a hook to me

vanGalilea:
Very clean PR @santhoshvai, toggle camera issue might be related or solved with this?

vanGalilea:
Please have a look at SOnar's 4o something percent duplicated code warning, resolve/fix

packages/react-native-sdk/src/contexts/StreamVideoContext/MediaDevices.tsx
santhoshvai: It looks like it can be made to a hook but sadly cant be

The use case is below

```
  return (
    <Provider>
      <MediaDevices />
      {props.children}
    </Provider>
  );
```

There needs to be a component that runs the effect and we shouldn't export the effect to integrators

vanGalilea: Add to a wrapper comp and call it from there or keep like this, your call @santhoshvai whatever you find elegant!

santhoshvai: keeping it the same @vanGalilea due to simplicity

PR #557, opened 2023-05-26
docs: quickstart tutorial updates

oliverlaz:
- fix `yarn` example
- lazy-load the codesandbox

### DIFF
--- a/packages/react-sdk/docusaurus/docs/React/01-basics/03-quickstart.mdx
+++ b/packages/react-sdk/docusaurus/docs/React/01-basics/03-quickstart.mdx
@@ -20,7 +20,7 @@ Install the SDK using your package manager of choice, such as `npm`, `yarn`, or 
 
 ```bash
 # example with yarn
-yarn install @stream-io/video-react-sdk
+yarn add @stream-io/video-react-sdk
 ```
 
 Collect the necessary configuration parameters, as:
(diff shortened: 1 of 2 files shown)


szuperaz (approved):

packages/react-sdk/docusaurus/docs/React/01-basics/03-quickstart.mdx
```diff
@@ -20,7 +20,7 @@ Install the SDK using your package manager of choice, such as `npm`, `yarn`, or
 
 ```bash
 # example with yarn
-yarn install @stream-io/video-react-sdk
+yarn add @stream-io/video-react-sdk
```
szuperaz, line 23: I've checked and in other places we seem to use `yarn add` in the docs. I assume we need to change because that command works with all Yarn versions?

szuperaz: Hm, no, livestream tutorial uses `yarn install` as well

szuperaz: Pushed a fix for that as well

oliverlaz: I didn't check the other pages, I just spotted I made a mistake here (and obviously in the Livestream tutorial 😅). 
Thanks for correcting it as well

PR #558, opened 2023-05-26
docs: create UI component pages

szuperaz:
## What's in this PR?

Empty doc pages for UI Components

## Guidelines
1. Not all components have a page
- The logic mainly follows the component folder structure in the source code, for example, in the source code, we have the `CallParticipantsList` folder, and we have a doc page called `Participants list`, but there is no separate page for example for the `CallParticipantsList/EmptyParticipantSearchList` component
- Some component folders are not included in the docs, for example the `Icon`. My logic was this: I've excluded all components that don't contain stream-video specific logic. While I can imagine a use-case where someone wants to use the `Icon` component to create their own UI, but it seems rather rare, and I think the component list would be rather long if we were to include all small components.
2. The ToC follows this structure
- Core - our core components
- Call - call-related components (except participant-related components)
- Participant - participant-related components
- Utility - misc call/participant related components - these are either not connected very tightly to calls (for example `Avatar`) or they are expected to be used less frequently (for example `CallStatistics`) 

image

Let me know what you think about the structure. @DaemonLoki I'm especially interested in your opinion because I know you've already given this area some thought.

Once we have the structure, I'll update the existing TODO component links in a separate PR.

### DIFF
--- a/packages/react-sdk/docusaurus/docs/React/03-ui-components/01-overview.mdx
+++ b/packages/react-sdk/docusaurus/docs/React/03-ui-components/01-overview.mdx
@@ -8,7 +8,13 @@ description: Overview of the UI components
 There are different ways to use the React Video SDK to build out UI. It comes with powerful built-in components ready for use.
 In addition, it allows for styling and custom theming options with custom CSS.
 
-For custom use cases it is also possible to create your own components to fully tailor the SDK to your needs. Please note that we suggest not to override a few, core components as those contain complex, low-level logic related to performance, the list of these components can be found in the [Core Components page](../../basics/core-components/).
+For custom use cases it is also possible to create your own components to fully tailor the SDK to your needs.
+
+:::note
+
+Please note that we suggest not to override a few, core components as those contain complex, low-level logic related to performance, the list of these components can be found in the [Core Components page](../../basics/core-components/).
+
+:::
 
 This page guides you through all the options you have with increasing levels of customization.
 
--- a/packages/react-sdk/docusaurus/docs/React/03-ui-components/01-overview.mdx
+++ b/packages/react-sdk/docusaurus/docs/React/03-ui-components/01-overview.mdx
@@ -3,10 +3,6 @@ title: Overview
 description: Overview of the UI components
 ---
 
-:::warning
-TODO create pages for the components
-:::
-
 ## Introduction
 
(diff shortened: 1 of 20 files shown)


szuperaz:
> Hey Zita, I like the structure as is.
> 
> One food for thought though: Do we need the _Core_ section? Technically, all that's in there could be either put into the _Call_ section (`Call layout`, `StreamCall`) or the _Participant_ section (`Participant`).
> 
> Also: do we still require to add the `Audio` and/or `Video` components or is that not necessary anymore? Because if yes, they should definitely be documented as well!
> 
> But aside from that this looks great to me!

About the `Core` section: I was thinking about this a lot, the reason I've decided to include it to make it very clear to integrators that these are special components, not like the rest.

The `Audio` and `Video` elements are abstracted by the `ParticipantView` (`Core/Participant in the docs), so integrators shouldn't need to know about these components.

DaemonLoki:
> The Audio and Video elements are abstracted by the ParticipantView (Core/Participant in the docs), so integrators shouldn't need to know about these components.

> About the Core section: I was thinking about this a lot, the reason I've decided to include it to make it very clear to integrators that these are special components, not like the rest.

Okay, very good to know. I myself thought it may be confusing to integrators that we have this separation. But I understand your explanation and it's awesome that you put so much thought into this. Should be fine from my side! :) 

All right, that makes sense, then!

oliverlaz (approved):

packages/react-sdk/docusaurus/docs/React/03-ui-components/01-overview.mdx
```diff
@@ -3,10 +3,6 @@ title: Overview
 description: Overview of the UI components
 ---
 
-:::warning
-TODO create pages for the components
-:::
-
 ## Introduction
 
 There are different ways to use the React Video SDK to build out UI. It comes with powerful built-in components ready for use.
```
MartinCupela, line 8: Related to the conversation about the core components, I would maybe put the below paragraph into a note banner to make it stand out a bit more:

```
For custom use cases it is also possible to create your own components to fully tailor the SDK to your needs. Please note that we suggest not to override a few, core components as those contain complex, low-level logic related to performance, the list of these components can be found in the [Core Components page](../../basics/core-components/).
```

MartinCupela (approved):

PR #570, opened 2023-05-30
docs: update call lifecycle and ring call docs

oliverlaz:
### Overview

- Adds the information about Call Sessions
- Updates around ring call flow


packages/react-sdk/docusaurus/docs/React/02-guides/02-call-lifecycle.mdx
```diff
@@ -106,20 +143,23 @@ This is the happy path, when both parties join the active call. The
...
+    Back-end-->>Client_B: 3. Event: 'call.ring'
+    Client_B->>Back-end: 4. Accept the call
+    Back-end-->>Client_A: 5. Event: 'call.accepted'
+    Client_B->>SFU: 6.  Join the call
+    Client_A->>SFU: 7. Join the call
+    Back-end-->>Client_A: 8. Event: 'call.session_started'
+    Back-end-->>Client_B: 9. Event: 'call.session_started'
+    Back-end-->>Client_B: 8. Event: 'call.session_participant_joined'
+    Client_A->>SFU: 9. Leave the call
+    Back-end-->>Client_B: 10. Event 'call.session_participant_left'
```
MartinCupela, line 159: Is this the coordinator or SFU?

oliverlaz: This is the coordinator. I don't think that we should be documenting the SFU events as it is in our best interest to keep them internal. Especially for this use case, where we have two different events (with slightly different data) for the very same thing.

MartinCupela (approved):

PR #572, opened 2023-05-30
feat: add audio room demo app

DaemonLoki:


packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
 
-TODO
+In this guide, you'll build an audio room experience similar to Twitter Spaces / Clubhouse.
+
+The final result will look like this:
+
+// TODO add sample video
+
+The code for the final result can also be found on GitHub, so if you're only interested in seeing the resulting code, go [here](https://github.com/GetStream/stream-video-ios-examples/tree/main/AudioRooms).
+
```
vanGalilea, line 13: Links to an iOS SDK code

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+### Creating rooms via the API
+
+Now that we have user authentication out of our way we can start building up our app.
+The first thing we'll do is to allow users to create an audio room where others can then join.
+
+The data model for this needs a few properties. Create a new Typescript file, call it `AudioRoom` and fill it up with this:
+
+```ts
+export interface AudioRoom {
```
szuperaz, line 135: I'd include the import statements in the code snippets, it's always useful to see which types coming from the SDK and which are user-defined

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+Want to learn why we need the token provider and how it works? See [this page](../../guides/client-auth/#token-providers) for more detail.
+:::
+
+First, we can collect the API key from our `.env` file by simply importing it like this:
+
+```
+const apiKey = import.meta.env.VITE_STREAM_API_KEY as string;
+```
+
+Next, to get a token, we need to create one in our backend. Whenever this expires the SDK will call the token provider again to renew the token.
```
szuperaz, line 81: I think that part is already covered in the Authentication guide, so linking to that seems enough for me, but if we leave that part in, we should mention the [token generator](https://staging.getstream.io/video/docs/react/guides/client-auth/#introduction) as well, as probably most users won't have their token generator endpoint ready by the time they are doing this tutorial

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+const randomId = Math.random().toString(36).substring(2, 12);
+// highlight-next-line
+const call = client?.call('audio_room', randomId);
+call?.getOrCreate({
+  data: {
+    // highlight-next-line
+    members: [{ user_id: user?.id || '', role: 'admin' }],
+    custom: {
+      // highlight-start
+      audioRoomCall: true,
```
szuperaz, line 169: Isn't this enough to use the `type` field for this?

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+    });
+}, [client, setRooms]);
+```
+
+We need a reference to the `client` which we conveniently can get with the `useStreamVideoClient` hook.
+Then, we call the `.queryCalls` function in a `useEffect` hook. We can specify three things there:
+
+1. `filter_conditions`: we want to only have calls with the `audioRoomCall` property set to true (which we specified earlier)
+2. `sort`: we _could_ sort the calls we receive by some logic we want, for simplicity reasons we don't do this here
+3. `watch`: this simple, but powerful property gives us the power to listen to real-time changes; whenever a call changes (for example going live) we'll get notified, and our rooms list updates
```
szuperaz, line 246: We could link the "Querying calls" guide here for more information

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+});
+```
+
+This will tell the SDK to join the room when the component is rendered and with the callback, we return in `useEffect` it will also leave calls when it's unmounted.
+
+Next, to show the room and provide audio output and input we need to wrap the component that does that with two providers:
+
+1. `MediaDevicesProvider`: this one is needed to give access to audio and video devices; it has two parameters, `initialAudioEnabled` and `initialVideoEnabled` which do exactly what they describe.
+2. `StreamCallProvider`: it is initialized with a `Call` and provides us with a lot of useful and convenient hooks to have easy access to the necessary call parameters
+
```
szuperaz, line 277: There were some changes around this, you can use the `StreamCall` component here: https://github.com/GetStream/stream-video-js/blob/main/packages/react-sdk/src/core/components/StreamCall/StreamCall.tsx

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+That gives us everything we need to jump into the component that does the real call rendering.
+
+### Building up the UI for a room
+
+Again, we don't want to focus too much on the UI part of it but instead show the concepts of getting access to the necessary data.
+And just as a reminder, you can still take a look at the final project [here](TODO: link to project) to get inspiration.
+
+Inside our `RoomActive` component, we need access to a few things. Luckily, the Stream Video SDK exposes hooks for many of
+them (which is why we needed to wrap the component in a `StreamVideoProvider`).
```
szuperaz, line 300: We could reference the "Call & Participant State" guide here for more info

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+This will render the list of participants. If we want to render the speakers, we need to wrap them with the `Audio` element.
+It is required to allow for listening to their audio stream.
+
+Here's how to use it:
+
+```tsx
+<div>
+  {speakers.map((speaker) => (
+    <div key={speaker.userId}>
+      <Audio muted={false} audioStream={speaker.audioStream} />
```
szuperaz, line 342: Can't we use the `ParticipantView` here? I think that is a higher level component, but I'm not sure it would work for the use-case

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+
+/* more code */
+async function requestSpeakingPermission() {
+  await call?.requestPermissions({
+    permissions: [OwnCapability.SEND_AUDIO],
+  });
+}
+```
+
+Same as in the previous chapter, we can listen to new events of type `call.permission_request` to get notified when there's a new request coming in.
```
szuperaz, line 445: This is new, but we have the `usePermissionRequest` hook to avoid subscribing to WS events

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,543 @@ title: Audio Room Tutorial
...
+      unsubscribe();
+    }
+  };
+}, []);
+```
+
+:::note
+In our case we track the speakers in a list called `speakerIds` and filter them accordingly. You can do things differently here.
+:::
+
```
szuperaz, line 526: We could link to the "Permission Requests" guide for more info

szuperaz (commented):
One general remark: I think it's better to have more meaningful link names as opposed to "here", "this page" etc. but that's a small thing.

In general the tutorial is clear to me, and it only contains the important parts.

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,4 +3,542 @@ title: Audio Room Tutorial
...
+  tokenOrProvider: tokenProvider,
+  user,
+});
+```
+
+Whenever a user hits the login button we can authenticate them using a call to `connectUser`:
+
+```tsx
+await client.connectUser(user, user.token);
+```
```
oliverlaz, line 112: With the current state, `useCreateStreamVideoClient()` automatically connects the provided user. If you want to delay it, you'll need to use the other, more imperative way of constructing a `StreamVideoClient`:

```ts
const [client] = useState(() => new StreamVideoClient(apiKey));
<StreamVideo client={client}>
   <MyComponent onClick={() => client.connectUser(user)} />
</StreamVideo>
```

sample-apps/react/audio-rooms/src/components/home/Home.tsx
```diff
@@ -0,0 +1,48 @@
...
+  const client = useCreateStreamVideoClient({
+    apiKey,
+    tokenOrProvider: user?.token || '',
+    user: {
+      id: user?.id || '',
+    },
+  });
+
+  useEffect(() => {
+    login(userTapped, client);
```
MartinCupela, line 32: `client.connectUser` is already executed inside `useCreateStreamVideoClient`. You can set the state as logged in once there is connected user available (`useConnectedUser` hook) .

PR #576, opened 2023-05-31
fix(react-sdk/react-dogfood): move old layouts

arnautov-anton:
This PR moves deprecated `CallParticipantsScreenView` and `CallParticipantsView` layout components to Pronto.


MartinCupela:
Looks like messenger clone is using the component too
image

PR #579, opened 2023-06-01
feat(react-dogfood): keyboard shortcuts

arnautov-anton:
- add toggle mute shortcuts for audio and video
- add raise hand reaction shortcut


arnautov-anton:
I had to extract toggle functions from the actual buttons as they contain a ton of logic - was that what you meant before when you were writing the docs on CallControls, @MartinCupela?

packages/react-dogfood/hooks/useKeyboardShortcuts.ts
```diff
@@ -0,0 +1,83 @@
...
+  useCall,
+  useToggleAudioMuteState,
+  useToggleVideoMuteState,
+} from '@stream-io/video-react-sdk';
+import { useEffect } from 'react';
+
+import hotkeys from 'hotkeys-js';
+
+enum Hotkeys {
+  TOGGLE_AUDIO_MAC = 'cmd+d',
```
oliverlaz, line 12: Can we also have "cmd+space"? I think that is a more common shortcut that would be easier to remember.

arnautov-anton: I used the ones used in the [Google Meet](https://support.google.com/a/users/answer/9896256?hl=en) - would you rather me to switch to the Slack preset?

oliverlaz: can we have both? 😅

arnautov-anton: image

oliverlaz (approved):

PR #581, opened 2023-06-01
docs: Participant view

szuperaz:


oliverlaz (approved):

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
```diff
@@ -3,8 +3,130 @@ id: participant-view
...
+        // highlight-next-line
+        <ParticipantView participant={p} />
+      ))}
+      <CallControls />
+    </StreamCall>
+  );
+};
+```
+
+## Props
```
MartinCupela, line 56: I would prefer to have the same doc comments in the source code and in the documentation. WDYT?

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
```diff
@@ -3,8 +3,130 @@ id: participant-view
...
+
+### `participant`
+
+| Type                                                                                                            |
+| --------------------------------------------------------------------------------------------------------------- |
+| [`StreamVideoParticipant`](https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/types.ts) |
+
+The participant whose video/audio feed we want to show.
```
MartinCupela, line 64:
```suggestion
The participant whose video/audio feed we want to reproduce.
```

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
```diff
@@ -3,8 +3,130 @@ id: participant-view
...
+
+### `className`
+
+| Type                    |
+| ----------------------- |
+| `string` \| `undefined` |
+
+Custom class applied to the root DOM element.
+
+### `refs`
```
MartinCupela, line 116: I believe that this prop would definitely deserve more information about why and how.

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
szuperaz: hm, I wouldn't understand what reproduce means in this context, I think show is more clear, even though we can't see audio

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
```diff
@@ -3,8 +3,130 @@ id: participant-view
...
+
+![Preview of the Call controls component.](../../assets/03-ui-components/participant.png)
+
+:::note
+
+Please note that we suggest not to create a custom `ParticipantView` component as it contains complex, low-level logic related to performance. Please refer to the [Customization section](#customization) to see how you can tailor the component to your needs.
 
 :::
+
+## General usage
```
MartinCupela, line 23: I would mention somewhere in the document that the component is used in multiple call layout components + link to the call-layout.mdx doc.

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
MartinCupela: AI hints synonyms:
- present
- broadcast

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
arnautov-anton: These are currently used only in the `egress-composite` application to notify egress that the "application" is ready. Outside of that I haven't really found a use for these - maybe for the actual web picture in picture API.

But the main point is that the component is "isolated" and it'd be hard to access these elements without it.

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
```diff
@@ -3,8 +3,130 @@ id: participant-view
...
 
-:::warning
+The `ParticipantView` component is a core component that renders a participant's video and plays the participant's audio.
 
-TODO: write about: https://github.com/GetStream/stream-video-js/tree/main/packages/react-sdk/src/core/components/ParticipantView
+It can toggle between video and avatar based on the participant's video state. It displays information about the participant, their device settings and connection quality. It also provides action buttons (for example, to pin the participant).
+
+To optimize performance,
+
+- the component won't stream video for participants that are not visible in the viewport
```
arnautov-anton, line 12: *only when viewport element is set through `call.viewportTracker.setViewport`

We could probably direct them [here](https://github.com/GetStream/stream-video-js/blob/db0a25d1ac584d67e5b38d5cc4b3a0c834012991/packages/react-sdk/docusaurus/docs/React/06-ui-cookbook/13-custom-call-layout.mdx?plain=1#L314) for clarification.

szuperaz: I understand the comment, but I've looked at the linked section, and if we link it here it looks like as if integratiors always have to deal with this, which is confusing. They only have to deal with that when they want a custom layout. So I think it's better not to mention that here, they don't have to know about this if they are using a defaul layout, if they do a custom one, they will see that part anyway.

arnautov-anton: That's a good point. 👍

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
szuperaz: 👍 synced them

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
szuperaz: I settled for play 😄

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
szuperaz: Done

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
```diff
@@ -3,8 +3,130 @@ id: participant-view
...
+An object with set functions meant for exposing the video and video placeholder elements to the integrators.
+
+## Customization
+
+The [ParticipantView customizations guide](../../../ui-cookbook/participant-view-customizations) tells all the important informiation about the customization options.
+
+We also have additional guides focusing on specific parts:
+
+- [Video placeholder](../../../ui-cookbook/video-placeholder)
+- [Connection Quality Indicator](../../../ui-cookbook/connection-quality-indicator)
```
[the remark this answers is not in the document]
szuperaz: Done

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
```diff
@@ -3,8 +3,130 @@ id: participant-view
...
+  return (
+    <StreamCall
+      callType="default"
+      callId="my-first-call"
+      data={{ create: true }}
+      autoJoin
+    >
+      {participants.map((p) => (
+        // highlight-next-line
+        <ParticipantView participant={p} />
```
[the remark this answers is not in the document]
szuperaz: Thanks, done

packages/react-sdk/docusaurus/docs/React/03-ui-components/core/participant-view.mdx
szuperaz: Added some extra explanation (mentioned that they can use this if they need to add custom event handler to these elements)

PR #586, opened 2023-06-01
docs(react-sdk): rename 13-custom-call-layout.mdx to 04-video-layout.mdx

MartinCupela:


arnautov-anton:
So - my understanding is that we're dropping "call layouts" for "video layouts"? If that's the case then we should probably adjust all of the occurences across the whole documentation and probably rename folders in the SDK as well.

MartinCupela:
> So - my understanding is that we're dropping "call layouts" for "video layouts"? If that's the case then we should probably adjust all of the occurences across the whole documentation and probably rename folders in the SDK as well.

I don't think that documentation title has to force the names used in code.

arnautov-anton:
> > So - my understanding is that we're dropping "call layouts" for "video layouts"? If that's the case then we should probably adjust all of the occurences across the whole documentation and probably rename folders in the SDK as well.
> 
> I don't think that documentation title has to force the names used in code.

I believe it could cause a confusion or two when you read a guide on a "video layout" but everywhere else in the docs we refer to these as "call layouts".

szuperaz:
> > > So - my understanding is that we're dropping "call layouts" for "video layouts"? If that's the case then we should probably adjust all of the occurences across the whole documentation and probably rename folders in the SDK as well.
> > 
> > 
> > I don't think that documentation title has to force the names used in code.
> 
> I believe it could cause a confusion or two when you read a guide on a "video layout" but everywhere else in the docs we refer to these as "call layouts".

I would also update the docs, it's confusing to have two different names

oliverlaz (approved):

PR #588, opened 2023-06-01
docs(react-sdk): add CallRecordingList component documentation

MartinCupela:


szuperaz (approved):
A general remark: I think we should only use GH links when really necessary, because GH links will break silently (without any warning) when we rename/move something in the source code, so they are hard to maintain. I think in this case we don't need to provide links to component's source code or to the building-block components, integrators shouldn't care about those for the majority of the time.

oliverlaz (approved):

MartinCupela:
> A general remark: I think we should only use GH links when really necessary, because GH links will break silently (without any warning) when we rename/move something in the source code, so they are hard to maintain. I think in this case we don't need to provide links to component's source code or to the building-block components, integrators shouldn't care about those for the majority of the time.

I have remove the GH links to the subcomponents and kept only the main link to `CallRecordingList`

PR #597, opened 2023-06-05
fix: use and save toggled dtx to sdp and remove empty lines when munging

vanGalilea:
## Goal
- Use and save toggled DTX (prior to this PR, the munged SDP was ignored).
- Remove empty lines after removing a codec (Since a BE change that enabled RED we couldn't publish video/audio on RN)

### DIFF
--- a/packages/client/src/rtc/publisher.ts
+++ b/packages/client/src/rtc/publisher.ts
@@ -342,7 +342,7 @@ export class Publisher {
     const offer = await this.publisher.createOffer();
     let sdp = offer.sdp;
     if (sdp) {
-      toggleDtx(sdp, this.isDtxEnabled);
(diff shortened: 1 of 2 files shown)


packages/client/src/rtc/publisher.ts
```diff
@@ -342,7 +342,7 @@ export class Publisher {
     const offer = await this.publisher.createOffer();
     let sdp = offer.sdp;
     if (sdp) {
-      toggleDtx(sdp, this.isDtxEnabled);
+      sdp = toggleDtx(sdp, this.isDtxEnabled);
```
vanGalilea, line 345: Testing React DF is needed

vanGalilea: Tested 👍🏻 on DF preview

santhoshvai (approved):

PR #602, opened 2023-06-06
Don't get video if device id is undefined

vanGalilea:


packages/react-native-sdk/src/hooks/useLocalVideoStream.ts
```diff
@@ -18,8 +18,17 @@ export const useLocalVideoStream = () => {
...
+      if (!currentVideoDevice?.deviceId) return null;
+
+      const stream = await getVideoStream({
+        deviceId: currentVideoDevice.deviceId,
+      });
+      setVideoStream((previousStream) => {
+        if (!previousStream) return stream;
+
+        disposeOfMediaStream(previousStream);
+      });
```
santhoshvai, line 31: Currently this effect will never update the stream if the deviceId changes 

```suggestion
      setVideoStream((previousStream) => {
        if (previousStream) disposeOfMediaStream(previousStream);
        return stream;
      });
```

packages/react-native-sdk/src/hooks/useLocalVideoStream.ts
```diff
@@ -18,8 +18,17 @@ export const useLocalVideoStream = () => {
...
     };
     loadVideoStream();
   }, [currentVideoDevice]);
```
santhoshvai, line 34:
```suggestion
  }, [currentVideoDevice.deviceId]);
```

This will avoid unnecessary useEffect runs.. in case the deviceId is the same but the currentVideoDevice object updates

santhoshvai (approved):

PR #603, opened 2023-06-06
feat: SFU Session Migration and Graceful Shutdown

oliverlaz:
### Overview

Initial implementation of the SFU Session Migration and Graceful SFU shutdown handling.


packages/client/src/store/CallState.ts
```diff
@@ -60,6 +60,11 @@ export enum CallingState {
    */
   RECONNECTING = 'reconnecting',
 
+  /**
+   * The call is in the process of migrating from one node to another.
+   */
+  MIGRATING = 'migrating',
+
```
szuperaz, line 67: `CallState` is a public API that integrators can use to for example display loading indicators. I think they shouldn't care about if we are in `MIGRATING` or `RECONNECTING` state, I'd rather not leak this information to the public API.

packages/client/src/Call.ts
```diff
@@ -697,24 +726,53 @@ export class Call {
     // - SFU crash or restart
     // - network change
     sfuClient.signalReady.then(() => {
+      // register a handler for the "goAway" event
+      const unregisterGoAway = this.dispatcher.on('goAway', (event) => {
+        if (event.eventPayload.oneofKind !== 'goAway') return;
+        const { reason } = event.eventPayload.goAway;
+        console.log(
```
szuperaz, line 733: We should use the `logger` method here

szuperaz (approved):

PR #604, opened 2023-06-07
fix(react-native): revive call flow in dogfood app

khushal87:


packages/react-native-sdk/src/hooks/useCallCycleEffect.tsx
```diff
@@ -28,7 +28,7 @@ export const useCallCycleEffect = (
 
   useEffect(() => {
     if (!call || NON_ACTIVE_CALLING_STATES.includes(callingState)) return;
-    const isCallCreatedByMe = call.data?.created_by.id === call?.currentUserId;
+    const isCallCreatedByMe = call?.data?.created_by.id === call?.currentUserId;
```
vanGalilea, line 31:
```suggestion
    const isCallCreatedByMe = call.data?.created_by.id === call?.currentUserId;
```

vanGalilea (approved):

PR #610, opened 2023-06-08
feat(react-sdk): ScreenShareOverlay component

arnautov-anton:
localhost_3000_join_lKie2H3pSTgi


packages/styling/src/ScreenShareOverlay/ScreenShareOverlay-theme.scss
```diff
@@ -0,0 +1,19 @@
+.str-video__screen-share-overlay {
+  background: rgba(0, 0, 0, 0.9);
```
oliverlaz, line 2: Should we expose variables for this? Also, I think we should reduce the opacity to at least `0.8`. I think it is useful for presenters to be able to see what they are presenting :)

arnautov-anton: I'll either reuse the overlay color we have or create another one alongside the one we have. As for the opacity - design specification does not have this at all, this was my addition but we can bump it down to .8. 👍

arnautov-anton: localhost_3000_join_lKie2H3pSTgi (1)

This is what it looks like with the overlay color we already have defined - what do you think?

oliverlaz (approved):

PR #617, opened 2023-06-08
fix(react-sdk): do not try to watch devices, if browser permission is  not granted

MartinCupela:


packages/react-sdk/src/core/hooks/useDevices.ts
```diff
@@ -1,11 +1,50 @@
...
+    };
+    const checkPermissions = async () => {
+      try {
+        permissionState = await navigator.permissions.query({
+          name: permissionName,
+        });
+        if (permissionState.state !== 'granted') {
+          permissionState.addEventListener('change', handlePermissionChange);
+        } else {
+          enableSubscription(true);
```
szuperaz, line 30: I think we should register the change event handler if the state is granted, as it can be revoked anytime

szuperaz (approved):

packages/react-sdk/src/core/hooks/useDevices.ts
```diff
@@ -1,11 +1,50 @@
-import { useEffect, useState } from 'react';
+import { ChangeEvent, useEffect, useState } from 'react';
 import { Observable, pairwise } from 'rxjs';
 import {
   getAudioDevices,
   getAudioOutputDevices,
   getVideoDevices,
 } from '@stream-io/video-client';
 
+export const useCheckBrowserPermissions = (permissionName: PermissionName) => {
```
oliverlaz, line 9: I'd name this hook `useHasBrowserPermission`. "Check" feels like more "momentarily" assertion, while this hook can emit a new value anytime.

MartinCupela: Good idea, I have renamed it

packages/react-sdk/src/core/hooks/useDevices.ts
MartinCupela: Thank you Zita, I have fixed the code

oliverlaz (approved):

PR #618, opened 2023-06-08
feat(client): do not unregister left calls from client store

MartinCupela:
Unregistering leads to `calls` array returned from `useCalls()` not containing the left call.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -401,7 +401,6 @@ export class Call {
     // Call all leave call hooks, e.g. to clean up global event handlers
     this.leaveCallHooks.forEach((hook) => hook());
 
-    this.clientStore.unregisterCall(this);
     this.state.setCallingState(CallingState.LEFT);
   };
 
--- a/packages/react-native-call-starter-kit/src/components/MessengerWrapper.tsx
+++ b/packages/react-native-call-starter-kit/src/components/MessengerWrapper.tsx
@@ -7,6 +7,7 @@ import React, {
 } from 'react';
 import {
   ActiveCall,
+  CallingState,
   IncomingCallView,
   OutgoingCallView,
   StreamCall,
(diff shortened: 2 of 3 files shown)


khushal87 (approved):
Hey @MartinCupela, this is a good change, but it will break the RN SDK's implementation to handle/ignore more than one call. Therefore, I am pushing the fix for the same in your PR. Thanks 😄 

Cc: Tagging you @vanGalilea, so that you are aware of the change.

szuperaz (approved):

oliverlaz:
I wonder whether we need to expose some "clean-up" or "update" API. If we don't remove terminated calls from the store, it would be hard to re-register them.
Think about a client that deals with many calls that also re-joins calls... 
```ts
const call = client.call('default', 'a');
call.join();
call.leave();

const newCall = client.call('default', 'a');
call.join();

const calls = useCalls(); // here `calls` will have a reference to the first call (the one that we left before)
```

I think this would potentially cause a lot of problems in the future.

MartinCupela:
> I wonder whether we need to expose some "clean-up" or "update" API. If we don't remove terminated calls from the store, it would be hard to re-register them. Think about a client that deals with many calls that also re-joins calls...
> 
> ```ts
> const call = client.call('default', 'a');
> call.join();
> call.leave();
> 
> const newCall = client.call('default', 'a');
> call.join();
> 
> const calls = useCalls(); // here `calls` will have a reference to the first call (the one that we left before)
> ```
> 
> I think this would potentially cause a lot of problems in the future.

I would say, that a call with the same CID should replace the old call.

MartinCupela:
Closing this PR after discussion with @oliverlaz . The calls array emitted by `useCalls` is originally thought of as a array of watched calls that. `client.queryCalls` does not necessary return watched calls only. However `client.queryCalls` updates calls array emitted by `useCalls` a so we should not rely on `useCalls` for building dashboards but rather for ring call workflow. This caveats can bring some issues that will have to be thought of.

PR #621, opened 2023-06-09
chore: Configure npm release for react native

szuperaz:


packages/react-native-sdk/package.json
```diff
@@ -11,6 +11,15 @@
     "test": "jest --coverage",
     "generate-docs": "../../scripts/generate-docs.sh reactnative @stream-io/video-react-native-sdk react-native-sdk"
   },
+  "files": [
```
szuperaz, line 14: Copied from react sdk, please validate

szuperaz:
This is the release CI action: https://github.com/GetStream/stream-video-js/blob/main/.github/workflows/version-and-release.yml - feel free to add extra steps you want to run before the release

packages/react-native-sdk/package.json
vanGalilea: Should work 👍🏻

vanGalilea (approved):

PR #623, opened 2023-06-09
fix(react-native): incorrect way of sending device Id to get streams

santhoshvai:
Found this typing error via eslint

### DIFF
--- a/packages/react-native-sdk/src/hooks/useCallControls.tsx
+++ b/packages/react-native-sdk/src/hooks/useCallControls.tsx
@@ -47,14 +47,13 @@ export const useCallControls = () => {
     try {
       // Client picks up the default audio stream.
       // For mobile devices there will always be one audio input
-      if (audioDeviceId && isOnlineRef.current) {
-        const audioStream = await getAudioStream({
-          deviceId: audioDeviceId,
-        });
-        if (call) {
-          await call.publishAudioStream(audioStream);
-        }
-      }
+      if (!call || !audioDeviceId || !isOnlineRef.current) return;
+
+      const audioStream = await getAudioStream({
+        deviceId: audioDeviceId,
+      });
+        
+      await call.publishAudioStream(audioStream);
     } catch (e) {
       console.log('Failed to publish audio stream', e);
     }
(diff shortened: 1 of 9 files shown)


packages/react-native-sdk/src/hooks/useCallControls.tsx
```diff
@@ -40,33 +40,40 @@ export const useCallControls = () => {
...
-      if (currentAudioDevice && isOnlineRef.current) {
-        const audioStream = await getAudioStream(currentAudioDevice.deviceId);
+      if (audioDeviceId && isOnlineRef.current) {
+        const audioStream = await getAudioStream({
+          deviceId: audioDeviceId,
+        });
         if (call) {
           await call.publishAudioStream(audioStream);
         }
       }
```
vanGalilea, line 57:
```suggestion
      if (!call || !audioDeviceId || !isOnlineRef.current) return;

      const audioStream = await getAudioStream({
        deviceId: audioDeviceId,
      });
        
      await call.publishAudioStream(audioStream);
```

vanGalilea: if call is not there no need for getting the stream

packages/react-native-sdk/src/hooks/useCallControls.tsx
```diff
@@ -40,33 +40,40 @@ export const useCallControls = () => {
...
-      if (currentVideoDevice && isOnlineRef.current) {
-        const videoStream = await getVideoStream(currentVideoDevice.deviceId);
+      if (videoDeviceId && isOnlineRef.current) {
+        const videoStream = await getVideoStream({
+          deviceId: videoDeviceId,
+        });
         if (call) {
           await call.publishVideoStream(videoStream);
         }
       }
```
vanGalilea, line 72:
```suggestion
      if (!call || !videoDeviceId || !isOnlineRef.current) return;
      
      const videoStream = await getVideoStream({
        deviceId: videoDeviceId
      });
      
      await call.publishVideoStream(videoStream);
```

PR #626, opened 2023-06-10
feat(react-native): push notification for android

santhoshvai:
## New

- Most push code now lives inside SDK
- We send token to stream and we now get incoming call notification from stream
- We now show incoming call notification through notifee we dont use callkeep ( I did this based on the notion doc), but we can easily swap to callkeep 
- Unhappy path of decline call and caller must receive reject
- ask for notification permissions from SDK automatically

## Tested

The happy path was tested and works as shown in the video (call between iOS simular and Android device)

https://github.com/GetStream/stream-video-js/assets/3846977/c4a871e8-7751-43aa-ab4e-c592f103dc7a

The unhappy path was also tested.

## TODOs

- When caller closes the call, remove notification
- Docusaurus documentation.. better to do in a new PR so that its easy to review and also easier to do after iOS implementation


packages/react-native-dogfood/Gemfile
```diff
@@ -8,4 +8,4 @@ gem "fastlane"
 plugins_path = File.join(File.dirname(__FILE__), 'fastlane', 'Pluginfile')
 eval_gemfile(plugins_path) if File.exist?(plugins_path)
 
-gem 'cocoapods', '~> 1.11', '>= 1.11.3'
+gem 'cocoapods', '~> 1.12', '>= 1.12.1'
```
santhoshvai, line 11: unrelated change but important, aligned with the one in podfile.lock to prevent future mishaps

packages/react-native-dogfood/src/screens/Call/JoinCallScreen.tsx
```diff
@@ -134,6 +134,7 @@ const JoinCallScreen = () => {
           placeholder="Enter comma separated User Ids"
           style={styles.textInput}
           value={ringingUserIdsText}
+          autoCapitalize="none"
```
santhoshvai, line 137: unrelated change.. but we better shouldnt autocapitalise the custom user name calls

packages/react-native-sdk/src/components/ParticipantReaction.tsx
santhoshvai, line 1: unrelated change.. due to eslint issues

packages/react-native-sdk/src/components/ReactionsModal.tsx
santhoshvai, line 1: unrelated change.. due to eslint issues

packages/react-native-sdk/src/utils/push/index.ts
```diff
@@ -0,0 +1,169 @@
...
+//     additionalPermissions: [],
+//     // Required to get audio in background when using Android 11
+//     // foregroundService: {
+//     //   channelId: 'com.company.my',
+//     //   channelName: 'Foreground service for my app',
+//     //   notificationTitle: 'My app is running on background',
+//     //   notificationIcon: 'Path to the resource icon of the notification',
+//     // },
+//   },
+// };
```
santhoshvai, line 33: ignore the above comment for now.. callkeep stuff is still WIP

packages/react-native-sdk/src/utils/push/index.ts
```diff
@@ -0,0 +1,169 @@
...
+            },
+          },
+        ],
+      },
+    });
+    // const callkeep = getCallKeepLib();
+    // const uuid = message.data.call_cid;
+    // const localizedCallerName = message.data.created_by_display_name;
+    // const handle = 'handle'; // Phone number of the caller // TODO: unclear what this is used for?!
+    // callkeep.displayIncomingCall(uuid, handle, localizedCallerName);
```
santhoshvai, line 128: ignore above comment, we may swap to callkeep instead of notifee for incoming call in android

packages/react-native-dogfood/src/navigators/Call.tsx
santhoshvai, line 1: unlreated change.. but important.. unmounting active call component while joining would break rejoining flow.. so removed it..

we had already removed it in meeting before

vanGalilea:
> We now show incoming call notification through notifee we dont use callkeep ( I did this based on the notion doc), but we > can easily swap to callkeep

1. What's the idea behind it? Facilitate other PN, like starting a live stream etc. T
2. Why should we ever use callkeep? What are the benefits vs. Notifee?

packages/react-native-dogfood/src/screens/Meeting/GuestMeetingScreen.tsx
```diff
@@ -75,15 +70,7 @@ export const GuestMeetingScreen = (props: Props) => {
...
-  useEffect(() => {
-    if (!activeCall) {
-      return;
-    }
-    startForegroundService();
-    return () => {
-      stopForegroundService();
-    };
-  }, [activeCall]);
+  usePushRegisterEffect(client);
```
vanGalilea, line 73: Same here, can we move calling `usePushRegisterEffect` to our StreamVideo to make it future-proof?

packages/react-native-sdk/jest-setup.ts
```diff
@@ -23,6 +23,11 @@ const mockedDevices = [
   },
 ];
 
+// Mock the notifee module using the mock provided by @notifee/react-native itself
+jest.mock('@notifee/react-native', () =>
+  require('@notifee/react-native/jest-mock'),
+);
```
vanGalilea, line 29: Excellent, can we try to mock this more elegantly via` __mocks__` folder in the root of the SDK?

I.e. create a file `__mocks__/@notifee/react-native.ts` with content:

```
import mock from '@notifee/react-native/jest-mock';

export default mock
```

packages/react-native-sdk/src/components/ActiveCall.tsx
```diff
@@ -38,6 +39,8 @@ export const ActiveCall = (props: ActiveCallProps) => {
   const activeCallRef = useRef(activeCall);
   activeCallRef.current = activeCall;
 
+  useAndroidForegroundCallEffect();
```
vanGalilea, line 42: Can we move to StreamVideo?

vanGalilea (changes requested):

packages/react-native-dogfood/src/components/ActiveCallComponent.tsx
```diff
@@ -1,37 +1,16 @@
...
-  useEffect(() => {
-    if (!call) {
-      return;
-    }
-    startForegroundService();
-    return () => {
-      stopForegroundService();
-    };
-  }, [call]);
+  useAndroidForegroundCallEffect();
```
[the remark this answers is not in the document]
santhoshvai: not yet though.. to do this we need to remove StreamCall component and merge it to StreamVideo

which is a refactor we can do as a separate task.. here the focus is PN

packages/react-native-dogfood/src/components/VideoWrapper.tsx
```diff
@@ -36,6 +37,8 @@ export const VideoWrapper = ({ children }: PropsWithChildren<{}>) => {
     },
   });
 
+  usePushRegisterEffect(videoClient);
```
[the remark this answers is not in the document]
santhoshvai: usePushRegisterEffect must be called by integrator if and only if they want push notifications and as they initiate the client.

Some integrators may not use push notification since we ask for payment for it

packages/react-native-sdk/src/components/ActiveCall.tsx
santhoshvai: answered above

packages/react-native-dogfood/src/components/VideoWrapper.tsx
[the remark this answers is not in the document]
santhoshvai: So its basically an optional call.. but if you call this hook and don't have the optional dependencies setup we throw an error. Making the development life clear and easy fixing potential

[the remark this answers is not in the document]
santhoshvai: With v6 chat we can introduce a similar hook for chat.. just an idea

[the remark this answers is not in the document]
santhoshvai: actually on second thoughts, we can do this.. we can check if the push config is present and then call the hook..

So basically if push config is set -> push is enabled in SDK else it is not enabled

will make a commit for that

packages/react-native-sdk/src/utils/push/config.ts
```diff
@@ -0,0 +1,11 @@
+import { StreamVideoRN } from '../StreamVideoRN';
+
+export const getPushConfig = () => {
+  const pushConfig = StreamVideoRN.getConfig().push;
+  if (pushConfig === undefined) {
+    throw Error(
+      'No config for push notifications have been set in StreamVideoRN.config. So push notifications cannot be enabled for this app',
+    );
```
vanGalilea, line 8: Discussion: what if I have a meeting app and I don't want to set PN?
Don't get me wrong, I like the idea, that people won't be able to miss that

packages/react-native-sdk/src/utils/push/libs.ts
```diff
@@ -0,0 +1,50 @@
...
+} catch (e) {}
+
+try {
+  voipPushNotification = require('react-native-voip-push-notification').default;
+} catch (e) {}
+
+export function getCallKeepLib() {
+  if (!callkeep) {
+    throw Error(
+      "react-native-callkeep library is not installed. Please install it using 'yarn add react-native-callkeep' or 'npm install react-native-callkeep'",
```
vanGalilea, line 28: Maybe here also send to https://github.com/react-native-webrtc/react-native-callkeep#Installation more future resistant

packages/react-native-sdk/jest-setup.ts
vanGalilea: It's not about not following their docs. It's about having a more accessible and common way to do the same thing.

If it works I will raise a PR to their docs, because that's the way it should've been.

- the [__mocks__](https://jestjs.io/docs/manual-mocks#mocking-node-modules) folder is meant for auto mocking your modules.
- jest-setup.ts is meant for [global](https://jestjs.io/docs/configuration/#setupfiles-array) setup

So not elegance, yet correctness and using configuration as planned by Jest

vanGalilea:
- when app is closed (swipe up to close) I'm not receiving the call notification

packages/react-native-sdk/src/utils/StreamVideoRN.ts
```diff
@@ -12,24 +13,101 @@ type StreamVideoConfig = {
    * Note: This is an array of reactions that is rendered in the Reaction list.
    */
   supportedReactions: StreamReactionType[];
+  /**
+   * The configuration to be used for push notifications.
+   * If undefined, push notifications will not be enabled for the app
+   * @default undefined
+   */
+  push?: {
+    ios_appName: string;
```
vanGalilea, line 22: let's change to 

```
push: {
  ios: {
    appName: string;
    ...
  }
}

```

packages/react-native-sdk/src/utils/StreamVideoRN.ts
```diff
@@ -12,24 +13,101 @@ type StreamVideoConfig = {
...
+     * @example "production-apn-video" or "staging-apn-video" based on the environment
+     */
+    ios_pushProviderName: string;
+    /** The callback that is called when a call is accepted, used for navigation */
+    navigateAcceptCall: () => void;
+  };
+  /**
+   * The notification channel to keep call alive in the background for Android using a foreground service.
+   */
+  android_foregroundServiceChannel: AndroidChannel;
```
vanGalilea, line 70:
```
{
  push: ...,
  foregroundService: {
    serviceChannel: string
   ...
  }
}
```

packages/react-native-sdk/src/providers/ActiveCallManager/index.tsx
```diff
@@ -0,0 +1,17 @@
...
+import { useProcessPushCallEffect } from './useProcessPushCallEffect';
+
+/**
+ * A renderless component that does two things
+ * 1. keeps the call alive in background in Android
+ * 2. processes the incoming call push notification
+ * This component must be a child of StreamVideo
+ * @internal
+ */
+export const ActiveCallManager = (): React.ReactElement | null => {
```
khushal87, line 12: Why can't we call these hooks directly in the ActiveCall component? A component simply returning null and call hooks internally doesn't look good to me, may be convert it to a use `useActiveCall`.

packages/client/src/StreamVideoClient.ts
```diff
@@ -271,13 +271,15 @@ export class StreamVideoClient {
    *
    * @param type the type of the call.
    * @param id the id of the call, if not provided a unique random value is used
+   * @param {boolean} [ringing] whether the call should be created in the ringing state.
    */
-  call = (type: string, id: string) => {
+  call = (type: string, id: string, ringing?: boolean) => {
```
santhoshvai, line 276: necessary change for creating a new call in RN when app is awake from dead through push ringing but WS would not send call.ring event

packages/client/src/StreamVideoClient.ts
```diff
@@ -195,22 +195,22 @@ export class StreamVideoClient {
...
+        await prevCall?.leave();
+        // we create a new call
+        const theCall = new Call({
+          streamClient: this.streamClient,
+          type: call.type,
+          id: call.id,
+          members,
+          clientStore: this.writeableStateStore,
+          ringing: true,
+          metadata: prevMetadata,
```
santhoshvai, line 210: the change is adding prevMetadata, so that auto drop do not happen due to change from undefined metadata timeout  to new metadata timeout

packages/react-native-sdk/jest-setup.ts
santhoshvai: I tested it and It did not work..

I understand its a common way, but doesn't work notifee I guess when you import.. 

I suggest that you try it out if we really want to achieve this. This change did not work..

maybe there is something at play with the change in jest.config.ts

you can see https://github.com/invertase/notifee/tree/main/packages/react-native for details

packages/react-native-sdk/src/providers/StreamCall.tsx
```diff
@@ -57,22 +61,67 @@ export const StreamCall = ({
...
+  }, [activeCall]);
+
   // Effect to create a new call with the given call id and type if the call doesn't exist
   useEffect(() => {
-    if (!videoClient) return;
+    if (!videoClient) {
+      return;
+    }
 
     if (callId && callType && !activeCall) {
```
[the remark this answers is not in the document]
santhoshvai: how? I dont get it .. btw this change was from eslint not related to this PR

packages/react-native-sdk/src/providers/ActiveCallManager/index.tsx
vanGalilea: Can we reply before resolving @khushal87 has a question 👀

packages/react-native-sdk/src/providers/StreamCall.tsx
```diff
@@ -57,22 +57,30 @@ export const StreamCall = ({
...
   });
 
   // Effect to create a new call with the given call id and type if the call doesn't exist
   useEffect(() => {
-    if (!videoClient) return;
+    if (!videoClient) {
+      return;
+    }
 
     if (callId && callType && !activeCall) {
```
vanGalilea, line 75:
```suggestion
    if (!videoClient || !callId || !callType || activeCall) {
      return;
    }
```

packages/react-native-sdk/src/providers/StreamCall.tsx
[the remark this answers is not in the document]
vanGalilea: I suggested a snippet below, I get your frustration ;)

packages/react-native-sdk/jest-setup.ts
vanGalilea: Fair enough, tnx for trying ;)

vanGalilea (approved):

packages/react-native-sdk/src/providers/ActiveCallManager/index.tsx
santhoshvai: Now its moved it to StreamCall, we had a discussion on slack about not keeping business logic in ActiveCall component to allow for customisation so did not move it there..

since the comment is valid but it was moved to StreamCall I pressed on resolve.. apologies for the confusion if any

PR #627, opened 2023-06-12
feat: User connect and types

szuperaz:
## Fix the problem with user type definitions
- `User` type is used for `connectUser`
- `OwnUserResponse` is used for `useConnectedUser`

## Update `connectUser` signature

- Regular/guest/anon authentication all done using `connectUser`, `type` param is used to tell if we want to connect a regular, guest or anonymous user
- Guest users are now created by `connectUser`, no need to create them beforehand

Docs update will be in separate PR

### DIFF
--- a/packages/client/src/StreamVideoClient.ts
+++ b/packages/client/src/StreamVideoClient.ts
@@ -27,6 +27,7 @@ import type {
   StreamClientOptions,
   TokenOrProvider,
   User,
+  UserWithId,
 } from './coordinator/connection/types';
 
 /**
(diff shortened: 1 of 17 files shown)


oliverlaz (approved):

szuperaz:
@khushal87, @santhoshvai, @vanGalilea please verify that I didn't break RN

khushal87:
> @khushal87, @santhoshvai, @vanGalilea please verify that I didn't break RN

On it.

khushal87 (approved):
Hey @szuperaz, the RN SDK changes are fine and nothing breaks except the page renders twice because the `connectUser` API was called later. But this would probably be fixed once we remove the usage of `useCreateStreamVideoClient`.

PR #634, opened 2023-06-13
perf(react-sdk): add "visibilitychange" event handler

arnautov-anton:
Also:
- fixes issue where PIP participant becomes viewport "invisible" and freezes PIP video

### DIFF
--- a/packages/react-sdk/src/core/hooks/useTrackElementVisibility.ts
+++ b/packages/react-sdk/src/core/hooks/useTrackElementVisibility.ts
@@ -1,4 +1,4 @@
-import { useEffect } from 'react';
+import { useEffect, useMemo, useRef } from 'react';
 import { ViewportTracker, VisibilityState } from '@stream-io/video-client';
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/core/hooks/useTrackElementVisibility.ts
```diff
@@ -13,32 +13,89 @@ export const useTrackElementVisibility = <T extends HTMLElement>({
...
+    }),
+    [call, sessionId],
+  );
+
+  useEffect(() => {
+    if (!call || !trackedElement) return;
+
+    const handleVisibilityChange = () => {
+      const isDocumentVisible = document.visibilityState === 'visible';
+      const isPIP = trackedElement.contains(document.pictureInPictureElement);
```
MartinCupela, line 61: What does "PIP" mean? Could you add a comment @arnautov-anton ? I understand that it is picture in picture

MartinCupela (approved):

PR #637, opened 2023-06-13
feat: Update React sample apps for new connect flow

szuperaz:


packages/react-bindings/src/contexts/StreamVideoContext.tsx
```diff
@@ -39,12 +39,6 @@ export const StreamVideo = ({
   const prevClient = useRef<StreamVideoClient | undefined>(undefined);
 
   useEffect(() => {
```
szuperaz, line 41: I made a mistake with the connect-disconnect, I only noticed when testing the tutorial

oliverlaz (approved):

PR #639, opened 2023-06-14
refactor(react-native): remove useCreateStreamVideoClient and dogfood app refactor to have single StreamVideo on the root

khushal87:
This PR focus on removing `useCreateStreamVideoClient` from the RN SDK.
After the new client changes, we can have a single instance of `StreamVideo` in our dogfood app on the root, and the anonymous and guest login works fine with the change. This improves the flow of the app, as users no longer need to log in every time they change mode. App Mode can now be changed easily.

The starter kits and cookbook are also updated to prevent build errors.

Note: The documentation will be updated once the PR is merged.


packages/react-native-call-starter-kit/src/components/VideoWrapper.tsx
vanGalilea, line 1: remove meet starter kit

packages/react-native-dogfood/src/components/VideoWrapper.tsx
```diff
@@ -22,19 +26,19 @@ export const VideoWrapper = ({ children }: PropsWithChildren<{}>) => {
...
-    },
-  });
+  const [videoClient] = useState<StreamVideoClient>(
+    () =>
+      new StreamVideoClient({
+        apiKey: STREAM_API_KEY,
+        user,
+        tokenProvider,
+      }),
+  );
```
vanGalilea, line 41: Why does videoClient needs to be state? if instantiated once?

packages/react-native-call-starter-kit/src/components/VideoWrapper.tsx
khushal87: Already done

packages/react-native-dogfood/src/components/VideoWrapper.tsx
khushal87: Converted it to useRef

vanGalilea (approved):

packages/react-native-dogfood/src/components/VideoWrapper.tsx
oliverlaz: Apologies for jumping in here but I think you should keep the client within a useState variable.

The reasoning: the client should be typically instantiated only once via `useState(() => client)` factory. In case of dependency change, there should be an `useEffect` that would dispose of the previous client instance, set up a new one and eventually update the state. In case you don't want to react on dependency changes, you can omit the `useEffect` part.

The solution with `useRef` causes new StreamVideoClient instances to be created on every render and then immediately discarded with no way to have them properly disposed of.

PR #645, opened 2023-06-14
fix: do not send ring to backend if call was not created by the user

santhoshvai:
Without this fix, while ringing the BE keeps on sending call.ring events, it causes an infinite loop until joined/rejected as below 

```
call.ring -> send ring to BE -> call.ring -> send ring to BE call.ring -> send ring to BE  ......... until call is joined or rejected
```

Now if ring param is set.. only the ringing rx subject is set to true and we don't send to BE. This ensures that we have a local state of ringing.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -469,21 +469,22 @@ export class Call {
    * @param params.ring if set to true, a `call.ring` event will be sent to the call members.
    * @param params.notify if set to true, a `call.notification` event will be sent to the call members.
    * @param params.members_limit the members limit.
+   * @param makeRinging if set to true, the `state.callingState` will change to 'ringing'.
    */
-  get = async (params?: {
-    ring?: boolean;
-    notify?: boolean;
-    members_limit?: number;
-  }) => {
+  get = async (
+    params?: {
+      ring?: boolean;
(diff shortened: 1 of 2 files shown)


packages/client/src/Call.ts
```diff
@@ -477,7 +477,10 @@ export class Call {
   }) => {
     const response = await this.streamClient.get<GetCallResponse>(
       this.streamClientBasePath,
-      params,
+      {
+        ...params,
+        ring: this.isCreatedByMe ? params?.ring : false,
```
oliverlaz, line 482: I think this "decision" needs to be made at a higher level. I don't think having an API that receives one value but sends another is a good idea.

Can you lift this to the call sites of this method?

santhoshvai: I agree, will do that, does the function signature work

```
get(apiParams, makeRinging = false)

So in call site,

call.get(undefined, true) would lift the decision to the callsite
```

oliverlaz: I'd rather have:
```ts
if (makeRinging) {
  call.get({ ring: true });
}

// or

call.get({ ring: makeRinging });
```

oliverlaz (approved):

PR #659, opened 2023-06-16
docs(react sdk): update audio rooms tutorial

MartinCupela:
## Demo changes
- load the call data on when loading the room detail page directly
- group rooms by using filter conditions for API calls instead of filtering client-side
- add pagination
- extract client init logic to a separate provider component
- pass tokenProvider function directly to client constructor
- add loading indicator to button awaiting speak permission approval + reset permission awaiting state when leaving the room
- re-fetch the left call to re-register all the call event handlers


sample-apps/react/audio-rooms/src/components/RoomList/RoomListing.tsx
```diff
@@ -0,0 +1,158 @@
...
+        const newCall = client.call(callResponse.type, callResponse.id);
+        await newCall.get();
+        setCalls((prevCalls) => [newCall, ...prevCalls]);
+      } else if (liveState === 'upcoming') {
+        setCalls((prevCalls) =>
+          prevCalls.filter((c) => c.cid !== callResponse.cid),
+        );
+      }
+    };
+    client.on('call.live_started', handleGoLive);
```
oliverlaz, line 90:
```suggestion
    return client.on('call.live_started', handleGoLive);
```

Makes sure this event is cleaned up whenever dependencies change or this component is unmounted.

sample-apps/react/audio-rooms/src/components/RoomList/RoomListing.tsx
```diff
@@ -0,0 +1,158 @@
...
+      const [type, id] = call_cid.split(':');
+      if (liveState === 'ended') {
+        const newCall = client.call(type, id);
+        await newCall.get();
+        setCalls((prevCalls) => [newCall, ...prevCalls]);
+      } else if (liveState === 'live') {
+        setCalls((prevCalls) => prevCalls.filter((c) => c.cid === call_cid));
+      }
+    };
+    client.on('call.ended', handleCallEnded);
```
oliverlaz, line 106:
```suggestion
    return client.on('call.ended', handleCallEnded);
```

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,527 +3,961 @@ title: Audio Room Tutorial
...
+Routing in the app is done by using 3rd party library [react-router-dom](https://reactrouter.com/en/main)
 :::
 
-First, we can collect the API key from our `.env` file by simply importing it like this:
+### User Authentication
 
-```
-const apiKey = import.meta.env.VITE_STREAM_API_KEY as string;
-```
+The first important step is to initialize the Stream client. Therefore, we need an API key and a token provider.
```
oliverlaz, line 68:
```suggestion
The first important step is to initialize the Stream Video client. Therefore, we need an API key and a token provider.
```

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,527 +3,961 @@ title: Audio Room Tutorial
...
+export const RoomListing = ({ liveState }: { liveState: RoomLiveState }) => {
+  const client = useStreamVideoClient();
+  const [calls, setCalls] = useState<Call[]>([]);
+  // omitted code initiating state variables ...
+  const nextPage = useRef<string | undefined>();
+
+  const loadCalls = useCallback(async () => {
+    if (!client) return;
+    const result = await client.queryCalls({
+      ...QUERY_CALLS_PARAMS,
```
oliverlaz, line 319: I think we should inline the parameters here. IMO, having these constants reduces the readability of the tutorial, as the reader (me) has to jump back and forth on the page.

My experience:
- I scan the QUERY_CALLS_PARAMS value, but I'm not paying too much attention to it as I lack the context and I proceed forward
- a couple of lines below, I see it being used, and now I have to scroll back and actually see what it really means when used in `queryCalls` context...

oliverlaz (approved):
The app looks good and I think we are good to merge this PR.

There are few small remarks regarding the content of the tutorial that IMO, may be addressed in a follow up PR :)

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,527 +3,961 @@ title: Audio Room Tutorial
...
+        apiKey,
+        // highlight-next-line
+        tokenProvider,
+        user: {
+          id: user.id,
+          image: user.imageUrl,
+          name: user.name,
+        },
+      }),
+    );
```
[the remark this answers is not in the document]
szuperaz: The user disconnection should be handled by the `StreamVideo` https://github.com/GetStream/stream-video-js/blob/1e7815773806286ba1cba32c5b7a2b02c4c0fbc4/packages/react-bindings/src/contexts/StreamVideoContext.tsx#L53-L55 is that not enough in this case?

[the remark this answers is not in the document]
oliverlaz: ok... my bad, I didn't check this. Yes, this should be enough

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -3,527 +3,961 @@ title: Audio Room Tutorial
...
-3. `listeners`: can join the call and listen but not share their audio; they can ask for permission to speak
+Now that we are done with the user authentication, we can start building the app. The first thing we'll do is to allow users to create an audio room where others can then join.
 
-The last object in our type is the `Call` itself which is the Stream object with all the attached logic.
+We will populate `custom` attribute of a `Call` object with the following data:
 
-Now, after collecting the necessary info (`title` and `description`, maybe a list of hosts) we can create a call on the backend using the following call:
+:::note
+We keep an array of user ids called `speakerIds` in order to make it visible to all the room participants, who has been granted permission to speak.
+:::
```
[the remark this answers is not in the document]
MartinCupela: We should, but the BE does not provide the data yet. Still WIP on their side

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: @oliverlaz I agree. I have inlined the code.

PR #660, opened 2023-06-16
feat(react-demo): connection notification

zwaardje:
Add connection notifications to indicated if the connection is lost or trying to reconnect.


zwaardje:
Already in the code

PR #670, opened 2023-06-19
docs: Update custom call controls

szuperaz:


szuperaz:
> I am thinking, that we should keep using the function exposed by the context to keep things unified

What do you mean by unified? I'd think it's unified if we show the same code to integrators that we use in the built-in components. Especially if that's easier then what we had previously.

MartinCupela:
> > I am thinking, that we should keep using the function exposed by the context to keep things unified
> 
> What do you mean by unified? I'd think it's unified if we show the same code to integrators that we use in the built-in components. Especially if that's easier than what we had previously.

I see that it has been replaced everywhere. Please ignore my remark.

PR #672, opened 2023-06-20
fix(react-native): google sign in dogfood app

khushal87:


sample-apps/react-native/dogfood/src/screens/LoginScreen.tsx
```diff
@@ -36,7 +36,9 @@ const LoginScreen = () => {
 
   const loginHandler = async () => {
     try {
-      const _username = localUserName.replace(/\s/g, '-');
```
santhoshvai, line 39: why remove this.. 

empty spaces need to be replaced too

sample-apps/react-native/dogfood/src/screens/LoginScreen.tsx
```diff
@@ -52,8 +54,11 @@ const LoginScreen = () => {
     try {
       await GoogleSignin.hasPlayServices();
       const userInfo = await GoogleSignin.signIn();
+      const username = userInfo.user.email
+        .replace(/[^_\-0-9a-zA-Z@]/g, '_')
+        .replace('@getstream_io', '');
```
santhoshvai, line 59: Looks like both the logic can be put in a common function

Also I dont understand why `@getstream_io` need to be replaced, its compatible with streamchat

santhoshvai: but as short cut  I understand

Ideally it is

```
id: khushalagarwal@getstream.io
name: khushalagarwal
```

Now it is

```
id: khushalagarwal
name: khushalagarwal
```

The drawback is that from dashboard we would not be able to find easily if it was google login or not..

but we can live with this

szuperaz: Yes, there is a probably a more elegant way to do the replacement. We were asked to remove the `@` by Tommaso.

khushal87: https://getstream.slack.com/archives/C04ATV49DU3/p1687248654192859?thread_ts=1687248366.389239&cid=C04ATV49DU3

santhoshvai (approved):

vanGalilea (approved):

PR #675, opened 2023-06-20
fix(react-native): fix pronto link in dogfood app

khushal87:


sample-apps/react-native/dogfood/App.tsx
```diff
@@ -77,12 +77,14 @@ const StackNavigator = () => {
   useEffect(() => {
     const subscription = prontoCallId$.subscribe((prontoCallId) => {
       if (prontoCallId) {
-        setState({ appMode: 'Meeting' });
+        if (username && userImageUrl) {
```
santhoshvai, line 80: its better to move this if three lines before.. so that we don't subscribe if the user stuff is not available

khushal87: Done

sample-apps/react-native/dogfood/ios/Podfile.lock
```diff
@@ -955,4 +955,4 @@ SPEC CHECKSUMS:
 
 PODFILE CHECKSUM: 41e93cd2e8bce3f44104edb48b9b364a0e34bd71
 
-COCOAPODS: 1.12.1
+COCOAPODS: 1.11.3
```
santhoshvai, line 958: do a bundle install please.. you are changing the cocoapods version

PR #678, opened 2023-06-20
feat(react sdk): prefer URL credentials in audio rooms demo app

MartinCupela:
## Todo:

- verify why the the client does not wait with HTTP requests until a new user is created on the first load


sample-apps/react/messenger-clone/data/users.json
oliverlaz, line 1: Can you revert these changes? Tokens should never be committed to the repository. It is a good thing they are expired already 😅

oliverlaz (approved):

sample-apps/react/messenger-clone/data/users.json
MartinCupela: Not sure how they appeared there

PR #681, opened 2023-06-20
feat: DF cosmetic tweaks and invite feat

vanGalilea:
- zIndex update throughout app
- participants' info view and moderation resting and refactor
- share call link feat
- fallback to avatar on 1:1 calls


vanGalilea:
Self reviewed

PR #684, opened 2023-06-21
refactor(react-native): remove call creation logic from StreamCall component

khushal87:


packages/react-native-sdk/docusaurus/docs/reactnative/01-basics/01-tutorial.mdx
```diff
@@ -409,19 +414,28 @@ const MeetingScreen = (props: Props) => {
...
+  const call = useMemo<Call | undefined>(() => {
+    if (!videoClient) {
+      return undefined;
+    }
+    return videoClient.call(callType, callId);
+  }, [callId, callType, videoClient]);
+
+  if (!call) {
+    return;
+  }
```
santhoshvai, line 426: instead of this,

I think its better to do..

```
   <StreamVideo client={videoClient}>
      {!!call && <StreamCall call={call}>{children}</StreamCall>}
    </StreamVideo>
```

So that for integrators they would automatically know that call is not needed to render StreamVideo

santhoshvai (approved):

packages/react-native-sdk/docusaurus/docs/reactnative/01-basics/01-tutorial.mdx
khushal87: Did this because of an early exit that we all love, but I am happy to change it.

PR #685, opened 2023-06-21
fix(react-native): toggle camera bug and simplify the logic

khushal87:


vanGalilea (approved):

packages/react-native-sdk/src/hooks/useCallControls.tsx
```diff
@@ -142,13 +142,15 @@ export const useCallControls = () => {
...
-          ? device.facing === 'front'
-          : device.facing === 'environment'),
-    );
+    const videoDevice = videoDevices.find((device) => {
+      // Check to only switch between video devices
+      if (device.kind !== 'videoinput') {
+        return;
+      }
+      return !isCameraOnFrontFacingMode
+        ? device.facing === 'front'
```
vanGalilea, line 151: we should use enums in the future here

PR #692, opened 2023-06-22
refactor(react-native): change CallParticipantsView to CallParticipantsGridView and conditionally render the grid and spotlight component

khushal87:
Done as a step to remove the ActiveCall component in the future.


packages/react-native-sdk/src/components/CallParticipantsGridView.tsx
khushal87, line 1: INFO: No change from my side on this code. Previous CallParticipantsView is moved to CallParticipantsGridView.

vanGalilea (approved):

PR #704, opened 2023-06-23
feat: Add user that mutes others in `CallUserMuted` event

guerinoni:
Signed-off-by: Federico Guerinoni <federico.guerinoni@getstream.io>


oliverlaz (approved):
Thanks!

PR #709, opened 2023-06-24
docs: Tutorial rewrite

szuperaz:
Based on:
- https://staging.getstream.io/video/docs/android/basics/quickstart/
- https://staging.getstream.io/video/docs/android/tutorials/video-calling/


sample-apps/react/stream-video-react-tutorial/src/main.tsx
```diff
@@ -1,9 +1,6 @@
-import React from 'react';
 import ReactDOM from 'react-dom/client';
 import App from './App';
 
 ReactDOM.createRoot(document.getElementById('root') as HTMLElement).render(
-  <React.StrictMode>
-    <App />
-  </React.StrictMode>,
```
szuperaz, line 8: I turned off `StrictMode` because we have a problem there:
- Since [this effect](https://github.com/GetStream/stream-video-js/blob/43506301df19b21b327bcc060891a8dcd991ad03/packages/react-bindings/src/contexts/StreamVideoContext.tsx#L41-L48) runs twice we connect-disconnect-reconnect the user
- In the disconnect we [leave all calls](https://github.com/GetStream/stream-video-js/blob/43506301df19b21b327bcc060891a8dcd991ad03/packages/client/src/store/stateStore.ts#L23-L26) but in the connect we won't rejoin them (which makes sense we don't)
- The result will be that we get disconnected from the call, and we don't rejoin it

oliverlaz (approved):

PR #722, opened 2023-06-27
feat(react-dogfood): add "beforeunload" event handler

arnautov-anton:
This PR adds "beforeunload" handler which prompts users upon website reload or leave to prevent unwanted call leaves.

Known limitations:
- `event.returnValue` is deprecated, documentation suggests using `event.preventDefault()` which does not work - documentation also suggest using return value - which (surprise, surprise) also does not work ([compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#browser_compatibility))
- does not respect navigation changes (NextJS router ~solutions work~ - but prompt you twice)

Read more: https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event#compatibility_notes

### DIFF
--- a/sample-apps/react/react-dogfood/components/MeetingUI.tsx
+++ b/sample-apps/react/react-dogfood/components/MeetingUI.tsx
@@ -33,7 +33,11 @@ import {
   UnreadCountBadge,
 } from '.';
 import { ActiveCallHeader } from './ActiveCallHeader';
-import { useKeyboardShortcuts, useWatchChannel } from '../hooks';
+import {
+  useBeforeUnload,
+  useKeyboardShortcuts,
+  useWatchChannel,
+} from '../hooks';
 import { DEFAULT_LAYOUT, getLayoutSettings, LayoutMap } from './LayoutSelector';
 import { Stage } from './Stage';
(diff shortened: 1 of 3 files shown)


MartinCupela:
I tested this hook and it did not work. Looking into the code, we could add `window.confirm(message)` to already existing hook in `MeetingUI.tsx`:

```ts
  useEffect(() => {
    const handlePageLeave = async () => {
      if (!window.confirm('Call in progress, are you sure you want to leave?')) return;
      if (
        activeCall &&
        [CallingState.JOINING, CallingState.JOINED].includes(callState)
      ) {
        await activeCall.leave();
      }
    };
    router.events.on('routeChangeStart', handlePageLeave);
    return () => {
      router.events.off('routeChangeStart', handlePageLeave);
    };
  }, [activeCall, callState, router.events]);
```

Or better we could add `window.confirm(message)` to `onLeave` callback passed in the same file:

```ts
  const onLeave = useCallback(async () => {
    if (!window.confirm('Call in progress, are you sure you want to leave?'))
      return;
    setShow('loading');
    try {
      await router.push('/');
    } catch (e) {
      console.error(e);
      setShow('error-leave');
    }
  }, [router]);
```

PR #723, opened 2023-06-28
fix(react-native): vale lint issues

khushal87:


khushal87:
Self reviewed and Merged ✅

PR #730, opened 2023-06-28
chore: Make it possible to opt out from Sentry in the dogfood browser…

szuperaz:
… app


szuperaz:
I don't know if it's enough to turn off Sentry completely, as we also have it on server-side, but I guess we will see

oliverlaz (approved):

oliverlaz:
> I don't know if it's enough to turn off Sentry completely, as we also have it on server-side, but I guess we will see

I guess the majority of the traffic comes from the client side of things.

PR #733, opened 2023-06-28
fix(react-native): fix blank video/audio streams and muting state

santhoshvai:
Fixes the following issues: 

- muting state is now a global state.. (so when leave a call etc, it is still persisted).. also the persistence can be removed at any point by a component anywhere..
    how it should have been? muting state must be local to a lifecycle of call.. aka from lobbyview to active call, or incoming call to active call
- we initially publish streams without checking if call has been joined.. essentially creating a random race.. wherein we might randomly publish initially before the join happened and thereby streams wont be visible


packages/react-native-sdk/src/providers/MediaStreamManagement/index.tsx
vanGalilea, line 1: Very similar to the React [one](https://github.com/GetStream/stream-video-js/blob/4286d9855cc85c3f0f05d0d19b75e423c55591ce/packages/react-sdk/src/core/contexts/MediaDevicesContext.tsx#L234)

vanGalilea: Can't we know not benefit out of it and move it to the bindings?

packages/react-native-sdk/src/providers/MediaStreamManagement/useAudioPublisher.ts
```diff
@@ -0,0 +1,113 @@
...
+   * 1. joining the call from the lobby, and the video is not muted
+   * 2. reconnecting to the call with the video already published
+   * 3. when the video device is changed (this is handled by the dependency to publishVideoStream function)
+   */
+  useEffect(() => {
+    if (callingState === CallingState.JOINED) {
+      if (
+        (!initialAudioMuted && !initialPublishRun.current) ||
+        isPublishingAudio
+      ) {
```
vanGalilea, line 73: can we merge the if statements or take the early exit approach you already took above?

santhoshvai: done!

packages/react-native-sdk/src/providers/MediaStreamManagement/index.tsx
santhoshvai: No we cannot, it has more browser specific stuff.. 

like extra states for video inputs, listeners for device disconnection

santhoshvai: and we add native permission checkers

santhoshvai: also in web they use default device id for video.. which is not relevant for rn-webrtc

vanGalilea: 👍🏻

vanGalilea (approved):

packages/react-native-sdk/src/providers/MediaStreamManagement/useVideoPublisher.ts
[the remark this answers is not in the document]
santhoshvai: No in my view, reasons:

-- keeping close to react
-- we will need to do some special stuff for video in reconnection flow, like republishing when coming back to foreground
-- cleaner with separation, although its similar you can reuse any.. so separation makes it easier to maintain

vanGalilea:
Tested, no visible regression throughout main flow and E2E tests ✅

PR #737, opened 2023-06-29
feat(react-native): align call participants list with figma

santhoshvai:
The changes: https://www.loom.com/share/b1065f4762bd4bf88ebdb790fc21ad75?sid=9fd5f403-958a-4368-913d-c0d25ed53858

In figma, on 1:1 calls the video stream takes the full space. This pattern can be seen in 2:1 calls too. 4:1 and then 6:1. 

When its more than 6 participants we start to scroll.

This PR implements the above behavior.


vanGalilea:
Great PR!

vanGalilea (changes requested):
There is some regression

vanGalilea (approved):

PR #739, opened 2023-06-30
docs(react-sdk): add token snippet to audio rooms tutorial

MartinCupela:


packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -38,44 +44,90 @@ npm install @stream-io/video-react-sdk
...
+The credentials can then be passed to the `StreamVideoClient` constructor:
+
+```tsx
+import { useEffect, useState } from 'react';
+import {
+  ChildrenOnly,
+  StreamVideo,
+  StreamVideoClient,
+} from '@stream-io/video-react-sdk';
+
```
oliverlaz, line 88:
```suggestion
const apiKey = '<api-key>'; // or read from .env file
const userToken = '<user-token>';
const userId = '<user-id>';
const callId = '<call-id'>
```

Perhaps, we can skip the `.env` part entirely if these variables are initialized here. What do you think about this?

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -38,44 +44,90 @@ npm install @stream-io/video-react-sdk
...
+export const VideoClientProvider = ({ children }: ChildrenOnly) => {
+  // retrieve the API key, user and token...
+
+  const [client, setClient] = useState<StreamVideoClient>();
+
+  useEffect(() => {
+    if (!user) return;
```
szuperaz, line 95: I think `user` variable isn't initialized anywhere in the code

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
```diff
@@ -38,44 +44,90 @@ npm install @stream-io/video-react-sdk
...
+  }, [token, user]);
+
+  if (!client) return null;
+
+  return <StreamVideo client={client}>{children}</StreamVideo>;
+};
+```
+
+
+### Dynamic token retrieval
```
szuperaz, line 116: I don't think we need this part, given that we don't use dynamic tokens in the tutorial

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: it is mentioned in the comment. This is because we agreed, we would leave out the detail of how some parts are implemented.

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: We use them in the app.

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
szuperaz: But I think we should still present a code that compiles, it's just this: `user = {id: userId}`

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: I would not recommend something, that is not the best practice - meaning, including token in the src code :) Somebody could throw that into our face.

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: Then we would have to include token, api key and if we want to show additions to long code snippets, we would duplicate those snippets, just to add single or few lines.

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: I have removed this part

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: I have added the empty variables

packages/react-sdk/docusaurus/docs/React/02-tutorials/02-audio-room.mdx
MartinCupela: I have rewritten the tutorial, but some snippets keep being incomplete as they are huge. I tried to refactor som components by splitting them. Still there are many things to explain. Not sure how to do it in a more slick way

PR #740, opened 2023-07-01
fix: strict mode issue

szuperaz:
Another option, inspired by @arnautov-anton's one client-one user suggestion. 

- Remove react sample app
- Fix an issue with `assertCallJoined`
- `call.leave` now waits for ongoing `call.join` -> this way we can handle fast `call.join()` - `call.leave` cycles that happen with strict mode
- Client and call lifecycle now entirely handled by integrators


szuperaz:
Can someone help with the [yarn.lock](https://github.com/GetStream/stream-video-js/actions/runs/5486810131/jobs/9997391621?pr=740#step:6:131) issue? I know this error, but always forget the solution.

packages/client/src/Call.ts
```diff
@@ -1221,6 +1217,84 @@ export class Call {
     return this.publisher?.updateVideoPublishQuality(enabledRids);
   };
 
+  private handleOnTrack = (e: RTCTrackEvent) => {
+    const [primaryStream] = e.streams;
+    // example: `e3f6aaf8-b03d-4911-be36-83f47d37a76a:TRACK_TYPE_VIDEO`
+    const [trackId, trackType] = primaryStream.id.split(':');
+    this.logger('info', `Got remote ${trackType} track:`);
```
oliverlaz, line 1224: This is probably a bad merge. `handleOnTrack` now lives in `Subscriber`.

packages/client/src/StreamVideoClient.ts
```diff
@@ -126,30 +129,30 @@ export class StreamVideoClient {
    * @param token a token or a function that returns a token.
    */
   async connectUser(
-    user?: User,
-    token?: TokenOrProvider,
+    user: User,
+    token: TokenOrProvider,
```
oliverlaz, line 133: Should we store these values on the client instance?
```
this.user = user;
this.token = token;
```

packages/client/src/Call.ts
szuperaz: uh, thanks for catching that fixed

packages/react-bindings/src/contexts/StreamVideoContext.tsx
```diff
@@ -36,26 +36,9 @@ export const StreamVideo = ({
...
-    return () => {
-      prevClient.current
-        ?.disconnectUser()
-        .catch((error) => console.error(`Failed to disconnect`, error));
-    };
-  }, [client]);
-
+  if (!client) {
+    return null;
+  }
```
oliverlaz, line 41: I'd rather lift this to the apps. This way, it is a bit implicit.

packages/client/src/StreamVideoClient.ts
szuperaz: We need to have `user` and `token` as params in case someone does `connectUser` separate from `constructor`.  I think the API is cleaner if they are required than if they are optional.

packages/react-bindings/src/contexts/StreamVideoContext.tsx
szuperaz: I did this to avoid having to do this for all apps: 

```
{client && <StreamVideo>{call && <StreamCall></StreamCall>}</StreamVideo>}
```

It was lots of extra work since both `client` ans `call` are `undefined` initially

packages/client/src/StreamVideoClient.ts
oliverlaz: yes, that's fine. What I was suggesting is, we don't seem to update the instance variables:
```ts
const client = new StreamVideoClient({ user: 'oliver' });
client.connectUser('zita');
console.log(client.user): // oliver
```

packages/react-bindings/src/contexts/StreamVideoContext.tsx
oliverlaz: I worry if we leave it like this, we would be making a huge breaking change in case we stop accepting "undefined" `client` or `call` in those providers.

packages/client/src/StreamVideoClient.ts
szuperaz: I removed the instance variables since your comment, we don't need them anymore

packages/react-bindings/src/contexts/StreamVideoContext.tsx
szuperaz: Yes, if we change this, that'd be a breaking change. But since we don't have plans for doing that at the moment, I think it's better to live with this convenience than to remove it just because we might break this in the future.

MartinCupela (approved):

PR #744, opened 2023-07-03
feat: stop tracking permission requests in Call state

MartinCupela:
We stop tracking `call.permission_request` events in `Call` state to avoid the `useCallPermissionRequest` hook emitting repeatedly the same event on each re-render. That would lead to duplication of the same event in the component's local state of requests list.

### DIFF
--- a/packages/client/src/events/__tests__/call-permissions.test.ts
+++ b/packages/client/src/events/__tests__/call-permissions.test.ts
@@ -2,48 +2,12 @@ import { describe, expect, it } from 'vitest';
 import { CallState } from '../../store';
 import {
   watchCallGrantsUpdated,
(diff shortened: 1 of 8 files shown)


szuperaz:
@MartinCupela what's the way to reproduce the problem? I tested with the dogfood app on `main` and couldn't see duplicate permission requests. Is it a strict mode related problem?

szuperaz:
> @MartinCupela what's the way to reproduce the problem? I tested with the dogfood app on `main` and couldn't see duplicate permission requests. Is it a strict mode related problem?

@oliverlaz explained this in the sync meeting. I see the problem now.

szuperaz (approved):

oliverlaz (approved):

PR #745, opened 2023-07-03
chore: restructure comp

vanGalilea:
## Goal
This PR aims to align all UI components with other SDKs and future docs.

## Implementation
- Directory structure was changed
- Components were renamed
- Components sorted by 3 domains (call, participants and utility) and whether their internal or not


packages/react-native-sdk/src/components/call/CallParticipantsListView.tsx
vanGalilea, line 1: add View post fix

vanGalilea: add to index.ts to export

packages/react-native-sdk/src/components/participants/ParticipantsInfoBadge.tsx
vanGalilea, line 1: name file also ParticipantsInfoBadge

PR #746, opened 2023-07-03
feat: reconnection flow support for RN

santhoshvai:
## The main change

In `call.ts`, changed the window.eventlistener mechanism to detect online/offline to use the videoclient's connection change event. So that it works across RN and web

## Tested on

* ios device, android emulator, react web dogfood in desktop

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -103,7 +103,6 @@ import {
   StreamCallEvent,
 } from './coordinator/connection/types';
 import { getClientDetails } from './client-details';
-import { isReactNative } from './helpers/platforms';
 import { getLogger } from './logger';
 
 /**
@@ -134,16 +133,6 @@ export class Call {
    * The state of this call.
    */
   readonly state = new CallState();
-
-  private rejoinPromise: (() => Promise<void>) | undefined;
-
-  /**
-   * A promise that exposes the reconnection logic
-   * The use-case is for the react-native platform where online/offline events are not available in the window
-   */
-  get rejoin(): (() => Promise<void>) | undefined {
-    return this.rejoinPromise;
-  }
 
   /**
    * Flag indicating whether this call is "watched" and receives
@@ -387,7 +376,6 @@ export class Call {
     if (callingState === CallingState.LEFT) {
       throw new Error('Cannot leave call that has already been left.');
     }
-    this.rejoinPromise = undefined;
 
     if (this.ringing) {
       // I'm the one who started the call, so I should cancel it.
@@ -685,7 +673,7 @@ export class Call {
       // we shouldn't be republishing the streams if we're migrating
       // as the underlying peer connection will take care of it as part
       // of the ice-restart process
-      if (localParticipant && !isReactNative() && !migrate) {
+      if (localParticipant && !migrate) {
         const {
           audioStream,
           videoStream,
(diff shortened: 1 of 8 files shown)


packages/client/src/Call.ts
```diff
@@ -135,16 +134,6 @@ export class Call {
...
-  private rejoinPromise: (() => Promise<void>) | undefined;
-
-  /**
-   * A promise that exposes the reconnection logic
-   * The use-case is for the react-native platform where online/offline events are not available in the window
-   */
-  get rejoin(): (() => Promise<void>) | undefined {
-    return this.rejoinPromise;
-  }
-
```
oliverlaz, line 147: 🚀 🚀 🚀

packages/client/src/Call.ts
```diff
@@ -766,38 +750,44 @@ export class Call {
...
+      'connection.changed',
+      (e) => {
+        if (e.type !== 'connection.changed') return;
+        if (!e.online) return;
+        unsubscribeOnlineEvent();
+        if (
+          [CallingState.JOINED, CallingState.JOINING].includes(
+            this.state.callingState,
+          )
+        ) {
```
oliverlaz, line 763: I'm not sure about this... shouldn't we run this code only when the call state is `offline`?

packages/client/src/Call.ts
```diff
@@ -766,38 +750,44 @@ export class Call {
...
+    );
+    const unsubscribeOfflineEvent = this.streamClient.on(
+      'connection.changed',
+      (e) => {
+        if (e.type !== 'connection.changed') return;
+        if (e.online) return;
+        unsubscribeOfflineEvent();
+        this.state.setCallingState(CallingState.OFFLINE);
+      },
+    );
```
oliverlaz, line 785: Perhaps you can merge the two event handlers in one

santhoshvai: We unfortunately cant.. the event listeners have to be unsubscribed as soon as it is fired once so that it doesn't live until the entire life of the call until call is left.

packages/client/src/Call.ts
santhoshvai: In react native, the sfu close event comes very fast when going offline (airplane mode).. so rejoin starts before we can inform to call.ts that we are offline. Hence the state changes from "offline" to ""reconnecting_failed"

Thats why I had to do this instead. 

Alternatively we could choose to not set "reconnecting_failed" when offline

what is better?

oliverlaz: What is the SFU "close" code in this case? Reconnection should start only if the WS is terminated abruptly (1006).

packages/client/src/Call.ts
oliverlaz: oh, ok. Understood.

packages/client/src/Call.ts
santhoshvai: As discussed, I have changed this condition to 

```
        if (
          callingState === CallingState.OFFLINE ||
          callingState === CallingState.RECONNECTING_FAILED
        ) { rejoin()... }
```

oliverlaz (approved):

sample-apps/react-native/dogfood/src/components/ActiveCallHeader.tsx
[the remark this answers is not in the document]
santhoshvai: copied the name from react sdk, but yeah I agree that notification is a better name

will change it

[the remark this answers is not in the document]
santhoshvai: done

vanGalilea (approved):

PR #751, opened 2023-07-05
fix(client): prevent firing API requests until connection.ok WS even is received

MartinCupela:


packages/client/src/coordinator/connection/client.ts
```diff
@@ -399,11 +393,6 @@ export class StreamClient {
     this.anonymous = true;
     await this._setToken(user, tokenOrProvider, this.anonymous);
 
-    if (this.resolveConnectPromise) {
-      this.resolveConnectPromise();
-      this.waitForConnectPromise = undefined;
-      this.resolveConnectPromise = undefined;
-    }
```
oliverlaz, line 406: @szuperaz, correct me if I'm wrong, but I think this code needs to say here in order to "unblock" network requests for anon users (`call/{type}/{id}/join`).

MartinCupela: Is it possible not to call `startWaitingForConnection` in `StreamVideoClient` constructor if client is initialized for anon user?

szuperaz: I don't know, something seems weird for me, I don't understand why this would only cause errors to new users

szuperaz: But the promise that's modified, isn't the promise that should wait for the authentication to be done 😄 we have way too many promises

szuperaz: I see it now.......but I think there is an easier way to do this, let me quickly create a PR for that

szuperaz: @MartinCupela can you please check if this fix works: https://github.com/GetStream/stream-video-js/pull/752? Or I can do that as well, if you tell me the steps

PR #755, opened 2023-07-05
feat: Remove ringing flag from call client

szuperaz:
## Problem

 - Having the `ringing` flag in `client.call` can be misleading to integrators, for example they could do: `client.call('default', '123', true),join({ring: false})`
 - The flag also creates API mismatch with iOS and Android

## Solution

`client.onRingingCall` will create a ringing call from a push notification. The naming is based on: https://github.com/GetStream/stream-video-android/blob/develop/stream-video-android-core/src/main/kotlin/io/getstream/video/android/core/notifications/internal/VideoPushDelegate.kt#L82

@santhoshvai can you please help with the testing?

### DIFF
--- a/packages/client/src/StreamVideoClient.ts
+++ b/packages/client/src/StreamVideoClient.ts
@@ -303,15 +303,13 @@ export class StreamVideoClient {
    *
    * @param type the type of the call.
    * @param id the id of the call, if not provided a unique random value is used
-   * @param {boolean} [ringing] whether the call should be created in the ringing state.
    */
-  call = (type: string, id: string, ringing?: boolean) => {
+  call = (type: string, id: string) => {
     return new Call({
       streamClient: this.streamClient,
       id: id,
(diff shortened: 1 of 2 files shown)


santhoshvai (approved):

packages/client/src/StreamVideoClient.ts
```diff
@@ -484,6 +482,33 @@ export class StreamVideoClient {
     return this.streamClient.createToken(userID, exp, iat, call_cids);
   }
 
+  /**
+   * A callback that can be used to create ringing calls from push notifications. If the call already existis, it will do nothing.
```
santhoshvai, line 486:
```suggestion
   * A callback that can be used to create ringing calls from push notifications. If the call already exists, it will do nothing.
```

szuperaz: Thanks

PR #756, opened 2023-07-05
feat(react-native): ability to switch between spotlight and grid mode

khushal87:


vanGalilea (approved):

sample-apps/react-native/dogfood/src/components/ParticipantLayoutButton.tsx
```diff
@@ -0,0 +1,129 @@
+import React, { useState } from 'react';
+import { Pressable, Text, Modal, StyleSheet, View } from 'react-native';
+import GridIconSvg from '../assets/GridIconSvg';
+import { appTheme } from '../theme';
+import { CallParticipantsViewProps } from '@stream-io/video-react-native-sdk';
```
santhoshvai, line 5: no need to export props from SDK for this

The integrators can do `React.ComponentProps<CallParticipantsView>["mode"]`

this is the reason why prop types are not exported in react-native for example..

sample-apps/react-native/dogfood/src/components/ParticipantLayoutButton.tsx
```diff
@@ -0,0 +1,129 @@
...
+
+export type Mode = NonNullable<CallParticipantsViewProps['mode']>;
```
santhoshvai, line 7: no need to export the type Mode also..

again in any screen  ` Mode = React.ComponentProps<CallParticipantsView>["mode"] can be used`

packages/react-native-sdk/src/components/CallParticipantsView.tsx
```diff
@@ -4,7 +4,7 @@ import { CallParticipantsSpotlightView } from './CallParticipantsSpotlightView';
 import { useHasOngoingScreenShare } from '@stream-io/video-react-bindings';
 import { StyleSheet, View } from 'react-native';
 
-type CallParticipantsViewProps = {
+export type CallParticipantsViewProps = {
```
santhoshvai, line 7: lets avoid exporting props unless absolutely necessary.. explained below

santhoshvai (changes requested):
requesting to not export props

packages/react-native-sdk/src/components/CallParticipantsBadge.tsx
```diff
@@ -45,8 +45,6 @@ const styles = StyleSheet.create({
     paddingVertical: theme.padding.xs,
     paddingHorizontal: theme.padding.sm,
     position: 'absolute',
-    right: 0,
-    top: -theme.spacing.sm,
```
santhoshvai, line 49: nice removal.. negative margins are ugly unless

santhoshvai (approved):
Discussed props exporting.. and we preferrred that exporting is better as most people do not know React.ComponentProps yet

so lets keep as is!

khushal87:
> Discussed props exporting.. and we preferrred that exporting is better as most people do not know React.ComponentProps yet
> 
> so lets keep as is!

Yes, it's about knowing, that's one thing. Plus I don't see any harm if we export it. We are doing the same thing just in a different way.

PR #758, opened 2023-07-06
feat(react-dogfood): add useWakeLock hook

arnautov-anton:
This hook prevents your screen from dimming while in a long noninteractive call.

[Screen Wake Lock API](https://developer.mozilla.org/en-US/docs/Web/API/Screen_Wake_Lock_API) has not been yet implemented to the FF so this feature only works on other major browsers (apparently on Safari too 🤯).

### DIFF
--- a/sample-apps/react/react-dogfood/components/MeetingUI.tsx
+++ b/sample-apps/react/react-dogfood/components/MeetingUI.tsx
@@ -33,7 +33,7 @@ import {
   UnreadCountBadge,
 } from '.';
 import { ActiveCallHeader } from './ActiveCallHeader';
-import { useKeyboardShortcuts, useWatchChannel } from '../hooks';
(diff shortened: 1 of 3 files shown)


packages/react-dogfood/hooks/useWakeLock.ts
```diff
@@ -0,0 +1,32 @@
+import { CallingState, useCallCallingState } from '@stream-io/video-react-sdk';
+import { useEffect } from 'react';
+
+interface WakeLockSentinel {
+  release(): Promise<void>;
+}
+
+export const useWakeLock = () => {
```
oliverlaz, line 8: Nice! 🚀 

Shall we maybe include this hook in the SDK? And, perhaps enable it by default in the default layouts?

oliverlaz (approved):

MartinCupela (approved):

PR #767, opened 2023-07-07
fix: promise rejection on call leave

santhoshvai:
without this,

when pc.close is done, negotiate is called immediately on RN and maybe also non chromium browsers?! and results in the following error 

```
Possible Unhandled Promise Rejection (id: 1):
Error: CreateOffer called when PeerConnection is closed.
```


oliverlaz (approved):
Interesting... I haven't seen this issue on Web, but it shouldn't do any harm if we remove the event listener prior closing the peer connection.

PR #773, opened 2023-07-10
fix(react-native): video published on foreground even if it was muted

santhoshvai:


packages/react-native-sdk/src/hooks/useCallControls.tsx
```diff
@@ -30,36 +26,27 @@ export const useCallControls = () => {
...
 
   /** Attempt to republish video stream when app comes back to foreground */
-  useAppStateListener(
-    isVideoPublishedRef.current ? publishVideoStream : undefined,
-  );
+  useAppStateListener(() => {
+    if (isVideoPublished) {
+      publishVideoStream();
+    }
+  }, undefined);
```
vanGalilea, line 35:
```suggestion
  useAppStateListener(() => {
    if (isVideoPublished) {
      publishVideoStream();
    }
  });
```

vanGalilea (approved):

PR #787, opened 2023-07-12
fix: Attempt to recover Publisher PeerConnection

oliverlaz:
### Overview

Adds an automated ICE-restart mechanism on the Publisher Peer Connection that executes whenever the `iceConnectionState` becomes `failed` or `disconnected`.

This change aims to fix an issue in Firefox where we occasionally see failures in the Publisher peer connection.
Additionally, Pronto now has a "Dev" menu that can be used for forcefully triggering ICE restarts.


packages/client/src/rtc/Publisher.ts
```diff
@@ -635,11 +643,15 @@ export class Publisher {
...
 
   private onIceConnectionStateChange = () => {
-    this.logger(
-      'debug',
-      `ICE Connection state changed`,
-      this.pc.iceConnectionState,
-    );
+    const state = this.pc.iceConnectionState;
+    this.logger('debug', `ICE Connection state changed`, state);
+    if (state === 'failed' || state === 'disconnected') {
```
thesyncim, line 648: We can only recover from disconnected state

PR #791, opened 2023-07-13
refactor: RN Android testing pipeline

testableapple:
It was flaky... it should be more stable


.github/workflows/react-native-workflow.yml
```diff
@@ -186,18 +186,17 @@ jobs:
...
       - name: Test
         uses: reactivecircus/android-emulator-runner@v2
-        env:
-          ANDROID_API_LEVEL: 32
         with:
-          api-level: ${{ env.ANDROID_API_LEVEL }}
-          profile: pixel
+          api-level: 32
           arch: x86_64
+          target: google_apis # Change this to `default` ASAP to increase the build time
```
santhoshvai, line 192: have you tested with arm images? maybe those are faster 

Should be used with m1/m2 mac GH image though, I don't know if they exist

testableapple: They do exist, but they ain't free

PR #793, opened 2023-07-13
fix(react-demo): secrets and keys

zwaardje:
- And a tiny map update


szuperaz:
Screenshot 2023-07-13 at 12 29 37
The API key seems to be the old one

szuperaz (approved):

zwaardje:
> Screenshot 2023-07-13 at 12 29 37 The API key seems to be the old one

Thanks, good catch updated

PR #800, opened 2023-07-14
fix: Optimize release step

szuperaz:
No need to build all packages in the release step, only libs is enough


package.json
```diff
@@ -37,12 +37,14 @@
     "start:egress": "yarn workspace @stream-io/egress-composite start",
     "build:egress": "yarn workspace @stream-io/egress-composite build",
     "build:all": "yarn workspaces foreach -v --topological-dev run build",
+    "build:libs": "yarn workspaces foreach -v --topological-dev --no-private run build",
```
oliverlaz, line 40: a cool trick :)

oliverlaz (approved):

PR #804, opened 2023-07-17
chore: Refactor react sample app deployments

szuperaz:
- Move react-dogfood to `sample-apps`
- React dogfood is deployed together with other react sample apps
- Sample app deployment trigger changed:
1. PR affecting relevant files -> preview deployment
2. Pushes to `main` affecting files in `sample-apps/react` -> production deployment
3. New tags with format: `@stream-io/video-react-sdk-*` -> production deployment (this will ensure that sample apps will send the correct SDK version number)
- Some smaller CI trigger optimizations

If this system works, I'll update the react demo and RN sample apps as well in a separate PR.

The react dogfood deplyoment currently fails, I'll update the vercel config https://vercel.com/getstreamio/stream-calls-dogfood/settings before merge, I just didn't want to prematurely update this.


.github/workflows/test.yml
```diff
@@ -0,0 +1,61 @@
...
+        run: yarn install --immutable
+
+      - name: Lint Packages
+        run: yarn lint:ci:all
+
+      - name: Build packages
+        run: NODE_ENV=production yarn build:all
+
+      - name: Test packages
+        run: yarn test:ci:all
```
oliverlaz, line 55: Can we run the tests before we build everything? I don't think it makes sense to wait until all apps finish building and then stop because we have a failing test somewhere...

oliverlaz (approved):

.github/workflows/test.yml
szuperaz: I've changed the order

PR #808, opened 2023-07-17
docs(react-native): UI Component - CallParticipantsListView

khushal87:


packages/react-native-sdk/docusaurus/docs/reactnative/04-ui-components/call/call-participants-list-view.mdx
```diff
@@ -3,7 +3,63 @@ id: call-participants-list-view
...
+
+The list of participants to list in the view.
+
+### numberOfColumns
+
+| Type                    | Default Value |
+| ----------------------- | ------------- |
+| `number` \| `undefined` | 2             |
+
+The number of participants to be displayed in a single row. This works for more than 2 participants.
```
santhoshvai, line 57:
```suggestion
The number of participants to be displayed in a single row. This property is only used when there are more than 2 participants.
```

packages/react-native-sdk/src/components/call/CallParticipantsListView.tsx
```diff
@@ -36,7 +36,7 @@ interface CallParticipantsListProps {
    * The number of columns to display in the list of participants while in vertical or horizontal scrolling mode
```
santhoshvai, line 36:
```suggestion
   * The number of columns to display in the list of participants while in vertical or horizontal scrolling mode. This property is only used when there are more than 2 participants.
```

PR #819, opened 2023-07-19
docs(react-native): UI Component - LocalParticipantView

khushal87:
- Added the documentation of the component.
- Removed the unnecessary zIndex prop as the styles already have the Z_INDEX.IN_MIDDLE and if anyone wants to really customize it they can use the style prop.


packages/react-native-sdk/docusaurus/docs/reactnative/04-ui-components/participants/local-participant-view.mdx
```diff
@@ -3,8 +3,42 @@ id: local-participant-view
 title: LocalParticipantView
 ---
 
+LocalParticipantView displays the local video stream of the local participant if the video is unmuted. It can be used as both floating or full-screen view.
```
santhoshvai, line 6:
```suggestion
LocalParticipantView displays the local video stream of the local participant if the video is unmuted. It can be used as both a floating or full-screen view.
```

PR #827, opened 2023-07-20
fix(react-native): remove usage of crypto in df app

santhoshvai:


sample-apps/react-native/dogfood/src/modules/helpers/randomId.ts
khushal87, line 1: I think we should rename the file as `randomMeetingId` or keep it the same as `meetingId` to avoid confusion about its usage. What do you say?

santhoshvai: nah lets keep it as randomId.. it can be used for live stream, audioroom also in the future 

There is no confusion.. its a function that gives a random id

khushal87: In that case, it can be `randomCallId`. But anyways, it's on you. Approving the PR.

khushal87 (approved):

PR #828, opened 2023-07-20
feat(react-native): add translations to SDK and DF app

khushal87:
- Added translations for the DF app and SDK texts.
- Fixed the error of translations on app start.


santhoshvai:
What is the point of having de and es files? if its all a todo

khushal87:
> What is the point of having de and es files? if its all a todo

I didn't add them, it was already there. :)

packages/react-native-sdk/src/components/call/LobbyView.tsx
```diff
@@ -129,16 +133,22 @@ export const LobbyView = ({ joinCallButton }: LobbyViewType) => {
         )}
         <View style={styles.info}>
           <Text style={styles.infoText}>
-            You are about to join a call with id {call?.id} at Stream.{' '}
-            {participantsCount
-              ? `${participantsCount} participant(s) are in the call.`
-              : 'You are first to Join the call.'}
+            {t('You are about to join a call with id {{ callId }} at Stream.', {
+              callId: call?.id,
+            }) +
```
santhoshvai, line 138: problematic to say at Stream in SDK actually :( 

I actually don't know what to do here expect saying lets pass a prop for info

at stream basically means DF app

santhoshvai: I think the whole info text can be passed as prop 

Also we can add heading, subheading prop .. 

Expect for Join I don't see any other texts being useful in translations

sample-apps/react-native/dogfood/src/components/LobbyViewComponent.tsx
```diff
@@ -41,7 +42,7 @@ export const LobbyViewComponent = ({
           }}
         >
           <Text style={styles.anonymousButtonText}>
-            Join with your Stream Account
+            {t('Join with your Stream Account')}
```
santhoshvai, line 45: another problematic text

khushal87:
> we must remove them.. especially from SDK.. any TODOs inside the SDK is not good after release :(

We didn't add any text on the SDK. Most of the texts are DF-specific. I can't think of any issue here.

sample-apps/react-native/dogfood/src/components/LobbyViewComponent.tsx
khushal87: This is DF app so its fine.

packages/react-native-sdk/src/components/call/LobbyView.tsx
khushal87: Good find. Let me fix this.

khushal87: About the prop thing, what I believe is probably in the future is to design and keep LobbyView in the app itself and SDK won't export it in my opinion so it's fine for now.

sample-apps/react-native/dogfood/src/components/LobbyViewComponent.tsx
santhoshvai: ah ok got you

santhoshvai:
@khushal87 I got you sorry for the warn. 

But lets remove the DE and ES translations. Especially from SDK, exporting TODO texts is bad.

MartinCupela:
> @khushal87 I got you sorry for the warn.
> 
> But lets remove the DE and ES translations. Especially from SDK, exporting TODO texts is bad.

We should probably move them to the dogfood apps?

khushal87:
> > @khushal87 I got you sorry for the warn.
> > But lets remove the DE and ES translations. Especially from SDK, exporting TODO texts is bad.
> 
> We should probably move them to the dogfood apps?

Most of them are already in the DF app

PR #833, opened 2023-07-20
feat(client): runWithRetry - retryable 

arnautov-anton:


packages/client/src/helpers/runWithRetry.ts
```diff
@@ -0,0 +1,87 @@
...
+      isRetryable,
+      didValueChange,
+    }: {
+      retryAttempts?: number;
+      delayBetweenRetries?: number | ((attempt: number) => number);
+      isRetryable: (error: unknown) => boolean;
+      didValueChange?: (...functionArguments: Parameters<T>) => boolean;
+    },
+  ) =>
+  async (...functionArguments: Parameters<T>) => {
```
oliverlaz, line 47:
```suggestion
  async function retryable(...functionArguments: Parameters<T>) {
```

Let's convert this to a named function to see better stack traces when things go wild...

packages/client/src/helpers/runWithRetry.ts
```diff
@@ -0,0 +1,87 @@
...
+    {
+      retryAttempts = 3,
+      delayBetweenRetries,
+      isRetryable,
+      didValueChange,
+    }: {
+      retryAttempts?: number;
+      delayBetweenRetries?: number | ((attempt: number) => number);
+      isRetryable: (error: unknown) => boolean;
+      didValueChange?: (...functionArguments: Parameters<T>) => boolean;
```
oliverlaz, line 44: I think all of these inputs shall be marked as optional in their type definitions but they should have initial defaults:
- `delayBetweenRetries = retryInterval`
- `isRetryable = () => true`
- `didValueChange = () => false`

What do you think about this?

packages/client/src/helpers/runWithRetry.ts
```diff
@@ -0,0 +1,87 @@
+const sleep = (time: number) =>
+  new Promise<void>((resolve) => {
+    setTimeout(resolve, time);
+  });
```
oliverlaz, line 4: nit-pick, but I think there is a `sleep` function already available in `coordinator/connection/utils.ts`

packages/client/src/helpers/runWithRetry.ts
```diff
@@ -0,0 +1,87 @@
...
+    let retryAttempt = -1;
+
+    do {
+      let data: Awaited<ReturnType<T>> | null = null;
+      let error: unknown = null;
+      const isLastRun = retryAttempts === retryAttempt + 1;
+      const isInitialRun = retryAttempt === -1;
+
+      if (didValueChange?.(...functionArguments))
+        throw new RetryError({ type: 'abort' });
```
oliverlaz, line 58:
```suggestion
      if (didValueChange?.(...functionArguments)) {
        throw new RetryError({ type: 'abort' });
      }
```

packages/client/src/helpers/runWithRetry.ts
```diff
@@ -0,0 +1,87 @@
...
+
+      try {
+        data = await f(...functionArguments);
+      } catch (e) {
+        error = e;
+      }
+
+      if (didValueChange?.(...functionArguments))
+        throw new RetryError({ type: 'abort' });
```
oliverlaz, line 67:
```suggestion
      if (didValueChange?.(...functionArguments)) {
        throw new RetryError({ type: 'abort' });
      }
```

Perhaps we can fine-tune eslint to allow bracket-less `if` only for `if (cond) return <v>` case.

packages/client/src/helpers/runWithRetry.ts
arnautov-anton: I have applied most of your suggestions but `delayBetweenRetries` is the only one I'm not sure about - I feel like we should keep the default state of the retry mechanism as `FastAndSimple` - `retryInterval` function as default would make it something between `FastAndSimple` and `NeverGonnaGiveYouUp`.

packages/client/src/helpers/runWithRetry.ts
arnautov-anton: I agree - I like bracketless but it's more readable if it's only a one-liner. :+1:

packages/client/src/helpers/runWithRetry.ts
arnautov-anton: Also - I'm not going to provide default functions to not pollute the stack with unnecessary function executions, we'll execute only if they're available - otherwise we skip.

The only change is to `isRetryable` where we try to execute it but if it does not exist we fall back to `?? true`.

packages/client/package.json
```diff
@@ -52,14 +52,14 @@
...
+    "@vitest/coverage-v8": "^0.33.0",
     "dotenv": "^16.3.1",
     "prettier": "^2.8.4",
     "rimraf": "^3.0.2",
     "rollup": "^3.25.1",
     "typescript": "^4.9.5",
     "vite": "^4.3.9",
-    "vitest": "^0.30.1",
-    "vitest-mock-extended": "^1.1.3"
+    "vitest": "^0.33.0",
```
oliverlaz, line 62: I see it happening here as well... For some mysterious reason, whenever I attempted to upgrade `vitest`, sample apps couldn't be built anymore as `vite` stopped working 🤯

packages/client/src/helpers/runWithRetry.ts
oliverlaz: ok, sounds good.

packages/client/package.json
arnautov-anton: I needed it for `test.extend` but then got around without using it so I think we can revert these changes for now. :D

packages/client/src/Call.ts
```diff
@@ -237,7 +238,24 @@ export class Call {
...
+            didValueChange: (initialSubscriptions) => {
+              return (
+                initialSubscriptions !==
+                this.trackSubscriptionsSubject.getValue().data
+              );
+            },
+            isRetryable: (error) => {
+              // decide whether to retry execution
+              return true;
+            },
```
arnautov-anton, line 256: @oliverlaz We should probably adjust this part and generate some reusable `isRetryable` functions.

oliverlaz: We can have one that inspects SFU's `response.error.shouldRetry` as such instruction can be returned by the SFU even when the RPC succeeded (HTTP 200). Check `StreamSfuClient#retryable`.

packages/client/src/helpers/runWithRetry.ts
```diff
@@ -0,0 +1,181 @@
...
+
+      try {
+        data = await f(...functionArguments);
+      } catch (e) {
+        error = e;
+      }
+
+      if (await didValueChange?.(...functionArguments)) {
+        throw new RetryError({ type: 'abort' });
+      }
```
arnautov-anton, line 115: I've been thinking and realized that this abort check might be misleading and would cause stale-ish client-side state if the call goes through; local state would not represent server sate.

oliverlaz:
Retries were implemented differently in the meantime. Closing this stale PR.

PR #841, opened 2023-07-21
chore: exclude the angular-sdk from the workspace

oliverlaz:
### Overview

The Angular SDK has been in hibernation mode for a while but it hasn't been completely excluded from the workspace.
With this PR, we reduce the number of NPM packages we need to install in order to work with this repo.


szuperaz:
@oliverlaz pushed a commit the remove angular-sdk entirely

szuperaz (approved):

PR #842, opened 2023-07-21
feat(react-native): add permissions to docs

santhoshvai:


khushal87:
Seems like the CI is failing here because of the PR - https://github.com/GetStream/stream-video-js/actions/runs/5656007232/job/15322206823?pr=852

Merged without review 😢

PR #848, opened 2023-07-24
feat: support goLive({ notify: true })

sachaarbonel:
for context: https://github.com/GetStream/chat/pull/5211


szuperaz (approved):

packages/client/src/Call.ts
```diff
@@ -1435,10 +1435,11 @@ export class Call {
   /**
    * Starts the livestreaming of the call.
    */
-  goLive = async () => {
+  goLive = async (notify?: boolean) => {
```
oliverlaz, line 1438:
```suggestion
  goLive = async ({ notify = false }: { notify?: boolean }) => {
```

I think for consistency (see `call.get()`) and future-proofness, this method should accept an object with params.

sachaarbonel: make sense

oliverlaz (approved):

PR #850, opened 2023-07-25
chore: More integration tests

szuperaz:


packages/client/src/__tests__/server-side/call-members.test.ts
```diff
@@ -0,0 +1,89 @@
...
+    response = await call.queryMembers({
+      ...queryMembersReq,
+      next: response.next,
+    });
+
+    expect(response.members.length).toBe(1);
+
+    response = await call.queryMembers({
+      filter_conditions: { role: { $eq: 'admin' } },
+    });
```
oliverlaz, line 87: is this an extra request or we are missing an assertion afterwards? :)

oliverlaz (approved):

packages/client/src/__tests__/server-side/call-members.test.ts
szuperaz: Good catch, I forgot the assertion :D

PR #852, opened 2023-07-25
refactor(react-sdk): pass translations from SDK and not from i18n package

khushal87:
This PR is over #847


szuperaz (approved):

packages/react-sdk/src/core/components/StreamVideo/StreamVideo.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+export const StreamVideo = (props: PropsWithChildren<StreamVideoProps>) => {
+  const { children } = props;
+
+  return (
+    <StreamVideoProvider translationsOverrides={translations} {...props}>
+      {children}
+    </StreamVideoProvider>
+  );
+};
```
oliverlaz, line 16:
```suggestion
export const StreamVideo = (props: PropsWithChildren<StreamVideoProps>) => {
  return <StreamVideoProvider translationsOverrides={translations} {...props} />;
};
```

I don't think there is a need to unwrap `children` and then wrap them again. `{...props}` forwards `children` to the provider as well.

oliverlaz (approved):

packages/react-sdk/src/core/components/StreamVideo/StreamVideo.tsx
khushal87: Done @oliverlaz, thanks for the suggestion. Never thought this way. 😄

PR #881, opened 2023-07-31
feat!: Server-side participant pinning

oliverlaz:
### Overview

Adds support for Server-Side participant pinning.

Reference: https://github.com/GetStream/chat/pull/5239, https://github.com/GetStream/protocol/pull/174
Specs: https://www.notion.so/stream-wiki/Video-Pinning-2b348a962da742389f734b7fd9b1cf46

## Breaking changes
This PR brings a small breaking change in the participant pinning API. Previously, we used to have `call.setParticipantPinnedAt` API which is now replaced with `call.pin(sessionId)` and `call.unpin(sessionId)` APIs.

Additionally, the `participant.pinnedAt` property has been replaced with a new `pin` structure of this shape:
```
const participant = {
  ...,
  pin: {
    isLocal: boolean, // set to `true` when this participant is server-side pinned
    pinnedAt: number, // a timestamp, shows when "pinning" happened
  }
}
```


khushal87 (approved):
Approving changes for RN and pin and un-pin changes.

Pin for all needs to be tested.

Thanks, @oliverlaz for adding the changes to RN SDK. 🚀

packages/client/src/Call.ts
```diff
@@ -879,14 +883,15 @@ export class Call {
       const startedAt = callState?.startedAt
         ? Timestamp.toDate(callState.startedAt)
         : new Date();
+      const pins = callState?.pins || [];
```
khushal87, line 886:
```suggestion
      const pins = callState?.pins ?? [];
```

As suggested by sonarcloud

packages/react-sdk/src/components/CallParticipantsList/CallParticipantListingItem.tsx
```diff
@@ -238,90 +264,91 @@ export const ParticipantActionsContextMenu = ({
     <GenericMenu>
       <GenericMenuButtonItem onClick={toggleParticipantPinnedAt}>
         <Icon icon="pin" />
-        {participant.pinnedAt ? 'Unpin' : 'Pin'}
+        {pinnedAt ? t('Unpin') : t('Pin')}
       </GenericMenuButtonItem>
+      <Restricted requiredGrants={[OwnCapability.PIN_FOR_EVERYONE]}>
```
MartinCupela, line 269: Could this 2 buttons be wrapped in a single <Restricted> component?

oliverlaz: Sure. It is fixed in a recent commit.

packages/react-native-sdk/docusaurus/docs/reactnative/03-core/03-call-and-participant-state.mdx
```diff
@@ -123,21 +133,22 @@ const CallUI = () => {
...
+| `joinedAt`                | The time the participant joined the call.                                   |
+| `connectionQuality`       | The participant's connection quality.                                       |
+| `isSpeaking`              | It's `true` if the participant is currently speaking.                       |
+| `isDominantSpeaker`       | It's `true` if the participant is the current dominant speaker in the call. |
+| `audioLevel`              | The audio level of the participant.                                         |
+| `audioStream`             | The published audio `MediaStream`.                                          |
+| `videoStream`             | The published video `MediaStream`.                                          |
+| `screenShareStream`       | The published screen share `MediaStream`.                                   |
+| `isLocalParticipant`      | It's `true` if the participant is the local participant.                    |
+| `pinnedAt`                | The time the participant was pinned.                                        |
```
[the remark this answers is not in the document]
oliverlaz: This has been updated with the newest implementation. Now both, local and remote pins get a timestamp.

packages/client/src/store/CallState.ts
```diff
@@ -678,4 +678,33 @@ export class CallState {
...
+        lookup[pin.sessionId] = pinIndex + 1; // avoid setting to 0
+        return lookup;
+      },
+      {},
+    );
+
+    this.setParticipants((participants) =>
+      participants.map((participant) => {
+        // the local user might have overridden the pinning, skip updating
+        if (participant.isPinningSetByLocalUser) return participant;
```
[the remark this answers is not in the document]
oliverlaz: Updated according to the latest spec.

arnautov-anton (approved):

szuperaz (approved):

packages/react-native-sdk/src/components/participants/internal/ParticipantActions.tsx
```diff
@@ -176,11 +177,14 @@ export const ParticipantActions = (props: ParticipantActionsType) => {
         ]
       : [];
 
-  const pinParticipant: CallParticipantOptionType | null = {
-    icon: <Pin color={theme.dark.text_high_emphasis} />,
-    title: participant.pinnedAt ? 'Unpin' : 'Pin',
-    onPressHandler: toggleParticipantPinnedAt,
-  };
+  const pinParticipant: CallParticipantOptionType | null =
+    !participant.pin || participant.pin.isLocalPin
```
khushal87, line 181: Is there a way we can simplify this check? 🤔

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
```diff
@@ -273,7 +273,7 @@ export const ParticipantView = (props: ParticipantViewProps) => {
                 <VideoSlash color={theme.light.error} />
               </View>
             )}
-            {participant.pinnedAt && (
+            {participant.pin && participant.pin.isLocalPin && (
```
khushal87, line 276: Is there a way we can simplify this check? 🤔

oliverlaz: I'm avoiding using `?.` and `??` operators as I think they make the code less understandable as they increase its cognitive complexity.

Do you think it adds more value if I shorten this expression?

khushal87: That's true, but maybe we can extract this check into a variable and then use the variable here. Makes it easier to understand.

Something like:
const isPinnedEnabled = participant.pin && participant.pin.isLocalPin;

packages/react-native-sdk/src/components/participants/internal/ParticipantActions.tsx
oliverlaz: fixed.

khushal87 (approved):

PR #903, opened 2023-08-07
fix(react-native): disable camera toggle face button when local video is muted

khushal87:
The toggle camera face button should be disabled when the local video is muted. This problem is solved in this PR.


packages/react-native-sdk/src/components/utility/internal/CallControlsButton.tsx
```diff
@@ -18,6 +18,10 @@ interface CallControlsButtonProps {
    * The background color of the button rendered.
    */
   color?: string;
+  /**
+   * Boolean to enable/enable the button
```
santhoshvai, line 22:
```suggestion
   * Boolean to enable or disable the button
```

PR #905, opened 2023-08-07
refactor(react-native): remove View suffix from components

khushal87:
The Docs, Code refactor and tests are refactored as a part of the PR. This PR just focuses on `View` suffix removal. The customization and React Alignment will come up soon.
 
Note: The LocalParticipantView is not done in this PR as per request from @santhoshvai


packages/react-native-sdk/src/components/call/Lobby.tsx
```diff
@@ -21,27 +21,27 @@ import { useMediaStreamManagement } from '../../providers/MediaStreamManagement'
...
    */
   onPressHandler: () => void;
 };
 
 /**
- * Props for the Lobby View Component
+ * Props for the Lobby Component.
  */
-type LobbyViewType = {
+type LobbyType = {
```
vishalnarkhede, line 37: `Type` suffix feels redundant. `LobbyProps` will be a better choice

vishalnarkhede: Same goes for other component props typescript as well

khushal87: Let me fix it 😄

vishalnarkhede (approved):

PR #908, opened 2023-08-07
feat: New device API v1

szuperaz:
This is the first iteration of our new device management API.

The new API is only used inside the JS client for now, React and RN SDKs can adapt it at their own pace. This is not a breaking change, the old device management APIs will still work for React and RN SDK.

Missing features (coming in separate PRs):
- Speaker management
- Better device disconnection handling (not relevant to RN)
- Built-in support for speaking while muted notification

+1: added a client sample app


packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -0,0 +1,92 @@
...
+   * Stops camera/microphone
+   * @returns
+   */
+  async disable() {
+    if (!this.state.mediaStream) {
+      return;
+    }
+    if (this.call.state.callingState === CallingState.JOINED) {
+      await (this.kind === 'videoinput'
+        ? this.call.stopPublish(TrackType.VIDEO)
```
MartinCupela, line 61: Shouldn't we dispose of the stream here as well?

packages/client/src/devices/InputMediaDeviceManagerState.ts
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * An Observable that emits the currently selected device
+   */
+  selectedDevice$: Observable<string | undefined>;
+
+  /**
+   * An Observable that emits the device status
+   */
+  // TODO: we should extend this list with loading and error status options
+  status$: Observable<'enabled' | 'disabled'>;
```
MartinCupela, line 27: Could status be infered from `mediaStream$`? Meaning, that we would not need `status$` state variable?

szuperaz: We have the `status$` for parity with android, but status is computed from `mediaStream` https://github.com/GetStream/stream-video-js/blob/new-device-api/packages/client/src/devices/InputMediaDeviceManagerState.ts#L33

MartinCupela: Ok so this is about parity

PR #914, opened 2023-08-09
feat(Call): Dynascale support for Plain-JS SDK

arnautov-anton:
### Overview

Introduces `call.bindVideoElement` and `call.bindAudioElement` methods which apply dynascale handlers to the provided video and audio elements, simplifies SDK-specific implementations.  

### Breaking Changes
- `call.viewportTracker` has been replaced with `call.dynascaleManager`
- `call.updateSubscriptionsPartial`
  - `video` and `screen` kinds are deprecated and will be removed soon. Please switch to `videoTrack` or `screenShareTrack`
- `<Audio />` component:
  - `audioStream` prop is dropped and now it requires a `participant` prop
  - `sinkId` doesn't have to be provided anymore, as our SDK handles this
- `<Video />`
  - `kind` prop has been replaced with `trackType`. Accepted values: `videoTrack`, `screenShareTrack` and `none` 
- `<ParticipantView />`
  - `videoMode` has been renamed to `trackType`. Accepted values:  `videoTrack`, `screenShareTrack` and `none`


packages/client/src/Call.ts
```diff
@@ -1688,4 +1701,93 @@ export class Call {
       { custom: payload },
     );
   };
+
+  registerVideoElement = (
```
arnautov-anton, line 1705: I'm thinking to have such methods live in `CallWebAPI` class and accessing those would look like:

```ts
call.web.registerVideoElement(...);
```

But I guess that's for the future updates when we have more than one method related to web-only APIs.

szuperaz (commented):
@arnautov-anton do I understand it correctly that we won't be able to detect if a video element is not visible due to scrolling? Can't we do that if integrators provide a parent container that holds all video elements?

arnautov-anton:
> @arnautov-anton do I understand it correctly that we won't be able to detect if a video element is not visible due to scrolling? Can't we do that if integrators provide a parent container that holds all video elements?

The thing is that we don't apply visibility handlers to the video element but rather to the wrapper that holds both video element and video placeholder. I can't remember for sure why but we had a good reasoning - once I find out I'll let you know. Although we should probably provide simplified API around ViewportTracker - meaning it should at least with [update state function](https://github.com/GetStream/stream-video-js/blob/b820b946131e5cd4f897daf5f2f18fadcd980697/packages/react-sdk/src/core/hooks/useTrackElementVisibility.ts#L22-L30) pre-defined.

packages/client/src/Call.ts
```diff
@@ -1929,10 +1968,13 @@ export class Call {
...
         const source = kind === 'video' ? p.videoStream : p.screenShareStream;
-        if (videoElement.srcObject === source) return;
+        if (videoElement.srcObject === source || !source) return;
 
         setTimeout(() => {
-          videoElement.srcObject = source ?? null;
+          videoElement.srcObject = source;
+          videoElement.play().catch((e) => {
+            this.logger('warn', `Failed to play stream`, e);
+          });
```
arnautov-anton, line 1977: Wouldn't be this covered by the `autoplay` attribute or does resetting the source cause the autoplay to not work?

packages/client/src/Call.ts
```diff
@@ -1929,10 +1968,13 @@ export class Call {
       )
       .subscribe((p) => {
         const source = kind === 'video' ? p.videoStream : p.screenShareStream;
-        if (videoElement.srcObject === source) return;
+        if (videoElement.srcObject === source || !source) return;
```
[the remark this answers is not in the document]
oliverlaz: Good catch. I'll adjust.

packages/client/src/Call.ts
oliverlaz: Technically it should, but at some point, there were some issues with it in Firefox and Safari, and explicit invocation of `play()` helped to resolve them. I'd like to keep the same code

packages/client/src/Call.ts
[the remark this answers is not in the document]
oliverlaz: Adjusted.

packages/client/docusaurus/docs/javascript/02-guides/03-call-and-participant-state.mdx
```diff
@@ -82,21 +82,23 @@ If you want to display information about the joined participants of the call you
...
+| `isSpeaking`                | It's `true` if the participant is currently speaking.                       |
+| `isDominantSpeaker`         | It's `true` if the participant is the current dominant speaker in the call. |
+| `audioLevel`                | The audio level of the participant.                                         |
+| `audioStream`               | The published audio `MediaStream`.                                          |
+| `videoStream`               | The published video `MediaStream`.                                          |
+| `screenShareStream`         | The published screen share `MediaStream`.                                   |
+| `isLocalParticipant`        | It's `true` if the participant is the local participant.                    |
+| `pin`                       | Holds pinning information.                                                  |
+| `reaction`                  | The last reaction this user has sent to this call.                          |
+| `viewportVisibilityState    | The viewport visibility state of the participant.                           |
```
szuperaz, line 100:
```suggestion
| `viewportVisibilityState`    | The viewport visibility state of the participant.                           |
```

packages/client/docusaurus/docs/javascript/02-guides/03-call-and-participant-state.mdx
```diff
@@ -82,21 +82,23 @@ If you want to display information about the joined participants of the call you
...
+| `screenShareVisibilityState | The screen share visibility state of the participant.                       |
```
szuperaz, line 101:
```suggestion
| `screenShareVisibilityState` | The screen share visibility state of the participant.                       |
```

packages/client/docusaurus/docs/javascript/02-guides/12-visibility-tracking.mdx
```diff
@@ -0,0 +1,76 @@
...
+title: Participant visibility tracking
+description: Learn how to set up participant visibility tracking the screen.
+---
+
+Depending on the UI layout of your call, some participants may not be visible on the screen at all times.
+This is especially true for large calls with many participants. We can optimize the performance and bandwidth usage of our SDK
+by only subscribing to video for participants that are visible on the screen or within a certain viewport.
+
+To help you keep track of which participants are visible on the screen, our SDK provides a few helpers.
+
```
szuperaz, line 11:
```suggestion
Once you set up visibility tracking to a participant, the client can automatically subscribe and unsubscribe to the video and screen stream of the given participant. 
```

package.json
```diff
@@ -75,7 +75,8 @@
     "nx": "16.0.1",
     "prettier": "^2.8.4",
     "typescript": "^4.9.5",
-    "vercel": "^30.2.1"
+    "vercel": "^32.1.0",
+    "vite": "^4.4.9"
```
khushal87, line 79: What is the reason for having `vite` on root?

oliverlaz: it is a weird combination... for some reason, upon upgrading `vitest` to a newer version, yarn can't locate the `vite` binary anymore. If I recall correctly, you once experienced this issue.
Putting `vite` as a top-level dependency eliminates this issue.

khushal87: Yeah, I had experienced it back then. I thought this was added to the root mistakenly. If it makes sense, its fine.

khushal87 (approved):

PR #915, opened 2023-08-09
feat(react-native): ParticipantView customizations

khushal87:
The goal of the PR is to introduce customization of the ParticipantView component. This introduces smaller components that can be overridden by passing them as a prop to ParticipantView(this can be raised to higher levels upon planning. Would be a simple task).

The following are customizable:
- Audio component
- Label - used to customize the participant label with name, mute status, etc.
- Reaction - used to customize the participant reaction container.
- Video which renders the stream or displays the placeholder when the video is off/muted.
- Network quality indicator
- Video Placeholder

Note: The documentation will be updated in subsequent PRs. Also, I haven't introduced any additional style props that might make the implementation hard. We can decide on what we want to cover with a `style` prop later and strive to do so.


packages/react-native-sdk/src/components/participants/ParticipantView.tsx
```diff
@@ -0,0 +1,276 @@
...
+
+  return (
+    <View
+      style={[styles.containerBase, props.containerStyle, speakerStyle]}
+      testID={`participant-${userId}`}
+      accessibilityValue={{
+        text: isSpeaking
+          ? 'participant-is-speaking'
+          : 'participant-is-not-speaking',
+      }}
```
santhoshvai, line 233: can we remove this? is it used for testing?

packages/react-native-sdk/src/components/participants/ParticipantReaction.tsx
```diff
@@ -1,19 +1,28 @@
...
-  reaction?: StreamReaction;
-  sessionId: string;
+export type ParticipantReactionProps = {
+  /**
+   * The participant whose info will be displayed.
+   */
+  participant: StreamVideoParticipant | StreamVideoLocalParticipant;
+  /**
+   * The duration after which the reaction should disappear.
+   */
```
santhoshvai, line 18: can we add the default value to the comment here pls

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
```diff
@@ -0,0 +1,276 @@
...
+      onLayout={onLayout}
+    >
+      <View style={styles.topView}>
+        <ParticipantReaction participant={participant} />
+      </View>
```
santhoshvai, line 238: To allow the styles to be customisable, lets move the container inside participant reaction..

so that customer can move it top or bottom or center or anything

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
```diff
@@ -0,0 +1,276 @@
...
+  ParticipantVideoProps,
+} from './ParticipantVideo';
+import { ParticipantVideoPlaceholderProps } from './ParticipantVideoPlaceholder';
+
+export type ParticipantVideoType = 'video' | 'screen';
+
+/**
+ * Props to be passed for the Participant component.
+ */
+export type ParticipantViewProps = {
```
santhoshvai, line 36: we need a style prop here too for overriding the container style

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
khushal87: Yes its used for testing.

packages/react-native-sdk/src/components/participants/ParticipantReaction.tsx
khushal87: Sure

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
khushal87: Possible for Reaction, but I doubt it would be easy to do it for the other component like label and connection quality that is in the bottom in the default view they should be present in the bottom. This can be touched when we do theming.

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
khushal87: Styling is for later, as I mentioned in my PR description and this goes hand in hand with theming.

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
santhoshvai: @khushal87 lets remove it! move this to testid

khushal87: We also have a test id here 😅

packages/react-native-sdk/src/components/participants/VideoRenderer.tsx
```diff
@@ -0,0 +1,209 @@
...
+   *
+   * @default false
+   */
+  muteVideo?: boolean;
+};
+
+/**
+ * Lower level component, that represents only the video part (wrapper around the WebRTC)
+ */
+export const VideoRenderer = ({
```
santhoshvai, line 43: We need to move videoFallback inside as a prop to video renderer

as per [Android](https://staging.getstream.io/video/docs/android/ui-cookbook/video-fallback/)

santhoshvai: videoFallback prop 

lives in `ParticipantView` then passed to `VideoRenderer`

santhoshvai: also in this way onLayout within this component

packages/react-native-sdk/src/components/call/CallParticipantsList.tsx
```diff
@@ -137,15 +137,15 @@ export const CallParticipantsList = (props: CallParticipantsListProps) => {
...
       );
       return (
-        <Participant
+        <ParticipantView
           participant={participant}
           containerStyle={itemContainerStyle}
-          kind="video"
-          isVisible={isVisible}
+          videoMode="video"
+          muteVideo={!isVideoVisible}
```
santhoshvai, line 148: I see that `isVisible` has been changed to `muteVideo`

but why? what was the reason.. aligns to which platform here

packages/react-native-sdk/src/components/call/CallParticipantsList.tsx
```diff
@@ -159,15 +159,16 @@ export const CallParticipantsList = (props: CallParticipantsListProps) => {
...
-            kind="video"
-            isVisible={true}
-            key={keyExtractor(participant, index)}
-          />
-        ))}
+        {participants.map((participant, index) => {
+          return (
+            <ParticipantView
+              participant={participant}
+              containerStyle={styles.flexed}
```
santhoshvai, line 166:
```suggestion
              style={styles.flexed}
```

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
santhoshvai:
```
testID={isSpeaking ? `participant-${userId}-is-speaking` : `'participant-${userId}-is-not-speaking'`}
```

packages/react-native-sdk/src/components/participants/VideoRenderer.tsx
```diff
@@ -0,0 +1,209 @@
...
+        dimension,
+      },
+    });
+    subscribedVideoLayoutRef.current = dimension;
+    pendingVideoLayoutRef.current = undefined;
+  };
+
+  if (!canShowVideo) {
+    return (
+      <ParticipantVideoFallback participant={participant} onLayout={onLayout} />
```
santhoshvai, line 188: instead of passing an onLayout Prop to both ParticipantVideoFallback and RTCView why not we have a common container and use onLayout there

```
<View onLayout={onLayout}>
  {canShowVideo?  <RTCView /> :  <ParticipantVideoFallback />}
</View>
```

packages/react-native-sdk/src/components/participants/VideoRenderer.tsx
```diff
@@ -0,0 +1,209 @@
...
+          viewportVisibilityState: VisibilityState.VISIBLE,
+        }));
+      }
+    } else {
+      if (viewportVisibilityState !== VisibilityState.INVISIBLE) {
+        call.state.updateParticipant(sessionId, (p) => ({
+          ...p,
+          viewportVisibilityState: VisibilityState.INVISIBLE,
+        }));
+      }
```
santhoshvai, line 95: this will create a bug.. we need to inform invisible state at all cases..

that's why the prop was called `isVisible` if it was reported as `isVisible=false` we updated participant to invisible here

with this change it wont happen anymore!

santhoshvai: what triggered the change to call it muteVideo?

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
```diff
@@ -0,0 +1,133 @@
...
+   */
+  participant: StreamVideoParticipant;
+  /**
+   * The video kind that will be displayed
+   */
+  videoMode: ParticipantVideoType;
+  /**
+   * Any custom style to be merged with the participant view
+   */
+  containerStyle?: StyleProp<ViewStyle>;
```
santhoshvai, line 41:
```suggestion
  style?: StyleProp<ViewStyle>;
```

simple change right?!

santhoshvai: and consistent with other SDKs

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
santhoshvai: or 
```
testID={`participant-${userId}-is-${isSpeaking ? "" : "not-"}speaking`
```

packages/react-native-sdk/src/components/participants/VideoRenderer.tsx
```diff
@@ -0,0 +1,209 @@
...
+    // NOTE: We only want to update the subscription if the pendingVideoLayoutRef is set
+    const updateIsNeeded = pendingVideoLayoutRef.current;
+
+    if (!updateIsNeeded || !call || !isPublishingVideoTrack || !hasJoinedCall) {
+      return;
+    }
+
+    // NOTE: When the view is not visible, we want to subscribe to audio only.
+    // We unsubscribe their video by setting the dimension to undefined
+    const dimension = !muteVideo ? pendingVideoLayoutRef.current : undefined;
```
santhoshvai, line 128: again visible and muteVideo is confusing :(

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
khushal87: Yeah, that's the only solution

packages/react-native-sdk/src/components/call/CallParticipantsList.tsx
khushal87: Basically, android has no such prop, I completely agree. 
React has a prop named, `muteAudio` that only plays video, and audio is muted.
On a similar note, we have `muteVideo` that can be used to only play audio even when the video is off.

Since the convention that react follows is with the mute suffix I changed `isVisible` to `muteVideo`. Makes it more readable.

khushal87: Also, there is some visibility state logic in the useEffect of VideoRenderer(before ParticipantView) that revolves around this variable(before `isVisible` and now `muteVideo`). Initially, I thought of removing this variable once and for all but then I was not sure about the logic.

packages/react-native-sdk/src/components/participants/VideoRenderer.tsx
khushal87: isVisible was not relevant to any of the platforms hence it was renamed to `muteVideo` as React has `muteAudio`.

khushal87: Also, isVisible is kinda confusing as: to which stream's visibility you are talking about, audio/video.

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
khushal87: yup

packages/react-native-sdk/src/components/participants/VideoRenderer.tsx
khushal87: Done ✅

packages/react-native-sdk/src/components/call/CallParticipantsList.tsx
khushal87: Done ✅

khushal87:
isVisible changes done ✅

packages/react-native-sdk/src/components/participants/Participant.tsx
santhoshvai, line 1: good riddance 👏

packages/react-native-sdk/__tests__/components/Participant.test.tsx
santhoshvai, line 1: this file must be renamed to `ParticipantView.test.tsx`

packages/react-native-sdk/src/components/participants/ParticipantView.tsx
```diff
@@ -0,0 +1,135 @@
...
+  ParticipantVideoFallback?: ComponentType<ParticipantVideoFallbackProps>;
+  /**
+   * Component to customize the network quality indicator of the participant.
+   */
+  ParticipantNetworkQualityIndicator?: ComponentType<ParticipantNetworkQualityIndicatorProps>;
+};
+
+/**
+ * A component that renders the participants' video track or screenShare track
+ * and additional info. By an absence of a video track or when muteVideo is truthy,
```
santhoshvai, line 72:
```suggestion
 * and additional info. By an absence of a video track or when isVisible is false,
```

packages/react-native-sdk/src/components/call/internal/CallParticipantsSpotlight.tsx
```diff
@@ -60,13 +60,10 @@ export const CallParticipantsSpotlight = () => {
...
-    flex: 2,
-    overflow: 'hidden',
-    borderRadius: theme.rounded.sm,
+    flex: 1,
     marginHorizontal: theme.padding.sm,
   },
   participantVideoContainer: {
-    flex: 1,
     paddingVertical: theme.padding.sm,
   },
```
santhoshvai, line 68: this style seems like a big change, does it look the same or figma aligned now?

khushal87: Yes(from last test) it should be the same, may be if you can confirm after testing if something is broken. These were kinda unnecessary styles.

khushal87: Yeah, there were some issues that went unnoticed due to this change and I thought them to be working. They are solved now. ✅

packages/react-native-sdk/__tests__/components/Participant.test.tsx
khushal87: Done ✅

packages/react-native-sdk/src/components/participants/ParticipantLabel.tsx
```diff
@@ -0,0 +1,97 @@
...
+      <View style={styles.status}>
+        <Text style={styles.userNameLabel} numberOfLines={1}>
+          {participantLabel}
+        </Text>
+        {isAudioMuted && (
+          <View style={[styles.svgContainerStyle, theme.icon.xs]}>
+            <MicOff color={theme.light.error} />
+          </View>
+        )}
+        {call?.type !== 'audio_room' && isVideoMuted && (
```
santhoshvai, line 42:
```suggestion
        {isVideoMuted && (
```

santhoshvai (approved):

santhoshvai (approved):

PR #918, opened 2023-08-09
fix(ParticipantView): remove audio element while muted

arnautov-anton:
The `audio` element essentially does nothing when muted so to reduce amount of rendered elements we can safely remove it.

### DIFF
--- a/packages/react-sdk/src/core/components/ParticipantView/ParticipantView.tsx
+++ b/packages/react-sdk/src/core/components/ParticipantView/ParticipantView.tsx
@@ -160,12 +160,13 @@ export const ParticipantView = forwardRef<HTMLDivElement, ParticipantViewProps>(
         )}
       >
         <ParticipantViewContext.Provider value={participantViewContextValue}>
-          <Audio
-            // mute the local participant, as we don't want to hear ourselves
-            muted={isLocalParticipant || muteAudio}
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/core/components/ParticipantView/ParticipantView.tsx
```diff
@@ -160,12 +160,13 @@ export const ParticipantView = forwardRef<HTMLDivElement, ParticipantViewProps>(
...
       >
         <ParticipantViewContext.Provider value={participantViewContextValue}>
-          <Audio
-            // mute the local participant, as we don't want to hear ourselves
-            muted={isLocalParticipant || muteAudio}
-            sinkId={localParticipant?.audioOutputDeviceId}
-            audioStream={audioStream}
-          />
+          {/* mute the local participant, as we don't want to hear ourselves */}
+          {(isLocalParticipant || muteAudio) && (
```
szuperaz, line 164: Shouldn't this be `!isLocalParticipant && !muteAudio`?

arnautov-anton: 🤦‍♂️ You are right! I shouldn't be working... fixing right away. 👍

oliverlaz (approved):

szuperaz (approved):

PR #926, opened 2023-08-11
feat(react-native): add livestream docs

santhoshvai:


packages/react-native-sdk/docusaurus/docs/reactnative/02-tutorials/03-livestream.mdx
```diff
@@ -0,0 +1,448 @@
...
+To fix that, we need to provide valid credentials. They shall be set in the following variables:
+
+```ts
+const apiKey = 'REPLACE_WITH_API_KEY';
+const token = 'REPLACE_WITH_TOKEN';
+const userId = 'REPLACE_WITH_USER_ID';
+const callId = 'REPLACE_WITH_CALL_ID';
+```
+
+Replace them now with the appropriate values from the table below:
```
vishalnarkhede, line 181: Could we set these values before `Broadcast a livestream from your device` in .env file or even just directly in App.tsx or separate file?

So that when user adds the actual code he won't receive error.

PR #929, opened 2023-08-13
refactor(react-bindings): remove redundant language check

khushal87:
If you check the logic for `changeLanguage`, the first condition in it says https://github.com/GetStream/stream-video-js/blob/main/packages/i18n/src/StreamI18n.ts#L88 which internally throws the warning:
```
I18n instance is not initialized. Call yourStreamI18nInstance.init().
```

This is a yellow box warning for RN SDK.
Simulator Screenshot - iPhone SE (3rd generation) - 2023-08-13 at 22 28 47

-----

Now in StreamI18nContext, we have a redundant condition that seems to `changeLanguage` if there is no `language` passed. But that condition is used when the `i18n.isInitialized` is false/undefined, which will, in turn, call the `this._checkIsInitialized()` from the `changeLanguage` and return the warning.

This PR tries to solve this issue and remove the redundant condition.

### DIFF
--- a/packages/react-bindings/src/contexts/StreamI18nContext.tsx
+++ b/packages/react-bindings/src/contexts/StreamI18nContext.tsx
@@ -56,11 +56,13 @@ export const useCreateI18n = ({
   );
 
   useEffect(() => {
-    if (i18n.isInitialized && language && i18n?.currentLanguage !== language) {
-      i18n.changeLanguage(language);
-    } else if (!i18n.isInitialized) {
-      if (!language) i18n.changeLanguage();
+    const { isInitialized } = i18n;
+    if (!isInitialized) {
       i18n.init().then((_i18n) => setTranslationFn(() => _i18n.i18nInstance.t));
+    } else {
+      if (language && i18n?.currentLanguage !== language) {
+        i18n.changeLanguage(language);
+      }
     }
   }, [i18n, i18nInstance, language, translationsOverrides]);
 
--- a/packages/react-bindings/src/contexts/StreamI18nContext.tsx
+++ b/packages/react-bindings/src/contexts/StreamI18nContext.tsx
@@ -59,10 +59,10 @@ export const useCreateI18n = ({
     const { isInitialized } = i18n;
     if (!isInitialized) {
       i18n.init().then((_i18n) => setTranslationFn(() => _i18n.i18nInstance.t));
-    } else {
-      if (language && i18n?.currentLanguage !== language) {
-        i18n.changeLanguage(language);
-      }
+      return;
+    }
+    if (language && i18n?.currentLanguage !== language) {
+      i18n.changeLanguage(language);
     }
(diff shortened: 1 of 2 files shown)


packages/react-bindings/src/contexts/StreamI18nContext.tsx
```diff
@@ -56,11 +56,13 @@ export const useCreateI18n = ({
...
 
   useEffect(() => {
-    if (i18n.isInitialized && language && i18n?.currentLanguage !== language) {
-      i18n.changeLanguage(language);
-    } else if (!i18n.isInitialized) {
-      if (!language) i18n.changeLanguage();
+    const { isInitialized } = i18n;
+    if (!isInitialized) {
       i18n.init().then((_i18n) => setTranslationFn(() => _i18n.i18nInstance.t));
+      return;
```
santhoshvai, line 62:
```
      return;
```

I understand that this makes change language not to be executed when not initialised

but what would happen in this case

```
(!isInitialized && language)
```

dony we need to set the language here ? 

also when `translationsOverrides` changes, nothing changes here.. I think something is buggy with this change

khushal87: Even if you change language, https://github.com/GetStream/stream-video-js/blob/main/packages/i18n/src/StreamI18n.ts#L88 won't allow that.

santhoshvai: no what I mean is 

```
  const [i18n] = useState<StreamI18n>(
    () =>
      i18nInstance ||
      new StreamI18n({ currentLanguage: language, translationsOverrides }),
  );
```

this covers tthis `(!isInitialized && language)` part

what would happen if `translationsOverrides` changes but language is the same ? What sets it ?! that is the bug according to me

santhoshvai: also if `i18nInstance` changes nothing would happen 

looks like these cases are not handled 

cc @MartinCupela

santhoshvai: these 2 bugs seem to already exist before this change actually

khushal87:
```
when translationsOverrides changes, nothing changes here
```

This doesn't seem to happen in `main` also(unless you hard reload the app) so the PR is not related.

khushal87: Exactly, this PR doesn't focus on that part. It just removes the wrong condition that was present.

About setting the language, I didn't find a method on the i18n package to do so currently. This came up to my mind but I am not sure how should we be doing it with the current limitations of the i18n package. If we do it, we need to introduce a different way of doing it other than changeLanguage.

MartinCupela (approved):

PR #933, opened 2023-08-15
feat(react-native): add CallContent component with customization of inner components

khushal87:
This PR primarily focuses on the introduction of the CallContent component which is also the root component of the customization of inner components.

This PR also removes the `fullscreen` mode from `LocalParticipantView`.


packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+  useEffect(() => {
+    return () => {
+      if (activeCallRef.current?.state.callingState !== CallingState.LEFT) {
+        activeCallRef.current?.leave();
+      }
+    };
+  }, []);
+
+  const participantViewProps = {
```
santhoshvai, line 92:
```suggestion
  const participantViewProps: ParticipantProps = {
```

add some typing here so that future maintenance is easy

packages/react-native-sdk/src/components/Call/CallLayout/CallParticipantsGrid.tsx
```diff
@@ -57,25 +50,35 @@ export const CallParticipantsGrid = ({
...
 
-  if (showFloatingView && isUserAloneInCall) {
-    return <LocalParticipantView layout={'fullscreen'} />;
+  if (isUserAloneInCall) {
+    return (
+      <>
+        {LocalParticipantView && <LocalParticipantView layout={'fullscreen'} />}
+      </>
+    );
   }
```
santhoshvai, line 59: inteersting what this will do is..

if integrator doesn't want a floating view so passes null here..

but when he is alone he will not see anything

santhoshvai: Does other SDKs have a full screen local video component?

santhoshvai: the whole local participnat view component seems to have no customisation support right?

packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
khushal87: In that case, we will have to create a separate type for all these component props. Are you fine with it?

santhoshvai: for the shared ones, yes!

sample-apps/react-native/dogfood/src/components/CallControlsComponent.tsx
```diff
@@ -0,0 +1,57 @@
...
+  ToggleVideoPublishingButton,
+} from '@stream-io/video-react-native-sdk';
+import React from 'react';
+import { StyleSheet, View } from 'react-native';
+import { appTheme } from '../theme';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
+import { Z_INDEX } from '../constants';
+
+export type CallControlsComponentProps = {
+  chatButton?: ChatButtonProps;
```
vishalnarkhede, line 17: this prop needs to go or be replaced

khushal87: This was a DF app thing. But anyways I have break the props in two parts and used it now. ✅

packages/react-native-sdk/src/components/Call/CallLayout/CallParticipantsGrid.tsx
khushal87: I have removed the fullscreen mode all in all from LocalParticipantView

khushal87:
Merging this PR after self-review. ✅

PR #936, opened 2023-08-15
feat(client): prefer vp8 for android on react-native

santhoshvai:
This PR adds support for:

* Always prioritising VP8 on Android -- aligns with flutter
* Adds support to remove video codecs as well with check for `rtcp-fb` -- currently we don't remove video codecs though


santhoshvai:
We might not need this change, I am trying out livelit's fork of webrtc now to get setCodecPreferences. 

If that works we can close this PR and delete the munging also

PR #939, opened 2023-08-16
chore: optimize the Github Actions

oliverlaz:
### Overview

- prevent triggering runs from unrelated directories
- disable Sonar
- remove unnecessary matrix strategy

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,11 +12,11 @@ on:
     branches:
       - main
     paths-ignore:
-      - '!**/docusaurus/**'
+      - '**/docusaurus/**'
   pull_request:
     types: [opened, synchronize, reopened]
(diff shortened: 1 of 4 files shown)


.github/workflows/test.yml
```diff
@@ -12,30 +12,25 @@ on:
     branches:
       - main
     paths-ignore:
-      - 'packages/react-native-sdk/docusaurus/**'
-      - 'packages/react-sdk/docusaurus/**'
+      - '!**/docusaurus/**'
```
szuperaz, line 15:
```suggestion
      - '**/docusaurus/**'
```

oliverlaz: This is what happens when copy-pasting 😅 Thanks for catching this!

PR #941, opened 2023-08-16
feat(client): inform sfu about pause and resume of tracks

santhoshvai:


packages/client/src/devices/CameraManager.ts
```diff
@@ -56,18 +56,14 @@ export class CameraManager extends InputMediaDeviceManager<CameraManagerState> {
   /**
    * Disables the video tracks of the camera
    */
-  pause() {
-    this.state.mediaStream?.getVideoTracks().forEach((track) => {
-      track.enabled = false;
-    });
+  pause(): Promise<void> {
+    return this.call.pauseTrack(TrackType.VIDEO);
```
szuperaz, line 60: This won't work on the lobby screen

szuperaz:
We'll have API differences with Android, their `pause` and `resume` methods behave differently. How about this: 
- `disable` and `pause` methods have a param where we can specify if we want to stop or disable the track
- `resume` should work as in Android: it resumes (starts/enables) the media track if it was previously enabled

packages/client/src/devices/CameraManager.ts
santhoshvai: yes I see.. so the way is to get the stream here and pause the tracks.. but send to sfu once joined then

santhoshvai:
closing this PR since we will remove pause and resume and to move to this API below 

```
enable()
disable({ stop }). -- by default stop is false
```

PR #950, opened 2023-08-17
feat: Disable doesn't stop audio tracks

szuperaz:
- `call.microphone.disable()` should only disable tracks, but not stop them
- `call.camera.disable()` is unchanged, it stops the tracks
- `pause` and `resume` methods are removed -> RN fails becuase of that, we should fix that before merge

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -1092,10 +1092,11 @@ export class Call {
    *
    *
    * @param trackType the track type to stop publishing.
+   * @param stopTrack if `true` the track will be stopped, else it will be just disabled
    */
-  stopPublish = async (trackType: TrackType) => {
+  stopPublish = async (trackType: TrackType, stopTrack: boolean = true) => {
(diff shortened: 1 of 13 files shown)


santhoshvai (approved):

santhoshvai:
we can ignore the RN test failure, that is a non relevant thing.. it will be fixed by @khushal87 later

szuperaz:
> we can ignore the RN test failure, that is a non relevant thing.. it will be fixed by @khushal87 later

Thanks @santhoshvai, I'll keep this open for @oliverlaz to check if something seems wrong in publisher

oliverlaz (approved):

oliverlaz:
> > we can ignore the RN test failure, that is a non relevant thing.. it will be fixed by @khushal87 later
> 
> Thanks @santhoshvai, I'll keep this open for @oliverlaz to check if something seems wrong in publisher

It looks good to me.

PR #951, opened 2023-08-17
fix(react-demo): media permissions and chat css

zwaardje:
Issues: 

 - https://stream-io.atlassian.net/browse/WD-813
 - https://stream-io.atlassian.net/browse/WD-814

### DIFF
--- a/sample-apps/react/react-video-demo/src/App.tsx
+++ b/sample-apps/react/react-video-demo/src/App.tsx
@@ -1,5 +1,10 @@
 import { useCallback, useEffect, useState } from 'react';
 import { v1 as uuidv1 } from 'uuid';
+import {
+  MediaPermissionsError,
+  MediaPermissionsErrorType,
+  requestMediaPermissions,
+} from 'mic-check';
 
 import {
   adjectives,
@@ -51,6 +56,9 @@ const Init = () => {
   const [callHasEnded, setCallHasEnded] = useState(false);
   const [storedDeviceSettings, setStoredDeviceSettings] =
     useState<LocalDeviceSettings>();
+  const [permissionsEnabled, setPermissionsEnabled] = useState<
+    boolean | undefined
+  >(undefined);
 
   const [isjoiningCall, setIsJoiningCall] = useState(false);
   const { setSteps } = useTourContext();
@@ -58,6 +66,16 @@ const Init = () => {
   const [client, setClient] = useState<StreamVideoClient>();
 
   const { edges, fastestEdge } = useEdges(client);
+
+  useEffect(() => {
+    requestMediaPermissions()
+      .then(() => {
+        setPermissionsEnabled(true);
+      })
+      .catch(() => {
+        setPermissionsEnabled(false);
+      });
+  }, []);
 
   useEffect(() => {
     const _client = new StreamVideoClient({
@@ -138,14 +156,20 @@ const Init = () => {
     return null;
   }
 
+  console.log('enabled?', permissionsEnabled);
+
   if (storedDeviceSettings) {
     return (
       <StreamVideo client={client}>
         <StreamCall
           call={activeCall}
           mediaDevicesProviderProps={{
-            initialVideoEnabled: !storedDeviceSettings?.isVideoMute,
-            initialAudioEnabled: !storedDeviceSettings?.isAudioMute,
+            initialVideoEnabled: permissionsEnabled
+              ? !storedDeviceSettings?.isVideoMute
+              : false,
+            initialAudioEnabled: permissionsEnabled
+              ? !storedDeviceSettings?.isAudioMute
+              : false,
             initialAudioInputDeviceId:
               storedDeviceSettings?.selectedAudioInputDeviceId,
             initialVideoInputDeviceId:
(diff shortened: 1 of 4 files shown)


sample-apps/react/react-video-demo/src/App.tsx
```diff
@@ -1,5 +1,10 @@
 import { useCallback, useEffect, useState } from 'react';
 import { v1 as uuidv1 } from 'uuid';
+import {
+  MediaPermissionsError,
+  MediaPermissionsErrorType,
+  requestMediaPermissions,
+} from 'mic-check';
```
MartinCupela, line 7: I would avoid installing 3rd party lib, when the browser provides API to check microphone and camera permissions. The `video-react-sdk` has hooks for this (https://github.com/GetStream/stream-video-js/blob/6fac2e53fde35b3c8cc5bbaf26cf8c2d0018127f/packages/react-sdk/src/hooks/useRequestPermission.ts#L5) 

@oliverlaz shouldn't we export those hooks?

sample-apps/react/react-video-demo/src/App.tsx
```diff
@@ -59,6 +67,16 @@ const Init = () => {
 
   const { edges, fastestEdge } = useEdges(client);
 
+  useEffect(() => {
+    requestMediaPermissions()
```
MartinCupela, line 71: I propose to keep loading state meanwhile we do not know, whether the browser grants permissions. That loading state can be then represented on the video preview in the lobby. Currently, when I load the page, first I see, that I do not have permissions, even though they are already granted and then the video or video fallback is shown.

sample-apps/react/react-video-demo/src/components/LobbyPanel/LobbyPanel.tsx
```diff
@@ -65,18 +66,21 @@ export const LobbyPanel: FC<Props> = ({
...
         <EnableBrowserSettings />
+      ) : (
+        <DisabledVideoPreview name={user.name} />
       ),
     );
 
   useEffect(() => {
-    if (!permissionsEnabled) {
+    console.log('lalal enabled', permissionsEnabled);
+    if (permissionsEnabled === false || permissionsEnabled === undefined) {
```
MartinCupela, line 78:
```suggestion
    if (!permissionsEnabled) {
```

sample-apps/react/react-video-demo/src/components/LobbyPanel/LobbyPanel.tsx
```diff
@@ -65,18 +66,21 @@ export const LobbyPanel: FC<Props> = ({
...
+      permissionsEnabled === false || permissionsEnabled === undefined ? (
         <EnableBrowserSettings />
+      ) : (
+        <DisabledVideoPreview name={user.name} />
       ),
     );
 
   useEffect(() => {
-    if (!permissionsEnabled) {
+    console.log('lalal enabled', permissionsEnabled);
```
MartinCupela, line 77:
```suggestion
```

sample-apps/react/react-video-demo/src/App.tsx
```diff
@@ -138,14 +156,20 @@ const Init = () => {
     return null;
   }
 
+  console.log('enabled?', permissionsEnabled);
```
MartinCupela, line 159:
```suggestion
```

sample-apps/react/react-video-demo/src/components/LobbyPanel/LobbyPanel.tsx
```diff
@@ -65,18 +66,21 @@ export const LobbyPanel: FC<Props> = ({
 }) => {
   const [permissionsErrorComponent, setPermissionsErrorComponent] =
     useState<any>(() =>
-      permissionsEnabled ? (
-        <DisabledVideoPreview name={user.name} />
-      ) : (
+      permissionsEnabled === false || permissionsEnabled === undefined ? (
         <EnableBrowserSettings />
+      ) : (
+        <DisabledVideoPreview name={user.name} />
```
MartinCupela, line 72: I am not sure, if I understand, why there are 3 possible outcomes for a boolean value:

`permissionEnabled ? A : !permissionEnabled ? B : A`

MartinCupela:
@zwaardje could you please remove `FC<any>` from the `EnableBrowserSettings` return value type? We should avoid declaring `any` as props for a component.

image

sample-apps/react/react-video-demo/src/components/LobbyPanel/LobbyPanel.tsx
```diff
@@ -115,8 +119,8 @@ export const LobbyPanel: FC<Props> = ({
         </div>
 
         <VideoPreview
-          DisabledVideoPreview={() => <DisabledVideoPreview name={user.name} />}
-          NoCameraPreview={() => <DisabledVideoPreview name={user.name} />}
+          DisabledVideoPreview={() => permissionsErrorComponent}
```
MartinCupela, line 122: I believe, we do not need to keep the component in a state, because the update would be triggered by the changed value of prop permissionEnabled. That means that we could do here:

```suggestion
          DisabledVideoPreview={(permissionsEnabled ? EnableBrowserSettings :  DisabledVideoPreview }
```

The `DisabledVideoPreview` can probably take the name from the `useUserContext();` and avoid prop drilling.

sample-apps/react/react-video-demo/src/App.tsx
MartinCupela: We should probably check for microphone and video permissions separately.

PR #956, opened 2023-08-18
feat(react-native): add permission alert messages to translations

santhoshvai:


packages/react-native-sdk/src/providers/MediaStreamManagement.tsx
```diff
@@ -94,36 +95,35 @@ export const MediaStreamManagement = ({
...
-      Alert.alert('Camera permission not granted, can not enable video');
+      Alert.alert(t('Camera Permission Required To Enable Video'));
       return false;
     }
 
     call?.camera.state.status === 'disabled'
       ? call?.camera.enable()
       : call?.camera.disable();
-  }, [call]);
+  }, [call, t]);
```
khushal87, line 126: Not sure if we should pass the `t` function in the usecallback dep array. Other than it the PR looks good.

vishalnarkhede (approved):

PR #958, opened 2023-08-18
fix(react-native): participant label should check for sfu mute state

santhoshvai:
currently it checks for local camera's mute

### DIFF
--- a/packages/react-native-sdk/src/components/Participant/ParticipantView/ParticipantLabel.tsx
+++ b/packages/react-native-sdk/src/components/Participant/ParticipantView/ParticipantLabel.tsx
@@ -2,14 +2,11 @@ import React from 'react';
 import { Pressable, StyleSheet, Text, View } from 'react-native';
 import { MicOff, PinVertical, ScreenShare, VideoSlash } from '../../../icons';
 import { theme } from '../../../theme';
-import {
-  useCall,
-  useCallStateHooks,
-  useI18n,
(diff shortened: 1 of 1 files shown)


vishalnarkhede (approved):

packages/react-native-sdk/src/components/Participant/ParticipantView/ParticipantLabel.tsx
```diff
@@ -26,17 +23,15 @@ export const ParticipantLabel = ({
...
-  const { status: micStatus } = useMicrophoneState();
-  const { status: cameraStatus } = useCameraState();
   const { t } = useI18n();
   const participantName = name ?? userId;
   const participantLabel = isLocalParticipant ? t('You') : participantName;
   const isPinningEnabled = pin?.isLocalPin;
-  const isAudioMuted = micStatus === 'disabled';
-  const isVideoMuted = cameraStatus === 'disabled';
+  const isAudioMuted = !publishedTracks.includes(SfuModels.TrackType.AUDIO);
+  const isVideoMuted = !publishedTracks.includes(SfuModels.TrackType.VIDEO);
```
szuperaz, line 34: This should be supported in the future, I think the only missing piece is to update camera/mic status when we are muted remotely, it's on my todo list

santhoshvai: awesome, thanks

PR #977, opened 2023-08-23
feat: apply target resolution to video feed, sync camera/mic init

szuperaz:
- Backend sends us the target resolution for the video stream -> the new device API now applies these settings
- Additional safeguards to make sure we don't accidentally start two media streams / publish twice when we init the camera/mic in the join method

@khushal87, @vishalnarkhede  can you please test that everything works with RN as expected?

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -1718,6 +1718,10 @@ export class Call {
   };
 
   private async initCamera() {
+    // Wait for any in progress camera operation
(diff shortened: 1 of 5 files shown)


khushal87:
Hey @szuperaz I tested it and couldn't find any issue on RN SDK as such because of this change.

oliverlaz (approved):

szuperaz:
thank you @khushal87

PR #1006, opened 2023-08-29
feat(react-native): handle participant switch in floating participant view for 1:1 calls

khushal87:
Preview - https://getstream.slack.com/archives/C04SAU8QUS0/p1693233632680799


packages/react-native-sdk/src/components/Participant/FloatingParticipantView/index.tsx
```diff
@@ -131,25 +143,27 @@ export const FloatingParticipantView = ({
...
+                participant={participant}
+                videoMode={'video'}
+                style={[
+                  styles.participantViewContainer,
+                  style,
+                  {
+                    backgroundColor: colors.static_grey,
+                    shadowColor: colors.static_black,
+                  },
+                  localParticipantsView.participantViewContainer,
```
vishalnarkhede, line 158: I think we need to change the name `localParticipantsView` in this case, since floating view is not associated with local participant only.

khushal87: Oh damn, I forgot to do it when moving the LocalParticipantView to FloatingParticipantView back then. I have pushed the change, so it should be fixed now. Nice catch.

packages/react-native-sdk/docusaurus/docs/reactnative/05-ui-cookbook/11-runtime-layout-switching.mdx
santhoshvai, line 1: why do this change? the mode is still in the call content props right

packages/react-native-sdk/src/components/Call/CallLayout/CallParticipantsGrid.tsx
```diff
@@ -31,19 +35,26 @@ export const CallParticipantsGrid = ({
...
+  const localParticipant = useLocalParticipant();
 
   const showFloatingView =
     remoteParticipants.length > 0 && remoteParticipants.length < 3;
 
-  const participants = showFloatingView ? remoteParticipants : allParticipants;
+  const participants = showFloatingView
+    ? showLocalParticipant
+      ? [localParticipant!!]
+      : remoteParticipants
```
santhoshvai, line 56:
```suggestion
    ? showLocalParticipant && localParticipant
      ? [localParticipant]
      : remoteParticipants
```

this is safer I think.. otherwise can cause a crash by using this component individually and passing showLocalParticipant to be true with having a local participant

packages/react-native-sdk/src/components/Call/CallLayout/CallParticipantsSpotlight.tsx
santhoshvai, line 1: very nice change! eliminating an unnecessary state, this is a huge benefit than what it seems 👏

santhoshvai (approved):

packages/react-native-sdk/docusaurus/docs/reactnative/05-ui-cookbook/11-runtime-layout-switching.mdx
khushal87: Yes, it also has undefined which is not relevant for tutorial that's why I changed.

packages/react-native-sdk/src/components/Call/CallLayout/CallParticipantsSpotlight.tsx
khushal87: Yes

packages/react-native-sdk/src/components/Call/CallLayout/CallParticipantsGrid.tsx
khushal87: Done

packages/react-native-sdk/docusaurus/docs/reactnative/05-ui-cookbook/11-runtime-layout-switching.mdx
santhoshvai: then we could do

```
NonNullable<CallContentProps['mode']>
```

PR #1013, opened 2023-08-29
feat: speaker management

szuperaz:
We should combine this with the dynascale PR to apply device id and volume settings ✅

### DIFF
--- a/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
+++ b/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
@@ -205,7 +205,10 @@ call.speaker.state.isDeviceSelectionSupported;
 // The error handler is called if the user denies permission to use audio
 call.speaker
   .listDevices()
-  .subscribe({ next: console.log, error: console.error });
+  .subscribe({ 
+    next: (devices) => console.log(devices), 
+    error: (err) => console.error(err),
+  });
 
 // Select device
 call.speaker.select('device Id');
--- a/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
+++ b/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
@@ -220,7 +220,10 @@ Here is how you can access the selected device:
 
 ```typescript
 call.speaker.state.selectedDevice; // currently selected audio output
-call.speaker.state.selectedDevice$.subscribe(console.log); // Reactive value for selected device, you can subscribe to changes
+// Reactive value for selected device, you can subscribe to changes
+call.speaker.state.selectedDevice$.subscribe((selectedDevice) => {
+  console.log(selectedDevice);
+});
 ```
 
 ### Set volume
--- a/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
+++ b/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
@@ -190,16 +190,49 @@ This is how you can show a visual representation about the sound level coming fr
 
 ## Speaker management
 
-### Browser support
-
-Selecting audio output device for the call [isn't supported by all browsers](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/sinkId), this is how you can check the availability:
-
-```typescript
-// This feature is coming soon 🏗️
-```
-
 ### List and select devices
 
+#### Browser support
+
+Selecting audio output device [isn't supported by all browsers](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/sinkId), this is how you can check the availability:
+
 ```typescript
-// This feature is coming soon 🏗️
+call.speaker.state.isDeviceSelectionSupported;
+```
+
+```typescript
+// List devices
+// The error handler is called if the user denies permission to use audio
+call.speaker
+  .listDevices()
+  .subscribe({ next: console.log, error: console.error });
+
+// Select device
+call.speaker.select('device Id');
+```
+
+Device id can also be `undefined`, that means to use the system's default audio output.
+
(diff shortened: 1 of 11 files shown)


szuperaz:
@khushal87, @vishalnarkhede or @santhoshvai please take a look if this makes sense for RN

packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
```diff
@@ -203,16 +203,49 @@ This is how you can show a visual representation about the sound level coming fr
...
+call.speaker.state.isDeviceSelectionSupported;
 ```
 
-### List and select devices
+```typescript
+// List devices
+// The error handler is called if the user denies permission to use audio
+call.speaker
+  .listDevices()
+  .subscribe({ next: console.log, error: console.error });
```
oliverlaz, line 221:
```suggestion
  .subscribe({ 
    next: (devices) => console.log(devices), 
    error: (err) => console.error(err),
  });
```

I think it helps if we make the subscription more explicit.

packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
```diff
@@ -203,16 +203,49 @@ This is how you can show a visual representation about the sound level coming fr
...
+```
+
+Device id can also be `undefined`, that means to use the system's default audio output.
+
+Here is how you can access the selected device:
 
 ```typescript
-// This feature is coming soon 🏗️
+call.speaker.state.selectedDevice; // currently selected audio output
+call.speaker.state.selectedDevice$.subscribe(console.log); // Reactive value for selected device, you can subscribe to changes
```
oliverlaz, line 233:
```suggestion
// Reactive value for selected device, you can subscribe to changes
call.speaker.state.selectedDevice$.subscribe((selectedDevice) => {
  console.log(selectedDevice);
});
```

oliverlaz (approved):

packages/client/src/devices/SpeakerManager.ts
```diff
@@ -0,0 +1,45 @@
...
+    }
+    this.state.setDevice(deviceId);
+  }
+
+  /**
+   * Set the volume of the audio elements
+   * @param volume a number between 0 and 1, `undefined` means to use the default volume level
+   */
+  setVolume(volume: number | undefined) {
+    if (volume && (volume < 0 || volume > 1)) {
```
khushal87, line 40: Can we add a check of React Native here too? Just for the safe side

packages/client/src/devices/SpeakerManager.ts
```diff
@@ -0,0 +1,45 @@
...
+      throw new Error('This feature is not supported in React Native');
+    }
+    this.state.setDevice(deviceId);
+  }
+
+  /**
+   * Set the volume of the audio elements
+   * @param volume a number between 0 and 1, `undefined` means to use the default volume level
+   */
+  setVolume(volume: number | undefined) {
```
khushal87, line 39: Can we add a check of React Native here too? Just for the safe side

packages/client/src/devices/__tests__/SpeakerManager.test.ts
```diff
@@ -0,0 +1,66 @@
+import { afterEach, beforeEach, describe, vi, it, expect } from 'vitest';
+import { mockAudioDevices } from './mocks';
+import { of } from 'rxjs';
+import { SpeakerManager } from '../SpeakerManager';
+import { checkIfAudioOutputChangeSupported } from '../devices';
+
+vi.mock('../devices.ts', () => {
+  console.log('MOCKING devices');
+  return {
+    getAudioOutputDevices: vi.fn(() => of(mockAudioDevices)),
```
khushal87, line 10: This could be a stupid doubt but, why are we using audio input devices for mocking the audio output devices? I see that the `mockAudioDevices` mostly contains the microphone or audio input devices.

packages/client/src/devices/SpeakerManager.ts
szuperaz: Added, thanks

packages/client/src/devices/__tests__/SpeakerManager.test.ts
szuperaz: This is because for testing purposes the actual data doesn't matter, only the shape of the data. Technically we could use the same test data for the camera and mic manager as well.

packages/client/src/devices/__tests__/SpeakerManager.test.ts
```diff
@@ -0,0 +1,66 @@
...
+    checkIfAudioOutputChangeSupported: vi.fn(() => true),
+  };
+});
+
+describe('SpeakerManager.test', () => {
+  let manager: SpeakerManager;
+
+  beforeEach(() => {
+    manager = new SpeakerManager();
+  });
```
arnautov-anton, line 20: This is a nitpick: we should probably start switching to test fixtures to allow for tests to run concurrently which in turn should make them run slightly faster and it looks slightly more nicer. :)

arnautov-anton (approved):

khushal87 (approved):

packages/client/src/devices/__tests__/SpeakerManager.test.ts
szuperaz: I don't know what those are, but sounds good 😄 if you have an example code, I'm happy to start using them

arnautov-anton: https://github.com/GetStream/stream-video-js/blob/96dfb0d5194ddc2efdff8e2ea495cb030775a687/sample-apps/react/egress-composite/tests/layouts.spec.ts#L1-L32

It's these - this is Playwright but I believe Vitest has similar looking API. It runs code within your "extend" definition and provides the data for each test instance defined later. :)

PR #1018, opened 2023-08-29
fix: fit the video in renderer

vishalnarkhede:


packages/react-native-sdk/src/components/Participant/ParticipantView/VideoRenderer.tsx
khushal87, line 1: Hey, with this change, this is what happens to the video stream. Do we really want it?

Screenshot_1693371398

khushal87 (commented):
Hey, with this change, this is what happens to the video stream. Do we really want it?

Screenshot_1693371398

vishalnarkhede:
Lets hold this for later ... we will need to check with other SDKs as well.

PR #1019, opened 2023-08-30
fix: bring back reanimated and gesture handler

santhoshvai:


packages/react-native-sdk/package.json
```diff
@@ -19,14 +19,25 @@
...
+    "!ios/build",
+    "!android/build",
+    "!android/gradle",
+    "!android/gradlew",
+    "!android/gradlew.bat",
+    "!android/local.properties",
+    "!**/__tests__",
+    "!**/__fixtures__",
+    "!**/__mocks__",
+    "!**/.*"
```
santhoshvai, line 40: I aligned this part with create react native library output 

After I did yarn pack and saw the changes and tested it to be working.. we were unnecessarily packing gradlew for example

packages/react-native-sdk/tsconfig.json
santhoshvai, line 1: the changes in this file took the most time to find out foe me. 

when it is "module": "commonjs",  whenever we mix import and require TS behaves weirdly. 

I noticed that module is actually kept to esnext in the create react native library output. Upon that change I saw it to be working.

We should have basically aligned this file with create react native library output from the beginning.

vishalnarkhede (approved):

PR #1022, opened 2023-08-30
fix(react-native): blank stream on join

santhoshvai:
creating local video stream at 2 places causes blank stream when joined


packages/client/src/rtc/codecs.ts
```diff
@@ -21,9 +21,9 @@ export const getPreferredCodecs = (
     const codec = c.mimeType.toLowerCase();
     logger?.('debug', `Found supported codec: ${codec}`);
     const shouldRemoveCodec =
-      codecToRemove && codec === `${kind}/${codecToRemove}`;
+      codecToRemove && codec === `${kind}/${codecToRemove.toLowerCase()}`;
```
santhoshvai, line 24: this is an unrelated fix, without this if case was upper case for codec passed, it did not work

PR #1039, opened 2023-08-31
feat: Allow reusing call instance [WIP]

oliverlaz:
### Overview

Allows a `call` instance to be reused for multiple `.join() -> .leave() -> .join()` cycles.

Based on #1011 - I'll update the base once #1011 is tested and merged.

### DIFF
--- a/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
+++ b/packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
@@ -176,16 +176,17 @@ Follow our [Playing Video and Audio guide](../../guides/playing-video-and-audio/
 
 #### Lobby preview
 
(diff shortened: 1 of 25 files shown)


szuperaz (commented):
Maybe we could add an `unwatch` method that unregisters coordinator WS event listeners. It could be useful if a customer doesn't care about a call instance anymore, and then `unwatch` should make it possible to GC the call instance.

oliverlaz:
> Maybe we could add an `unwatch` method that unregisters coordinator WS event listeners. It could be useful if a customer doesn't care about a call instance anymore, and then `unwatch` should make it possible to GC the call instance.

I don't think our Coordinator API has such an API. We can handle this in the client, but the backend will still send events for unwatched calls.

oliverlaz:
Implemented with: #1433

PR #1044, opened 2023-09-01
docs(react-native): add expo installation docs

santhoshvai:


packages/react-native-sdk/docusaurus/docs/reactnative/01-setup/02-installation/01-react-native.mdx
```diff
@@ -10,6 +9,10 @@ Installation and usage of our React Native SDK is simple and involves the follow
 First things first, make sure you have set up the development environment for React Native.
 You can find the official guide [here](https://reactnative.dev/docs/environment-setup).
 
+### Expo
+
+Stream's Video SDK is available on Expo through development builds. See the instructions found [here](../expo).
```
khushal87, line 14: Since we clearly show a new page and it is evident in the page contents, we can remove it.

packages/react-native-sdk/docusaurus/docs/reactnative/01-setup/02-installation/02-expo.mdx
```diff
@@ -0,0 +1,99 @@
...
+          "cameraPermission": "Allow $(PRODUCT_NAME) to access your camera",
+          "microphonePermission": "Allow $(PRODUCT_NAME) to access your microphone"
+        }
+      ]
+    ]
+  }
+}
+```
+
+You can now create development builds with `eas build --profile development --local`.
```
khushal87, line 99: We should add something about permissions here, either a link to keep the call alive and the native permissions page would also be fine. Currently I have no clue if we need the foreground permission is needed unless I am aware about the feature.

khushal87 (changes requested):

khushal87:
Since the steps are mostly the same we can club both installations together(and just add a section about config plugin coz that's the only different thing I find) and add but this is subjective. @vishalnarkhede your thoughts on it?

PR #1069, opened 2023-09-08
feat(react-native)!: set reactions through props and not config

khushal87:
The purpose of the PR is to update supported reactions for the call using a prop rather than a global config.

Currently, the reactions are used at two places in the call components hierarchy, the `CallContent` and the `ReactionsButton`, and thereby the `ReactionsPicker` internally. The `ReactionsButton` is not a part of the default design of `CallControls` by design for now hence integrators have the flexibility to add/not use the button at all.

In the PR, the new prop `reactions` is introduced in the CallContent that is drilled down to ParticipantReaction, which functions to show the participant's current reaction in the ParticipantView and FloatingParticipantView. A prop similar to `reactions` is introduced in ReactionsButton which will take up the supported reactions and pass them to the ReactionPicker. 

The idea is to keep the data the same for each `reactions` prop which adhere's this TS definition:
```ts
export type StreamReactionType = StreamReaction & {
  icon: string;
};
```

The data looks like this:
```ts
export const defaultEmojiReactions: StreamReactionType[] = [
  {
    type: 'reaction',
    emoji_code: ':like:',
    custom: {},
    icon: '👍',
  },
  {
    type: 'raised-hand',
    emoji_code: ':raise-hand:',
    custom: {},
    icon: '✋',
  },
  {
    type: 'reaction',
    emoji_code: ':fireworks:',
    custom: {},
    icon: '🎉',
  },
];
```

The idea to keep the data the same for both is, that we do not store the reactions that are passed in CallContent somewhere so that it can be used in a detached component like ReactionsButton. 

If we go with the model of React SDK that has emojiMap at some place and reactions at others you will face the bug as described here just because we don't store the map. Ref: https://getstream.slack.com/archives/C04ATV49DU3/p1694152109174239

Note: This PR also renames the ReactionButton to ReactionsButton same as React SDK.

PS: Docs will be added once the changes are approved.


packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
```diff
@@ -78,6 +90,7 @@ export const CallContent = ({
   ParticipantsInfoBadge,
   VideoRenderer,
   layout = 'grid',
+  reactions,
```
vishalnarkhede, line 93: I would recommend renaming it to `supportedReactions`.

khushal87: Done ✅

vishalnarkhede (approved):

PR #1072, opened 2023-09-11
fix(react-native): remove  method to inform SDK about native permissions

santhoshvai:
With the new device API, the devices are fetched inside the client. With which the webrtc will automatically ask for permissions for mic and camera.

So the method to inform SDK about native permissions must be removed after using the new device API.

## Additonal important changes

* The Call settings from BE for initial device states are now taken into account (very useful for live stream host for example - video/audio gets enabled automatically)
* Relevant Documentation updated
* Fixed this UI bug (was caused when device status was undefined) 
image


packages/react-native-sdk/docusaurus/docs/reactnative/03-core/08-native-permissions.mdx
```diff
@@ -3,38 +3,25 @@ id: native-permissions
...
+Once the function is called, we should see permissions being requested like below:
 
 ![Preview of the final result](../assets/03-core/08-native-permissions/permissions.png)
 
 ## Setup
 
 Ensure that relevant permissions are declared in your `AndroidManifest.xml` and `Info.plist` as mentioned in the [installation](../setup/installation/) guide. 
 
-Additionally, to request permissions in both platforms easily we will use the [`react-native-permissions`](https://github.com/zoontek/react-native-permissions) library. You can run the following command to install it:
+Additionally, to request permissions on both platforms easily we will use the [`react-native-permissions`](https://github.com/zoontek/react-native-permissions) library. You can run the following command to install it:
```
vishalnarkhede, line 16:
```suggestion
Additionally, to easily request permissions on both platforms, we will use the [`react-native-permissions`](https://github.com/zoontek/react-native-permissions) library. You can run the following command to install it:
```

vishalnarkhede (approved):

PR #1075, opened 2023-09-12
fix(react-native): add missing commonjs support

santhoshvai:
# Problem

Our SDK lacks common js support. Also expo config plugin works only with common JS. 

This PR fixes them both.

## Testing

I tested the build to work by packing it in `yarn pack`

And putting the path to the SDK in package.json of a completely new app 

for example:

```
 "@stream-io/video-react-native-sdk": "file:/Users/santhoshvaiyapuri/Projects/stream-video-js/packages/react-native-sdk/package.tgz",
```

### DIFF
--- a/packages/react-native-sdk/package.json
+++ b/packages/react-native-sdk/package.json
@@ -2,8 +2,10 @@
   "name": "@stream-io/video-react-native-sdk",
   "version": "0.0.9",
   "packageManager": "yarn@3.2.4",
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "main": "dist/commonjs/index.js",
+  "module": "dist/module/index.js",
+  "react-native": "src/index.ts",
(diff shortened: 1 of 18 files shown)


packages/react-native-sdk/package.json
```diff
@@ -2,8 +2,10 @@
   "name": "@stream-io/video-react-native-sdk",
   "version": "0.0.9",
   "packageManager": "yarn@3.2.4",
-  "main": "./dist/index.js",
-  "types": "./dist/index.d.ts",
+  "main": "dist/commonjs/index.js",
+  "module": "dist/module/index.js",
+  "react-native": "src/index.ts",
```
vishalnarkhede, line 7: noiceee!! Finally I can cmd + click on function to go to its definition directly 😄

vishalnarkhede (approved):

PR #1082, opened 2023-09-13
fix: enable GTM for the video demo

oliverlaz:
### Overview

Google TagManager didn't initialize properly during build time. This PR aims to fix that.

### DIFF
--- a/.github/workflows/copy-react-demo-s3.yml
+++ b/.github/workflows/copy-react-demo-s3.yml
@@ -57,11 +57,11 @@ jobs:
 
       - name: Build React Demo staging
         if: ${{ github.ref_name != 'main' }}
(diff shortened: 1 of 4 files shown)


.github/workflows/copy-react-demo-s3.yml
```diff
@@ -57,11 +57,11 @@ jobs:
 
       - name: Build React Demo staging
         if: ${{ github.ref_name != 'main' }}
-        run: yarn build:react:video-demo --mode staging
+        run: yarn build:react:video-demo
```
zwaardje, line 60: The gleap api key depends on the mode set to "staging". Gleap will only be visible on staging.

zwaardje (changes requested):

PR #1089, opened 2023-09-15
fix(DynascaleManager): update subscription upon cleanup

arnautov-anton:
- removing registered element from the screen wouldn't update subscriptions
- subscribing to `participant$` within `bindVideoElement` would call `map` operator 3 times on each `call.state.participants$` update per `bindVideoElement` call (per `sessionId`)

### DIFF
--- a/packages/client/src/helpers/DynascaleManager.ts
+++ b/packages/client/src/helpers/DynascaleManager.ts
@@ -16,6 +16,7 @@ import {
   distinctUntilChanged,
   distinctUntilKeyChanged,
   map,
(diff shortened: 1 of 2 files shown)


oliverlaz (approved):

oliverlaz (approved):
Can you also add an assertion in the tests that verifies the state is updated once clean-up is done?

arnautov-anton:
> Can you also add an assertion in the tests that verifies the state is updated once clean-up is done?

Good catch, done! 👍

PR #1090, opened 2023-09-15
feat(SpeakerLayout): add participantsBarLimit

arnautov-anton:
Adds `participantsBarLimit` property to `SpeakerLayout` component which hard-limits number of rendered participants in the participants bar.

### DIFF
--- a/packages/react-sdk/src/core/components/CallLayout/SpeakerLayout.tsx
+++ b/packages/react-sdk/src/core/components/CallLayout/SpeakerLayout.tsx
@@ -1,5 +1,4 @@
 import { useEffect, useState } from 'react';
-
 import {
   CallTypes,
   combineComparators,
@@ -11,6 +10,7 @@ import {
   StreamVideoParticipant,
 } from '@stream-io/video-client';
 import { useCall, useCallStateHooks } from '@stream-io/video-react-bindings';
+import clsx from 'clsx';
 
 import {
   DefaultParticipantViewUI,
@@ -22,8 +22,8 @@ import {
   useHorizontalScrollPosition,
   useVerticalScrollPosition,
 } from '../../../hooks';
-import clsx from 'clsx';
 import { useCalculateHardLimit } from '../../hooks/useCalculateHardLimit';
+import { Audio } from '../Audio';
 
 export type SpeakerLayoutProps = {
   ParticipantViewUISpotlight?: ParticipantViewProps['ParticipantViewUI'];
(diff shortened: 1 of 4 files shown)


packages/react-sdk/src/core/hooks/useCalculateHardLimit.ts
```diff
@@ -0,0 +1,64 @@
...
+    if (
+      !hostElement ||
+      !wrapperElement ||
+      typeof limit === 'number' ||
+      typeof limit === 'undefined'
+    )
+      return;
+
+    let childWidth = 280;
+    let childHeight = 160;
```
oliverlaz, line 32: Leftovers from testing?

packages/react-sdk/src/core/components/CallLayout/SpeakerLayout.tsx
```diff
@@ -32,6 +33,11 @@ export type SpeakerLayoutProps = {
    * Providing `null` will hide the bar.
    */
   participantsBarPosition?: 'top' | 'bottom' | 'left' | 'right' | null;
+  /**
+   * Hard limits the number of the participants rendered in the participants bar.
+   * Providing string `dynamic` will calculate hard limit based on screen width/height.
+   */
+  participantsBarLimit?: 'dynamic' | number;
```
oliverlaz, line 40: Why not make this an optional number? I think it is already used like that. By default it is `undefined` but when provided, we use the value. I don't think we need a `dynamic` as a value here.

arnautov-anton: The functionality can reach three states - without the limit (undefined) - it's used with the normal - scrolling behaviour where no limitation is applied, number to limit it to that specific amount and "dynamic" to calculate the amount based on the screen width/height which will render maximal amount that fits into the parent and does not trigger scrolling behavior. Optional number would mean to limit it only to two of the aforemention behaviours.

packages/react-sdk/src/core/hooks/useCalculateHardLimit.ts
arnautov-anton: Defaults to allow at least one element to render to get proper values for future calculations but now looking at it I'd rather do `calculatedLimit || 1` in the return statement.

arnautov-anton: Resolved here: https://github.com/GetStream/stream-video-js/pull/1090/commits/6537678ace8f208f30acce565cc1cfe4b76be864

oliverlaz (approved):

PR #1101, opened 2023-09-22
fix(react-native): device unmute and mute behaviour using the new device API

khushal87:
This PR tends to solve the `track.enabled` not changing to be `true` when unmuting the stream because it was previously done by the `unPublishStream` call. This PR tends to solve that behavior.

Note: The PR doesn't solve the bug of the microphone issue on RN yet but improves it one step further.

### DIFF
--- a/packages/client/src/devices/InputMediaDeviceManager.ts
+++ b/packages/client/src/devices/InputMediaDeviceManager.ts
@@ -141,25 +141,6 @@ export abstract class InputMediaDeviceManager<
 
   protected abstract getTrack(): undefined | MediaStreamTrack;
 
-  protected async muteStream(stopTracks: boolean = true) {
-    if (!this.state.mediaStream) {
-      return;
-    }
-    this.logger('debug', `${stopTracks ? 'Stopping' : 'Disabling'} stream`);
-    if (this.call.state.callingState === CallingState.JOINED) {
-      await this.stopPublishStream(stopTracks);
-    }
-    this.muteLocalStream(stopTracks);
-    if (this.getTrack()?.readyState === 'ended') {
-      // @ts-expect-error release() is present in react-native-webrtc and must be called to dispose the stream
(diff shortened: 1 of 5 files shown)


packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -191,16 +172,32 @@ export abstract class InputMediaDeviceManager<
...
+  }
+
   protected async unmuteStream() {
     this.logger('debug', 'Starting stream');
     let stream: MediaStream;
     if (this.state.mediaStream && this.getTrack()?.readyState === 'live') {
       stream = this.state.mediaStream;
       this.unmuteTrack();
     } else {
-      if (this.state.mediaStream) {
```
khushal87, line 201: As discussed with @szuperaz, this is not at all called in any case of the `unmuteStream` execution. Hence, it is removed.

packages/client/src/rtc/Publisher.ts
```diff
@@ -305,21 +305,26 @@ export class Publisher {
...
       transceiver.sender.track &&
       (stopTrack
         ? transceiver.sender.track.readyState === 'live'
         : transceiver.sender.track.enabled)
     ) {
-      stopTrack
-        ? transceiver.sender.track.stop()
-        : (transceiver.sender.track.enabled = false);
+      // Since React Native follows the new Device API this is not needed
+      if (!isReactNative()) {
```
szuperaz, line 317: Does that mean that RN doesn't stop publishing, just sends a mute state update to the SFU?

khushal87: Agh no, the purpose here is there was a repetition in the logic of muting here and at https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/devices/InputMediaDeviceManager.ts#L163, hence the mute in the `muteTrack` function doesn't run at all. Ideally, it should run in only one place. Because of this, the enabled always remained false even after update.

PR #1108, opened 2023-09-26
fix: add type check of deviceId before setting sinkId

arnautov-anton:
Prevent setting `undefined` as `sinkId` which caused issues.

image


oliverlaz (approved):

szuperaz (approved):
How am I co-author in this commit? 🤔

arnautov-anton:
> How am I co-author in this commit? 🤔

magic

arnautov-anton:
> How am I co-author in this commit? 🤔

You helped a great deal with the debugging so it's only fair that I added you as co-author. :)

szuperaz:
> > How am I co-author in this commit? 🤔
> 
> You helped a great deal with the debugging so it's only fair that I added you as co-author. :)

Ah, that's very nice of you, thanks.

PR #1110, opened 2023-09-26
fix(react-native): initial media stream management according to BE and SDK settings

khushal87:
Currently, the initial media stream doesn't account for the fact that local SDK settings. That is fixed in the PR. The BE settings are also applied and the final result is accounted for when the call is joined. The priority is Local SDK settings > BE settings > User manual mutes/unmutes.

Note: The PR is tested both on android and iOS and works fine for all the cases.

### DIFF
--- a/packages/react-native-sdk/src/providers/MediaStreamManagement.tsx
+++ b/packages/react-native-sdk/src/providers/MediaStreamManagement.tsx
@@ -50,25 +50,6 @@ export const MediaStreamManagement = ({
 
   const settings = useCallSettings();
 
-  // Apply SDK settings to set the initial audio/video enabled state
-  useEffect(() => {
-    setInitialDeviceState((prev) => {
-      let initAudio = prev.initialAudioEnabled;
-      if (typeof propInitialAudioEnabled !== 'undefined') {
-        initAudio = propInitialAudioEnabled;
-      }
-      let initVideo = prev.initialVideoEnabled;
-      if (typeof propInitialVideoEnabled !== 'undefined') {
-        initVideo = propInitialVideoEnabled;
-      }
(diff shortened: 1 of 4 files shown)


packages/react-native-sdk/src/providers/MediaStreamManagement.tsx
```diff
@@ -71,121 +43,74 @@ export const MediaStreamManagement = ({
   );
 
   const [{ initialAudioEnabled, initialVideoEnabled }, setInitialDeviceState] =
-    useState({
-      initialAudioEnabled: !!propInitialAudioEnabled,
-      initialVideoEnabled: !!propInitialVideoEnabled,
+    useState<MediaDevicesInitialState>({
+      initialAudioEnabled: undefined,
+      initialVideoEnabled: undefined,
     });
```
santhoshvai, line 49: need not change this part though.. the previous one is alright according to me.. its initilaised with the prop that is first passed and never undefined after

packages/react-native-sdk/src/providers/MediaStreamManagement.tsx
```diff
@@ -71,121 +43,74 @@ export const MediaStreamManagement = ({
...
-      initialAudioEnabled &&
-      (call?.microphone.state.status === undefined ||
-        call?.microphone.state.status === 'disabled')
-    ) {
+    if (initialAudioEnabled === undefined) {
+      return;
+    }
+    if (initialVideoEnabled === undefined) {
+      return;
+    }
```
santhoshvai, line 101: can be removed if the state is never made to be undefined

santhoshvai (approved):

PR #1111, opened 2023-09-27
fix(client): optimized device enumeration

khushal87:
### Overview

Optimizes the device enumeration and utilizes the `getUserMedia()` API only when necessary.


packages/client/src/devices/devices.ts
```diff
@@ -12,22 +12,47 @@ import {
...
+      // in that case, we need to request permissions (via getUserMedia)
+      // to be able to get the device labels
+      const needsGetUserMedia = devices.some(
+        (device) => device.kind === kind && device.label === '',
+      );
+      if (needsGetUserMedia) {
+        let mediaStream: MediaStream | undefined;
+        try {
+          mediaStream = await navigator.mediaDevices.getUserMedia(constraints);
+          devices = await navigator.mediaDevices.enumerateDevices();
```
khushal87, line 39: We will already have devices info(we call enumerateDevices) as per line 28, isn't it?

oliverlaz: Yes, we do, but some browsers require "getUserMedia" to be called before we can retrieve the device's label.

That is why we first enumerate the devices naively, and we check if their labels are there. If not, we retrieve a temporary MediaStream (with getUserMedia) and we run the device enumeration again.

khushal87: Ohh, about the browser thing that was clear to me, about calling the `device enumeration again` was unclear. But now it is. Thanks @oliverlaz 😄

szuperaz (approved):

oliverlaz (approved):

PR #1118, opened 2023-09-29
feat: ScreenShare Audio support

oliverlaz:
### Overview

Adds support for Screen Share Audio tracks in React and Plain-JS SDKs.
Additionally, certain Screen Share parameters (FPS, max bitrate) can be configured by integrators as well.


packages/client/src/devices/InputMediaDeviceManagerState.ts
```diff
@@ -1,16 +1,20 @@
...
-export abstract class InputMediaDeviceManagerState {
+export abstract class InputMediaDeviceManagerState<C = MediaTrackConstraints> {
   protected statusSubject = new BehaviorSubject<InputDeviceStatus>(undefined);
   protected mediaStreamSubject = new BehaviorSubject<MediaStream | undefined>(
     undefined,
   );
   protected selectedDeviceSubject = new BehaviorSubject<string | undefined>(
     undefined,
   );
+  protected defaultConstraintsSubject = new BehaviorSubject<C | undefined>(
```
szuperaz, line 14: What's the aim of this object? As far as I can tell, it's always undefined

oliverlaz: It should allow easier override of the default constraints. Example:

```
call.microphone.setDefaultConstraints({
  echoCancellation: true,
})
```

One method is missing in this implementation, I'll add it now: `InputMediaDeviceManager#setDefaultConstraints`.

packages/client/src/devices/InputMediaDeviceManagerState.ts
```diff
@@ -1,16 +1,20 @@
...
+    undefined,
+  );
```
arnautov-anton, line 16: nit: BehaviorSubject -> ReplaySubject?

arnautov-anton: Every subject is initialized with `undefined` state so that was the idea - still not sure if using RS over BS when initial state is `undefined` is a generally accepted good practice or just something I saw us do once and it stuck with me - it's fine either way, honestly. 😄

packages/client/src/devices/InputMediaDeviceManagerState.ts
szuperaz: I see, it makes sense now, thanks

packages/client/src/devices/InputMediaDeviceManagerState.ts
szuperaz: We use `BehaviorSubject` to have an initial value. If we were to use `ReplaySubject`, the `getCurrentValue` wouldn't be able to return a value unless the `ReplaySubject` emits for the first time, we would have to add a fallback value there. 

I think this is a safer approach because that way, each state variable has to define its own default value, even if that's `undefined` in the majority of the cases.

szuperaz (approved):

santhoshvai (approved):
Tested on iOS RN 💯

PR #1124, opened 2023-10-03
fix: check if `currentParticipant` is actually initialized

oliverlaz:
### Overview

Rendering `<LivestreamLayout />` before participants are initialized may lead to various errors where some of the rendered participants are `undefined` while TS says otherwise. 
Until we sort out the types, we are adding an explicit null check to the `currentParticipant` variable.

### DIFF
--- a/packages/react-sdk/src/core/components/CallLayout/LivestreamLayout.tsx
+++ b/packages/react-sdk/src/core/components/CallLayout/LivestreamLayout.tsx
@@ -108,9 +108,7 @@ export const LivestreamLayout = (props: LivestreamLayoutProps) => {
             hasOngoingScreenShare &&
               'str-video__livestream-layout__floating-participant',
             (hasOngoingScreenShare &&
-              floatingParticipantProps?.position &&
-              `str-video__livestream-layout__floating-participant--${floatingParticipantProps.position}`) ??
-              'str-video__livestream-layout__floating-participant--top-right',
+              `str-video__livestream-layout__floating-participant--${floatingParticipantProps?.position ?? 'top-right'}`),
           )}
           participant={currentSpeaker}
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/core/components/CallLayout/LivestreamLayout.tsx
```diff
@@ -102,19 +102,21 @@ export const LivestreamLayout = (props: LivestreamLayoutProps) => {
...
-      />
+      {currentSpeaker && (
+        <ParticipantView
+          className={clsx(
+            hasOngoingScreenShare &&
+              'str-video__livestream-layout__floating-participant',
+            (hasOngoingScreenShare &&
+              floatingParticipantProps?.position &&
+              `str-video__livestream-layout__floating-participant--${floatingParticipantProps.position}`) ??
+              'str-video__livestream-layout__floating-participant--top-right',
```
arnautov-anton, line 113:
```suggestion
              `str-video__livestream-layout__floating-participant--${floatingParticipantProps.position ?? 'top-right'}`),
```
Shorter and still readable I think. 🤔

arnautov-anton (approved):

packages/react-sdk/src/core/components/CallLayout/LivestreamLayout.tsx
arnautov-anton: (consider it a nitpick) 😄

oliverlaz: good point. I just updated your suggestion as well:
- before `floatingParticipantProps.position ?? 'top-right'`
- after `floatingParticipantProps?.position ?? 'top-right'`

PR #1130, opened 2023-10-05
fix: ensure stable sort

oliverlaz:
### Overview

Ensure Stable Sorting by mutating the source participant array.

### DIFF
--- a/packages/client/src/store/__tests__/CallState.test.ts
+++ b/packages/client/src/store/__tests__/CallState.test.ts
@@ -69,7 +69,7 @@ describe('CallState', () => {
 
       const ps2 = state.participants;
       // should resolve in initial - non-mutated state as set at the beginning
-      expect(ps2.map((p) => p.name)).toEqual(['A', 'B', 'C', 'D', 'E', 'F']);
+      expect(ps2.map((p) => p.name)).toEqual(['F', 'B', 'E', 'A', 'C', 'D']);
     });
 
     it('should support custom sorting', () => {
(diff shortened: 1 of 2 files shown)


packages/client/src/store/__tests__/CallState.test.ts
```diff
@@ -69,7 +69,7 @@ describe('CallState', () => {
 
       const ps2 = state.participants;
       // should resolve in initial - non-mutated state as set at the beginning
-      expect(ps2.map((p) => p.name)).toEqual(['A', 'B', 'C', 'D', 'E', 'F']);
+      expect(ps2.map((p) => p.name)).toEqual(['F', 'B', 'E', 'A', 'C', 'D']);
```
arnautov-anton, line 72:
```suggestion
      expect(ps2.map((p) => p.name)).toEqual(['F', 'B', 'E', 'A', 'C', 'D']);
```

Can we do `.toMatchInlineSnapshot` instead ([like here](https://github.com/GetStream/stream-video-js/blob/d86be1b6de25391996ff8dd8b093e2fb3d4c5dba/packages/client/src/sorting/__tests__/presets.test.ts#L17-L27))? The outputs that `vitest` gives when it compares arrays is super unreadable and updating it manually is super annoying (not that we should do that often but still).

arnautov-anton (approved):

packages/client/src/store/__tests__/CallState.test.ts
oliverlaz: I'd rather do that at some other time. There are plenty of places we would need to align.

PR #1131, opened 2023-10-06
fix: Video Demo improvements

oliverlaz:


sample-apps/react/react-video-demo/src/components/ParticipantsSlider/ParticipantsSlider.tsx
```diff
@@ -2,8 +2,11 @@ import { FC, useEffect, useState } from 'react';
 import { FreeMode, Grid as GridModule, Mousewheel, Navigation } from 'swiper';
 import { Swiper, SwiperSlide } from 'swiper/react';
 import classnames from 'classnames';
-import { Call, StreamVideoParticipant } from '@stream-io/video-react-sdk';
-import { SfuModels } from '@stream-io/video-client';
```
szuperaz, line 6: hm, it's interesting to me how this import used to work when the package doesn't list `@stream-io/video-client` as a dependency

szuperaz (approved):

sample-apps/react/react-video-demo/src/components/ParticipantsSlider/ParticipantsSlider.tsx
oliverlaz: it works because `video-client` is available in the root `node_modules`. If we moved the sample app outside of our repo it wouldn't have worked. We should have an eslint rule that prevents these wrong imports.

szuperaz: I see, thanks for the explanation

PR #1133, opened 2023-10-06
feat(react-native): add picture-in-picture support for Android

santhoshvai:
# picture-in-picture support for Android

- [x] Our Native module exposes methods and events to manage PiP in Android
- [x] Docs for PiP
- [x] Expo config plugin support
- [x] Do not disable camera when on PiP


packages/react-native-sdk/expo-config-plugin/src/withAndroidManifest.ts
```diff
@@ -46,8 +52,29 @@ const withStreamVideoReactNativeSDKManifest: ConfigPlugin = (configuration) => {
       );
       services.push(getNotifeeService());
       mainApplication.service = services;
+
+      if (props.androidPictureInPicture) {
+        const mainActivity = getMainActivityOrThrow(androidManifest);
+        ('keyboard|keyboardHidden|orientation|screenSize|uiMode');
```
vishalnarkhede, line 58: Does this do anything?

vishalnarkhede (approved):

packages/react-native-sdk/expo-config-plugin/src/withAndroidManifest.ts
santhoshvai: What this says is.. "dont kill my screen even if these things change.. I can handle it "

it is the android docs to do this if PiP is to be enabled

PR #1142, opened 2023-10-13
fix(client): skip broken update call types test

khushal87:


packages/client/src/__tests__/server-side/call-types.test.ts
```diff
@@ -92,22 +89,22 @@ describe('call types CRUD API', () => {
     });
   });
 
-  it('update', async () => {
```
oliverlaz, line 95: I'd rather use `it.skip` instead of commenting out the whole test.

khushal87: Agh okay, let me fix this

oliverlaz (approved):

PR #1157, opened 2023-10-20
chore: fix e2e failure on not finding the info button randomly

santhoshvai:
occassionally e2e fails on not finding the button

https://github.com/GetStream/stream-video-js/actions/runs/6587748280/job/17899157544

I think its because sometimes the call takes a while to join..

### DIFF
--- a/packages/react-native-sdk/src/components/Call/CallTopView/ParticipantsInfoBadge.tsx
+++ b/packages/react-native-sdk/src/components/Call/CallTopView/ParticipantsInfoBadge.tsx
@@ -80,7 +80,6 @@ export const ParticipantsInfoBadge = ({
           },
           participantInfoBadge.participantCountContainer,
         ]}
-        testID={ButtonTestIds.PARTICIPANTS_COUNT}
       >
         <Text
           style={[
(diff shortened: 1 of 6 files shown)


packages/react-native-sdk/src/components/Call/CallTopView/ParticipantsInfoBadge.tsx
```diff
@@ -90,6 +91,7 @@ export const ParticipantsInfoBadge = ({
             typefaces.subtitle,
             participantInfoBadge.participantsCountText,
           ]}
+          testID={ButtonTestIds.PARTICIPANTS_COUNT_TEXT}
```
testableapple, line 94: For some reason, iOS does not see the text inside this element, but it sees it in the parent element. That's why I had to add two different locators and access them in the tests through regex `participants-count-.*`.

testableapple (approved):
🙌

sample-apps/react-native/dogfood/e2e/test.yaml
```diff
@@ -2,6 +2,10 @@ appId: io.getstream.rnvideosample
 ---
 - runScript: js/setup.js
 - launchApp
+- runScript:
+    file: js/sleep.js
+    env:
+      seconds: 5
```
santhoshvai, line 8:
```suggestion
#- waits for ios system alert to pop up, then waits for maestro to handle it
- runScript:
    file: js/sleep.js
    env:
      seconds: 5
```

PR #1165, opened 2023-10-27
fix(react-native-dogfood): do not allow invalid call ids

khushal87:
Added a simple validation for custom call IDs in meeting and livestream join flow.

The call ID strings with alphabets, numbers, hyphens(-), and underscore(_) in them are only allowed and enabled the button for a call.

Regex Used for validation - `/^[A-Za-z0-9_-]*$/g`

### DIFF
--- a/sample-apps/react-native/dogfood/src/screens/LiveStream/JoinLiveStream.tsx
+++ b/sample-apps/react-native/dogfood/src/screens/LiveStream/JoinLiveStream.tsx
@@ -51,11 +51,9 @@ export const JoinLiveStream = ({
     flexDirection: orientation === 'landscape' ? 'row' : 'column',
   };
 
-  const isValidCallId = () => {
-    // Allows only alphabets, numbers, -(hyphen) and _(underscore)
-    const callIdRegex = /^[A-Za-z0-9_-]*$/g;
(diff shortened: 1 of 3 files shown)


sample-apps/react-native/dogfood/src/screens/LiveStream/JoinLiveStream.tsx
```diff
@@ -51,6 +51,12 @@ export const JoinLiveStream = ({
     flexDirection: orientation === 'landscape' ? 'row' : 'column',
   };
 
+  const isValidCallId = () => {
+    // Allows only alphabets, numbers, -(hyphen) and _(underscore)
+    const callIdRegex = /^[A-Za-z0-9_-]*$/g;
+    return callId && callId.match(callIdRegex);
+  };
```
vishalnarkhede, line 58:
```suggestion
  const isValidCallId = callId && callId.match(/^[A-Za-z0-9_-]*$/g);
```

This doesn't need to be a function I think

khushal87: Fixed

vishalnarkhede (approved):

PR #1174, opened 2023-10-31
feat: handle device disconnection

szuperaz:
This PRs adds the last missing piece to the new device management API: handling device disconnections.

### Audio, video manager
- If a device is disconnected, we deselect that device (not available in RN)
- If a track ends, we update the device manager status to `disabled`
- If the "default" device is replaced, we restart the device (not available in RN)

### Screenshare manager
- If a track ends, we update the device manager status to `disabled`

### Speaker manager
- If a device is disconnected, we deselect that device (not available in RN)

### Watching for disconnected devices
The `deviceIds$` Observable notifies the manager about device connections/disconnections. This Observable uses `enumerateDevices`, so it will never ask for permissions (we only need to ask for permissions if we need device labels, but we don't need them for watching disconnected devices).

### Handling track ending
We add the `ended` event handler to all tracks of a media stream, if any track ends, we update the manager's status to `disabled`.

### Handling new default device
If a device's `deviceId` stays the same, but it's `groupId` is changed, a device is replaced. We restart the stream if the device was previously enabled.

⚠️ The PR also removed unused device helper methods, that are remainings of the old device API that's already removed, on the offchance a customer uses any of these helpers they should upgrade to the new, simlified device management API:
- For JS client: https://getstream.io/video/docs/javascript/guides/camera-and-microphone/
- For React: https://getstream.io/video/docs/react/guides/camera-and-microphone/
- For RN: https://getstream.io/video/docs/reactnative/core/camera-and-microphone/

### DIFF
--- a/packages/client/src/devices/devices.ts
+++ b/packages/client/src/devices/devices.ts
@@ -258,17 +258,34 @@ export const getScreenShareStream = async (
   }
 };
 
-const getDeviceIds = memoizedObservable(() =>
-  merge(
-    from(navigator.mediaDevices.enumerateDevices()),
-    getDeviceChangeObserver(),
-  ).pipe(shareReplay(1)),
-);
+export const deviceIds$ =
+  typeof navigator !== 'undefined' &&
+  typeof navigator.mediaDevices !== 'undefined'
+    ? memoizedObservable(() =>
+        merge(
+          from(navigator.mediaDevices.enumerateDevices()),
+          getDeviceChangeObserver(),
+        ).pipe(shareReplay(1)),
+      )()
+    : undefined;
 
 export const watchForDisconnectedDevice = (
+  kind: MediaDeviceKind,
   deviceId$: Observable<string | undefined>,
 ) => {
-  return combineLatest([getDeviceIds(), deviceId$]).pipe(
+  let devices$;
+  switch (kind) {
+    case 'audioinput':
+      devices$ = getAudioDevices();
+      break;
+    case 'videoinput':
+      devices$ = getVideoDevices();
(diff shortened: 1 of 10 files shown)


packages/client/src/devices/devices.ts
```diff
@@ -289,6 +300,8 @@ const watchForDisconnectedDevice = (
  * @angular It's recommended to use the [`DeviceManagerService`](./DeviceManagerService.md) for a higher level API, use this low-level method only if the `DeviceManagerService` doesn't suit your requirements.
  * @param deviceId$ an Observable that specifies which device to watch for
  * @returns
+ *
+ * @deprecated use the [new device API](https://getstream.io/video/docs/javascript/guides/camera-and-microphone/)
```
oliverlaz, line 304: Why not completely remove these APIs, as we are now presenting a better alternative?

szuperaz: I think technically, that would be a breaking change, we export these methods, so if someone uses them, it will lead to an error on their side. That's why I only deprecated them

szuperaz: With the previous device API, you needed to use these methods in your own code

oliverlaz: they were used in React's specific device management API before. Perhaps I should have removed them as part of #1127.

szuperaz: I think we didn't know we won't need them. But in any case, I think it's fine to remove them in the next breaking release, in practice probably no one uses them.

khushal87 (approved):

PR #1175, opened 2023-10-31
fix: respect server-side settings in the lobby

oliverlaz:
### Overview

- Respect server-side settings for mic and camera on the lobby screen
- Fix ScreenShareButton class names
- Fix ScreenSharing in Video Demo


testableapple (approved):
✅

PR #1178, opened 2023-11-03
fix: set `key` prop to the correct element

oliverlaz:
### Overview

Follow up on #1176. Sets the `key` prop on the top-most `Fragment` instead to the wrapped elements.


packages/react-sdk/src/core/components/Audio/ParticipantsAudio.tsx
```diff
@@ -46,15 +45,14 @@ export const ParticipantsAudio = (props: ParticipantsAudioProps) => {
               {...audioProps}
               trackType="screenShareAudioTrack"
               participant={participant}
-              key={`${sessionId}-screen-share-audio`}
             />
           );
 
         return (
-          <>
+          <Fragment key={sessionId}>
```
arnautov-anton, line 52: NextJS thing? 🤔

arnautov-anton (approved):

PR #1184, opened 2023-11-09
feat(device-api): Browser Permissions API

oliverlaz:
### Overview

Support for querying browser permissions for camera and microphone.

### Breaking changes (React-SDK):
- `useHasBrowserPermissions` is not available in the SDK anymore. Please use the appropriate helper hooks:
```typescript
const { useMicrophoneState, useCameraState } = useCallStateHooks();

const { hasBrowserPermission: hasMicrophonePermission } = useMicrophoneState();
const { hasBrowserPermission: hasCameraPermission } = useCameraState();
```


packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
```diff
@@ -55,6 +55,20 @@ call.camera.state.selectedDevice; // currently selected camera
...
+In a browser, you can use the following API to check whether the user has granted permission to access the connected cameras:
+
+```typescript
+call.camera.state.hasBrowserPermission$.subscribe((value) => {
+  if (value) {
+    // User has granted permission
+  } else {
+    // User has denied or not yet granted permission
+  }
+});
```
szuperaz, line 69: For consistency, can we add a getter for the static value as well? Like this:

Screenshot 2023-11-10 at 09 04 10

packages/client/src/devices/InputMediaDeviceManagerState.ts
```diff
@@ -43,10 +44,47 @@ export abstract class InputMediaDeviceManagerState<C = MediaTrackConstraints> {
...
+        // permission doesn't exist or can't be queried -> assume it's granted
+        // an example would be Firefox,
+        // where neither camera microphone permission can be queried
+        notifyGranted();
+      });
+
+    return () => {
+      permissionState?.removeEventListener('change', notify);
+    };
+  });
```
szuperaz, line 74: I think we can use the `shareReplay(1)` pipe, and then the source method (the parameter of the constructor), will only run once, regardless of the number of subscribers

packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
oliverlaz: I explicitly didn't add one, as that would mean we'd need to run the permission querying during the SDK load time.
We also can't query in the getter, as they are supposed to run synchronously -> so, opted out of adding a getter :)

Do you maybe know a better way?

packages/client/src/devices/InputMediaDeviceManagerState.ts
oliverlaz: Good point, I'll update the implementation accordingly.

packages/client/docusaurus/docs/javascript/02-guides/04-camera-and-microphone.mdx
szuperaz: ah, I see, it's like the `listDevices` -> we only want to query if the user cares about this information. Yeah, not not having a getter makes sense.

oliverlaz: Yeah, it works the same as `listDevices`.

packages/client/src/devices/InputMediaDeviceManagerState.ts
oliverlaz: Going with `shareReplay({ bufferSize: 1, refCount: true })` as otherwise, the "observable cleanup" never runs once others unsubscribe.

szuperaz (approved):

PR #1191, opened 2023-11-15
chore(react-native): get apikey from endpoint

santhoshvai:


khushal87 (commented):
If you try to join the call as a user or an anonymous user in the meeting flow you will get:

```
Error: Call connectUser or connectAnonymousUser before creating a channel
```

khushal87 (approved):

PR #1202, opened 2023-11-28
fix(client): receive audioLevels when joined the call with mic muted

khushal87:
Client Changes:
- The `muteStream` API is changed now to `muteOrDestroyStream`.
- The `unmuteStream` API is changed to `unmuteOrCreateStream`.
- We have introduced a `disableTrackWhilePublish` argument that will not enable the track while publish, which currently happens every time [here](https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/rtc/Publisher.ts#L258-L260).

### DIFF
--- a/packages/client/src/devices/InputMediaDeviceManager.ts
+++ b/packages/client/src/devices/InputMediaDeviceManager.ts
@@ -8,6 +8,13 @@ import { getLogger } from '../logger';
 import { TrackType } from '../gen/video/sfu/models/models';
 import { deviceIds$ } from './devices';
 
+export type UnmuteOrCreateStreamSettings = {
+  /**
+   * Create a stream that is muted. Useful for React Native SDK audioLevel Stats.
+   */
+  createMutedStream?: boolean;
+};
+
 export abstract class InputMediaDeviceManager<
   T extends InputMediaDeviceManagerState<C>,
   C = MediaTrackConstraints,
@@ -55,7 +62,7 @@ export abstract class InputMediaDeviceManager<
    */
   async enable() {
     if (this.state.status === 'enabled') return;
-    this.enablePromise = this.unmuteStream();
+    this.enablePromise = this.unmuteOrCreateStream();
     try {
       await this.enablePromise;
       this.state.setStatus('enabled');
@@ -73,7 +80,7 @@ export abstract class InputMediaDeviceManager<
     this.state.prevStatus = this.state.status;
     if (this.state.status === 'disabled') return;
     const stopTracks = this.state.disableMode === 'stop-tracks';
-    this.disablePromise = this.muteStream(stopTracks);
+    this.disablePromise = this.muteOrDestroyStream(stopTracks);
     try {
       await this.disablePromise;
       this.state.setStatus('disabled');
@@ -140,8 +147,8 @@ export abstract class InputMediaDeviceManager<
 
   protected async applySettingsToStream() {
     if (this.state.status === 'enabled') {
-      await this.muteStream();
-      await this.unmuteStream();
+      await this.muteOrDestroyStream();
+      await this.unmuteOrCreateStream();
     }
   }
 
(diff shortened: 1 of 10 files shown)


santhoshvai:
if we have
`If there is no stream we create one and publish it first.`
do we still need?
`// In case of React Native we need to publish the stream everytime to get the audioLevels`

santhoshvai (approved):

packages/client/src/devices/MicrophoneManager.ts
```diff
@@ -61,6 +61,16 @@ export class MicrophoneManager extends InputMediaDeviceManager<MicrophoneManager
   private async startSpeakingWhileMutedDetection(deviceId?: string) {
     await this.stopSpeakingWhileMutedDetection();
     if (isReactNative()) {
+      // If there is no stream we create one and publish it first.
+      if (!this.call.publisher?.isPublishing(this.trackType)) {
+        const stream = await this.getStream({ deviceId });
```
szuperaz, line 66: I think you should add a flag to `unmuteStream` to create a muted stream (the naming will be weird, that the `unmuteStream` function creates a muted stream, but I think we can live with that, or at least I don't have a better name)

And then call `unmuteStream` here, the reason:
- the `unmuteStream` also applies default constraints: https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/devices/InputMediaDeviceManager.ts#L220
- register necessary event handlers, in case the stream end unexpectedly: https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/devices/InputMediaDeviceManager.ts#L233

We will miss all those functionality if you create the stream like this

szuperaz: maybe we can rename `unmuteStream` to `createOrUnmuteStream` and similarly `muteStream` could be `muteOrDestroyStream`

szuperaz: One more thing: `this.call.publisher?.isPublishing(this.trackType)` -> only check if we're publishing, you should also check if `this.state.mediaStream === undefined`, and only create the stream if that's true as well

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -207,25 +196,60 @@ export abstract class InputMediaDeviceManager<
...
-    } else {
+    if (!this.state.mediaStream && settings && settings.createMutedStream) {
       const defaultConstraints = this.state.defaultConstraints;
       const constraints: MediaTrackConstraints = {
         ...defaultConstraints,
         deviceId: this.state.selectedDevice,
       };
       stream = await this.getStream(constraints as C);
+      this.state.setMediaStream(stream);
+      await this.muteOrDestroyStream(this.state.disableMode === 'stop-tracks');
```
szuperaz, line 233:
```suggestion
      await this.muteOrDestroyStream(false);
```

We always want to just mute the stream here, if we call it with `stop-tracks` we destroy the stream that we've just created

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -207,25 +196,60 @@ export abstract class InputMediaDeviceManager<
...
     this.logger('debug', 'Starting stream');
     let stream: MediaStream;
-    if (
-      this.state.mediaStream &&
-      this.getTracks().every((t) => t.readyState === 'live')
-    ) {
-      stream = this.state.mediaStream;
-      this.unmuteTracks();
-    } else {
+    if (!this.state.mediaStream && settings && settings.createMutedStream) {
```
szuperaz, line 225: I think we can do this with less copy-paste, and we can remove this whole if statement, see below

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -207,25 +196,60 @@ export abstract class InputMediaDeviceManager<
...
+      ) {
+        stream = this.state.mediaStream;
+        this.unmuteTracks();
+      } else {
+        const defaultConstraints = this.state.defaultConstraints;
+        const constraints: MediaTrackConstraints = {
+          ...defaultConstraints,
+          deviceId: this.state.selectedDevice,
+        };
+        stream = await this.getStream(constraints as C);
```
szuperaz, line 247: Check here if we need to mute

```
if (settings && settings.createMutedStream) {
  await this.muteOrDestroyStream(false);
}
```

PR #1204, opened 2023-11-29
chore(ci): Automatic update BE QA client on new release

guerinoni:
This is triggered on new tag and will checkout BE + update client library opening a PR in that repo. It is useful because right now we have to do it manually and it is easier catch bugs and incosistency.


.github/workflows/update-be.yml
```diff
@@ -0,0 +1,36 @@
+name: Update BE QA version
+on:
+  push:
+    tags:
+    - v[0-9]+.[0-9]+.[0-9]+
```
oliverlaz, line 5:
```suggestion
    - @stream-io/video-client@v[0-9]+.[0-9]+.[0-9]+
```

I think we should be more specific here, as we don't want to open PRs for non-client related releases, right?

Issue #1215, opened 2023-12-06
The owner of a call do not automatically cancel the call when Auto-Cancel Timeout reached


Phu96:
**Which package/packages do you use?**

- [x ] `@stream-io/video-react-sdk`
- [ ] `@stream-io/video-react-native-sdk`
- [ ] `@stream-io/video-client`

**Describe the bug**
Currently, we're having a problem that when the user makes a ringing call, the logic that automatically cancels the call should be triggered when the Auto-Cancel Timeout config is reached, but in reality it's not.

**To Reproduce**

**Expected behavior**
The logic of automatically cancelling the call when Auto-Cancel Timeout reached should be triggered.

**Screenshots**
![Screenshot from 2023-12-06 17-17-28](https://github.com/GetStream/stream-video-js/assets/30786617/bfdd8fe4-0cb6-426c-8e3e-eb6f182228d2)
I've digged into the source code and saw that when prevTimeoutMs and timeoutMs are the same, it returns and does not run the setTimeout function that triggers the cancel flow. In reality, both prevTimeoutMs and timeoutMs are always the same when the user makes a call, so I am doubting that this is a bug.

**Version**

- SDK version:  @stream-io/video-react-sdk: 0.4.14

**Env:**

Fill out what's applicable

**React / JS client**

- Node version: v18.16.0

**Additional context**
this is the web application built using React JS

oliverlaz:
hi @Phu96, thanks for reporting this issue. We'll take a look at it and get back to you.

PR #1217, opened 2023-12-08
fix(ringing): Auto-Cancel outgoing calls

oliverlaz:
### Overview

This PR fixes a bug in the `scheduleAutoDrop()` function that bailed out from scheduling a cancellation timer when server-side defined timeouts remained the same after consecutive call settings updates.

Also, this PR aligns the implementation and relies only on `auto_cancel_timeout_ms`. More context:
- https://getstream.slack.com/archives/C022N8JNQGZ/p1697701678593769

Fixes: #1215

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -82,10 +82,8 @@ import {
   debounce,
   filter,
   map,
-  pairwise,
   Subject,
   takeWhile,
-  tap,
   timer,
 } from 'rxjs';
 import { TrackSubscriptionDetails } from './gen/video/sfu/signal_rpc/signal';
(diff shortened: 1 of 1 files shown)


szuperaz (approved):
@oliverlaz does this mean that `incoming_call_timeout_ms` will be removed from BE? I couldn't find that info in the linked Slack conversation

oliverlaz:
> @oliverlaz does this mean that `incoming_call_timeout_ms` will be removed from BE? I couldn't find that info in the linked Slack conversation

I'm not sure when this is going to happen, as this is a breaking change. But, if we keep using the two values, we may run into a weird situation.

PR #1223, opened 2023-12-25
fix(devices): prevent mute/unmute when switching a device

oliverlaz:
### Overview

This PR changes the "muted" SFU notification flow when switching a device.

### Implementation notes
Before this PR, when switching a device the client used to do the following:
1. Stop the existing media stream, send "user muted" notification to the SFU
2. Create a new media stream, publish it to the SFU, and send a "user unmuted" notification.

Now, the flow is changed to this:
1. Stop the existing media stream
2. Create a new media stream and publish it to the SFU

We aren't sending any "user muted/unmuted" notifications when switching a mic or camera anymore.

### Breaking changes
- `call.stopPublish` signature is changed to `stopPublish(trackType, options)`. Although our integrators aren't supposed to use this method directly in their apps, the migration is quite trivial:
```typescript

// old code
await call.stopPublish(TrackType.VIDEO, true);

// new code
await call.stopPublish(TrackType.VIDEO, { stopTracks: true });
```


szuperaz (approved):

oliverlaz:
Closing stale PR.

PR #1224, opened 2023-12-26
perf: speed up the SFU WS message decoding

oliverlaz:
### Overview

Includes the newly generated protobuf-ts models based on https://github.com/GetStream/protocol/pull/347
The newest version of `protobuf-ts` includes an important performance optimization that we can benefit from in our SFU WS message decoding flow.

More about it: https://github.com/timostamm/protobuf-ts/pull/582


khushal87 (approved):
Approved, assuming its all generated and no manual changes. ✅

oliverlaz:
> Approved, assuming its all generated and there are no manual changes. ✅

Yep. All we need to do is to regenerate the models and use the latest `@protobuf-ts/runtime` version.

PR #1226, opened 2024-01-08
docs(react-native): add sending and receiving custom events advanced guide

khushal87:


khushal87:
Merging after self review.

PR #1228, opened 2024-01-08
feat(react-sdk): update CompositeButton layout

myandrienko:
🚂 #1194 

Previously, CompositeButton relied on its `children` to be a button themselves. That led to a mismatch between a hover target (the whole CompositeButton) and a click target (e.g. an IconButton passed as `children`).

With this update, CompositeButton renders a `button` element itself. So, you should no longer pass IconButton or TextButton as `children` - just pass an `Icon` or text.

This is a breaking change, since props for CompositeButton have changed.

As a bonus, most of the styling for react-dogfood's DualButton was removed, since it just re-implemented CompositeButton with a slightly different styling.

### DIFF
--- a/packages/react-sdk/src/components/Button/CompositeButton.tsx
+++ b/packages/react-sdk/src/components/Button/CompositeButton.tsx
@@ -1,6 +1,12 @@
 import clsx from 'clsx';
 import { MenuToggle, ToggleMenuButtonProps } from '../Menu';
-import { ComponentType, forwardRef, JSX, PropsWithChildren } from 'react';
+import {
+  ComponentProps,
+  ComponentType,
+  forwardRef,
+  JSX,
+  PropsWithChildren,
+} from 'react';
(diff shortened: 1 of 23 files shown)


packages/react-sdk/src/components/Button/CompositeButton.tsx
```diff
@@ -52,7 +61,16 @@ export const CompositeButton = forwardRef<
             variant === 'secondary' && active,
         })}
       >
-        {children}
+        <button
+          className={clsx('str-video__composite-button__button')}
```
oliverlaz, line 65:
```suggestion
          className={clsx('str-video__composite-button__button')}
          type="button"
```
I think it is a good practice to add `type="button"` to all `<button>` elements as they otherwise may mess-up possible forms.

myandrienko: Good catch!

PR #1230, opened 2024-01-09
feat(react-dogfood): fix push-to-talk behavior

myandrienko:
This PR fixes the following issues with push-to-talk:

1. Pressing space for push-to-talk conflicted with default browser behavior when there was a focused element on the page (both the focused element and push-to-talk were activated).
2. Pressing space disabled microphone when it was previously manually enabled.
3. Microphone was not always disabled at the end of push-to-talk if you managed to press and let go of the spacebar quickly enough.

### DIFF
--- a/sample-apps/react/react-dogfood/hooks/useKeyboardShortcuts.ts
+++ b/sample-apps/react/react-dogfood/hooks/useKeyboardShortcuts.ts
@@ -34,6 +34,8 @@ export const usePushToTalk = (key: string) => {
       return;
     }
 
+    e.preventDefault();
+
     if (e.type === 'keydown' && isMute) {
       enableMicWithPushToTalkPromiseRef.current = microphone
         .enable()
--- a/sample-apps/react/react-dogfood/hooks/useKeyboardShortcuts.ts
+++ b/sample-apps/react/react-dogfood/hooks/useKeyboardShortcuts.ts
@@ -24,36 +24,39 @@ const isMacOS = () => {
 const [, raiseHandReaction] = defaultReactions;
 
 export const usePushToTalk = (key: string) => {
-  const [isTalking, setIsTalking] = useState(false);
-  const interactedRef = useRef(false);
-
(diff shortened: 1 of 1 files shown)


sample-apps/react/react-dogfood/hooks/useKeyboardShortcuts.ts
```diff
@@ -24,36 +24,41 @@ const isMacOS = () => {
 const [, raiseHandReaction] = defaultReactions;
 
 export const usePushToTalk = (key: string) => {
-  const [isTalking, setIsTalking] = useState(false);
-  const interactedRef = useRef(false);
-
   const { useMicrophoneState } = useCallStateHooks();
-  const { microphone } = useMicrophoneState();
+  const { microphone, isMute } = useMicrophoneState();
+  const hotkeyHandlerRef = useRef<(e: KeyboardEvent) => void>();
```
myandrienko, line 29: Putting the event handler in a ref allows it it have access to the current microphone state, without the need to have microphone state as an effect dependency. This is kind of like what the [`useEvent` hook](https://github.com/reactjs/rfcs/blob/useevent/text/0000-useevent.md) does.

sample-apps/react/react-dogfood/hooks/useKeyboardShortcuts.ts
```diff
@@ -24,36 +24,41 @@ const isMacOS = () => {
...
+    e.preventDefault();
+
+    if (e.type === 'keydown' && isMute) {
+      enableMicWithPushToTalkPromiseRef.current = microphone
+        .enable()
+        .catch(console.error);
+    }
+
+    if (e.type === 'keyup' && enableMicWithPushToTalkPromiseRef.current) {
+      enableMicWithPushToTalkPromiseRef.current
```
myandrienko, line 46: Waiting for mic to enable before disabling it to prevent race condition.

oliverlaz:
hey @myandrienko, can you change the base of this PR to #1194? The rest looks good to me 👍

PR #1237, opened 2024-01-12
chore(react-native): add play store release config

khushal87:


sample-apps/react-native/dogfood/android/app/build.gradle
```diff
@@ -78,6 +78,23 @@ def reactNativeArchitectures() {
...
+       standardOutput = stdout
+       errorOutput = stderr
+       ignoreExitValue true
+   }
+   //noinspection GroovyAssignabilityCheck
+      stdout.toString().trim()
+}
+
+// Call the `getPassword` function to get the release password
+def releasePassword = getPassword("Khushal Agarwal", "stream-video-calls-rn.keystore")
```
santhoshvai, line 96: i hope there are better ways than this..

We could store the file in the CI and then use that in firebase - I think

santhoshvai: and this seems  to be mac os sepcific, while we use ubuntu in CI

santhoshvai (changes requested):
we can also alternatively pick this up later while doing CI

PR #1246, opened 2024-01-18
feat(events)!: improved type narrowing on call events

oliverlaz:
### Overview

Improves the type narrowing on `call` events. Event listeners used to be quite general and thus, required them to do type narrowing inside their body. With this change that is no longer necessary as now TS can narrow the type for us.
```ts
// Call events (before):
call.on('call.created', (e) => {
  if (e.type !== 'call.created') return;
  // rest of the body
});
// Call events (after):
call.on('call.created', (e) => {
  // rest of the body, the type of 'e' is correctly narrowed
});

// SFU events (before):
call.on('trackPublished', (e) => {
  if (e.eventPayload.oneofKind !== 'trackPublished') return;
  const data = e.eventPayload.trackPublished;
  const trackType = data.type;
  // rest of the body
});
// SFU events (after):
call.on('trackPublished', (e) => {
  const trackType = e.type;
  // rest of the body
});
```

### Breaking Changes
- Some type definitions are removed (`EventHandler`, `CallEventHandler`)
- SFU events are now dispatching unwrapped data - these are considered internal, and customers should be using `call.state`

### Other important notes
- Reduced the number of functions that are invoked when dispatching an event on the coordinator client
- Coordinator web socket events parsing isn't duplicated anymore - we should have some perf gains here.


szuperaz (approved):

packages/client/src/rtc/Dispatcher.ts
```diff
@@ -1,8 +1,51 @@
...
+  joinResponse: JoinResponse;
+  healthCheckResponse: HealthCheckResponse;
+  trackPublished: TrackPublished;
+  trackUnpublished: TrackUnpublished;
+  error: SfuError;
+  callGrantsUpdated: CallGrantsUpdated;
+  goAway: GoAway;
+  iceRestart: ICERestart;
+  pinsUpdated: PinsChanged;
+};
```
myandrienko, line 48: TS just needs a little more assurance that K can be used to index the extracted type :)

```suggestion
export type AllSfuEvents = {
  [K in SfuEventKinds]: K extends keyof Extract<
    SfuEvent['eventPayload'],
    { oneofKind: K }
  >
    ? Extract<SfuEvent['eventPayload'], { oneofKind: K }>[K]
    : never;
};
```

packages/client/src/rtc/Dispatcher.ts
```diff
@@ -31,41 +74,43 @@ export const isSfuEvent = (
...
 
   dispatch = (message: SfuEvent) => {
     const eventKind = message.eventPayload.oneofKind;
     if (eventKind) {
-      this.logger(
-        'debug',
-        `Dispatching ${eventKind}`,
-        (message.eventPayload as any)[eventKind],
-      );
+      const payload = (message.eventPayload as any)[eventKind];
```
myandrienko, line 86: To prevent explicit casting here, you can use a helper type:

```ts
type DispatchableMessage<K extends SfuEventKinds> = {
  eventPayload: {
    oneofKind: K;
  } & {
    [Key in K]: AllSfuEvents[Key];
  };
};
```

And use it like this:

```ts
dispatch = <K extends SfuEventKinds>(message: DispatchableMessage<K>) => {
  /* ... */
}
```

packages/client/src/rtc/Dispatcher.ts
oliverlaz: Nice! thank you for the optimization!

packages/client/src/rtc/Dispatcher.ts
oliverlaz: hey @myandrienko, can you check this one? https://github.com/GetStream/stream-video-js/pull/1246/commits/37caaa477cba7392906d53d467ddf8d65c25e58e

PR #1247, opened 2024-01-19
fix(client): automatic call join for other participants when someone accepts

### LINKED ISSUE #1245: Other members automatically join a call when the ringing call has more than two members

- [x ] `@stream-io/video-react-sdk`

**Describe the bug**
When creating a ringing call with more than two members, for instance, the call has four members that are A,B,C,D respectively. The user A is the owner of the call, when the others receive a call notification, and one of them accepts the call, let say the user B accepts the call, then the user C and D will automatically join the call, that is not expected. My expectation is that when the user B accepts the call, only the owner of the call which is the user A will join that call, the others will only join if any of them accepts the call manually.

**To Reproduce**

**Expected behavior**
Only the first user accept the call and the owner will join the call, the others have to accept the call manually to join.

**Screenshots**

**Version**

- SDK version: 0.4.17

**Env:**

Fill out what's applicable

**React / JS client**

**Additional context**

khushal87:
Fixes #1245


oliverlaz:
hey @khushal87, since you are touching this, can you also add a test that verifies that other callees don't automatically join?

oliverlaz (approved):
Looks good to me now :) Thank you @khushal87!

PR #1248, opened 2024-01-23
fix(permissions): relax device permission handling for Safari

oliverlaz:
### Overview

On Safari, `PermissionStatus#change` event doesn't emit reliably. As a consequence, most of the time our SDK thinks that Mic/Camera permissions aren't granted and shows a warning icon next to the Mic and Camera button.

With this PR, the permission notification system is relaxed a bit for Safari.


szuperaz (approved):

szuperaz:
Just out of curiosity, do you have any more context about this issue in Safari @oliverlaz?

oliverlaz:
> Just out of curiosity, do you have any more context about this issue in Safari @oliverlaz?

I only managed to find this report: https://forums.developer.apple.com/forums/thread/742439
Unfortunately, no answers or solutions but it seems other people were experiencing this. Until we figure out what's wrong with this, I think it makes sense to keep this validation relaxed.

khushal87 (approved):

PR #1252, opened 2024-01-29
fix(react-sdk): consider call setting permissions in CallControls

oliverlaz:
### Overview

Besides browser permissions and own capabilities, we now include call settings to determine whether one is allowed to publish audio, video, or screen share.


myandrienko (approved):

oliverlaz:
Cherry-picked in: https://github.com/GetStream/stream-video-js/commit/c2ff1f98c005ce6165743082882da6d62835ad99

PR #1256, opened 2024-01-30
fix(egress-composite): titleText renders 0 if it's an empty string

arnautov-anton:


sample-apps/react/egress-composite/src/components/LogoAndTitleOverlay.tsx
```diff
@@ -12,7 +12,7 @@ export const LogoAndTitleOverlay = () => {
       data-testid="logo-and-title-overlay"
       className="eca__logo-and-title-overlay"
     >
-      {titleText?.length && (
+      {!!titleText?.length && (
```
oliverlaz, line 15:
```suggestion
      {titleText && (
```

Why do we need to check the `.lenght` at all?

oliverlaz (approved):

sample-apps/react/egress-composite/src/components/LogoAndTitleOverlay.tsx
arnautov-anton: Good point, adjusted.

PR #1266, opened 2024-02-12
feat(client): add stopOnLeave param to device manager

### LINKED ISSUE #1236: Video/audio tracks not properly closed when the stream is unmounted on iOS.

- [X] `@stream-io/video-react-native-sdk`

**Describe the bug**
After a stream is ended/unmounted, the video/audio tracks are still active. This can be observed by seeing that the indicator for camera/microphone on the phone is active (yellow/green indicator light). It seems like for a split second the tracks are really stopped since the indicator turns off, but after a second it seems like they become active again.

**To Reproduce**

1. Open Sample app/Dogfooding app (but this happens in my own app as well)
2. Start a livestream
3. Leave the livestream
4. Observe that the camera/microphone active indicators are still present, even after leaving the livestream.

**Expected behavior**
After leaving the livestream, the camera/microphone active indicators should immediately stop.

**Screenshots**
N/A

**Version**

- `@stream-io/video-react-native-sdk 0.3.7`

**React Native**

- OS: iOS
- Device: real iPhone 14 Pro

**Additional context**
This issue does NOT appear on Android. It seems to only affect iOS.

santhoshvai:
## Goal

Adds a parameter to device manager to enable automatically stopping streams when call is left

fixes #1236

### DIFF
--- a/packages/client/src/devices/InputMediaDeviceManager.ts
+++ b/packages/client/src/devices/InputMediaDeviceManager.ts
@@ -23,7 +23,7 @@ export abstract class InputMediaDeviceManager<
   /**
    * if true, stops the media stream when call is left
    */
-  stopOnLeave = false;
+  stopOnLeave = true;
   logger: Logger;
   private subscriptions: Subscription[] = [];
   private isTrackStoppedDueToTrackEnd = false;
--- a/packages/client/src/devices/InputMediaDeviceManager.ts
+++ b/packages/client/src/devices/InputMediaDeviceManager.ts
@@ -20,6 +20,10 @@ export abstract class InputMediaDeviceManager<
    * @internal
    */
   disablePromise?: Promise<void>;
+  /**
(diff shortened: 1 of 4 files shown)


packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -20,6 +20,10 @@ export abstract class InputMediaDeviceManager<
    * @internal
    */
   disablePromise?: Promise<void>;
+  /**
+   * if true, stops the media stream when call is left
+   */
+  stopOnLeave = false;
```
oliverlaz, line 26: hey @santhoshvai, Wouldn't it be better if we set this to `true` by default?

santhoshvai: well, yes, sorry was testing them and missed to change it

oliverlaz (approved):

PR #1272, opened 2024-02-27
chore: remove the legacy video demo app

oliverlaz:
### Overview

As part of #1194, this application was merged with the one we use for internal testing (react-dogfood). As we don't want to maintain this codebase anymore, this PR removes the app's code.
We know some customers might have looked into it and that is why the `react-video-demo` directory stays in the React sample apps. The Readme is updated with a hint to look at the React Dogfood app instead.


myandrienko (approved):
image

PR #1276, opened 2024-02-29
feat: SFU stats reporter (experimental)

oliverlaz:
### Overview

Adds experimental support for WebRTC stats reporting towards the SFU.


packages/client/src/client-details.ts
```diff
@@ -30,13 +35,26 @@ export const getDeviceInfo = () => {
...
+
+export const setWebRTCInfo = (info: WebRTCInfoType) => {
+  webRtcInfo = info;
+};
+
+export type LocalClientDetailsType = ClientDetails & {
+  webRTCInfo?: WebRTCInfoType;
+};
+
+export const getClientDetails = (): LocalClientDetailsType => {
```
oliverlaz, line 50: I wouldn't change the return type of this method as it is used in `Call.join()`.

I'm perfectly fine with using `getWebRtcVersion()` in `SfuStatsReporter`

packages/react-native-sdk/src/utils/setClientDetails.ts
```diff
@@ -3,9 +3,11 @@ import {
   setDeviceInfo,
   setOSInfo,
   SfuModels,
+  setWebRTCInfo,
 } from '@stream-io/video-client';
 import { Platform } from 'react-native';
 import { version } from '../version';
+import RNWebRTCPackageJSON from '@stream-io/react-native-webrtc/package.json';
```
oliverlaz, line 10: Does this work? I think the SDK is doing self-importing. Perhaps change to `import package from '../../package.json'?
Or have this value injected as part of the build process?
- https://github.com/GetStream/stream-video-js/blob/main/packages/react-sdk/rollup.config.mjs#L32-L35
- https://github.com/GetStream/stream-video-js/blob/main/packages/react-sdk/index.ts#L17-L26

oliverlaz:
Replaced with #1297.

PR #1283, opened 2024-03-08
feat(react): Support for Background Filters and Background Blurring

oliverlaz:
### Overview

Adds API for registering video filters, default plugin based on TensorFlow Lite, and a sample integration in our Demo app.

**New APIs:**
- `camera.registerFilter(async (mediaStream) => mediaStream)` the core part. Allows you to chain multiple video and audio filters.
- `@stream-io/video-filters` - a new lib that is responsible for the ML part and WebGL rendering
- `BackgroundFiltersProvider` - a component in the React SDK that helps you to manage the applied filters.

Our dogfooding app integrates these new APIs and provides the necessary input (background images) and styling.

### Todos:
- [ ] Deploy internally and test with a wider audience
- [ ] Document the `registerFilter` API
- [ ] Cookbook for integrating this feature in a customer application for React


packages/react-sdk/src/components/BackgroundFilters/BackgroundFilters.tsx
```diff
@@ -0,0 +1,263 @@
...
+      {children}
+      <BackgroundFilters />
+    </BackgroundFiltersContext.Provider>
+  );
+};
+
+const BackgroundFilters = () => {
+  const call = useCall();
+  const { backgroundImage, backgroundFilter } = useBackgroundFilters() || {};
+  const [videoRef, setVideoRef] = useState<HTMLVideoElement>();
```
arnautov-anton, line 118:
```suggestion
  const [videoRef, setVideoRef] = useState<ElementRef<'video'> | null>(null);
```
I think all of these should be initiated with `null` as that's what's being set when the element de-renders. That way you can get rid of all the @ts-expect-error directives.

santhoshvai:
can we name it `@stream-io/video-filters-web` instead to avoid confusion on other platforms

oliverlaz:
> can we name it `@stream-io/video-filters-web` instead to avoid confusion on other platforms

Sure, good point! I'll rename it 👍

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -113,6 +117,29 @@ export abstract class InputMediaDeviceManager<
...
+  async registerFilter(filter: MediaStreamFilter) {
+    this.filters.push(filter);
+    if (this.state.status === 'enabled') {
+      await this.disable();
+      await this.enable();
+    }
+    return async () => {
+      this.filters = this.filters.filter((f) => f !== filter);
+      if (this.state.status === 'enabled') {
+        await this.disable();
```
szuperaz, line 137: We have the `applySettingsToStream` method to avoid having to do a status update (`enabled` -> `disabled` -> `enabled`) on settings change, can't we use that method here as well?

oliverlaz: Good point. I'll adjust accordingly!

oliverlaz: Adjusted in [cef2749](https://github.com/GetStream/stream-video-js/pull/1283/commits/cef27495475061e20fc939bf010b6dbfe5cfdf0b)

szuperaz (approved):

packages/react-sdk/docusaurus/docs/React/10-advanced/06-apply-video-filters.mdx
```diff
@@ -6,30 +6,21 @@ description: How to build video or audio filters
 ## Video Filters
 
 In this guide, we will show you how to apply custom video filters to a video stream.
-We are going to use the `@stream-io/video-filters-web` library that provides background blur and background replacement filters.
+We are going to use the [Stream React Video SDK](https://getstream.io/video/sdk/react/) provided components that enable background blur and background replacement filters.
```
myandrienko, line 9: The link here seems a bit excessive, since it's the docs for this package :)

packages/react-sdk/docusaurus/docs/React/10-advanced/06-apply-video-filters.mdx
```diff
@@ -6,30 +6,21 @@ description: How to build video or audio filters
...
-```bash
-npm install @stream-io/video-filters-web
-# or
-yarn add @stream-io/video-filters-web
-```
+These components use the [`camera.registerFilter()`](../02-guides/04-camera-and-microphone.mdx#camera-and-microphone-filters) API under the hood to register custom filters.
 
 :::note
-This library supports only modern desktop browsers.
+We currently support video filters only on modern desktop browsers.
```
myandrienko, line 14: It would be great to mention which exact features are expected from the browser, maybe even link caniuse?

myandrienko (approved):

packages/react-sdk/docusaurus/docs/React/10-advanced/06-apply-video-filters.mdx
oliverlaz: I wasn't sure if we should go too technical here which is why I excluded this info. Essentially it would be this: https://github.com/GetStream/stream-video-js/blob/79bb660a9d615427e8d940951692c6031b5158cb/packages/video-filters-web/src/compatibility.ts

PR #1296, opened 2024-03-22
feat(call): Add getCallStats method

su225:


packages/client/src/gen/coordinator/index.ts
```diff
@@ -4407,6 +4610,31 @@ export interface UpdatedCallPermissionsEvent {
...
+   * @memberof UserInfoResponse
+   */
+  Image: string;
+  /**
+   *
+   * @type {string}
+   * @memberof UserInfoResponse
+   */
+  Name: string;
+}
```
oliverlaz, line 4637: @su225, I'll update these once we get an updated schema (lowercaps props).

su225: This is fixed: https://github.com/GetStream/protocol/blob/d3f2446403f533bd624aad41209436964e555586/openapi/video-openapi.yaml#L2482-L2492

su225:
Restoring the branch to make CI in BE happy :D

PR #1302, opened 2024-03-29
feat(react-native): add screen share and chat to livestream sample

santhoshvai:


packages/react-native-sdk/src/components/Participant/ParticipantView/VideoRenderer.tsx
santhoshvai, line 1: fixes in this file

* do not subscribe or inform visibility states for local-participant
* add missing visibility states for screen share

khushal87:
On the host side, the video of the host is not visible at all if he is sharing screen, not even in a participant view.

On the viewer side, as you have already put in, we need a label that someone is sharing a screen, also the floating participant view is behind the top and bottom view, although this is not something that you have introduced in your PR.

khushal87:
When the screen sharing is started, the button of screen sharing should change the icon to stop and the color should also be red IMO. Currently there is no user feedback on if the screen has been shared by a host.

packages/react-native-sdk/src/components/Call/CallControls/ScreenShareButton.tsx
```diff
@@ -43,78 +33,20 @@ export const ScreenShareButton = ({
...
-        onScreenShareStoppedHandlerRef.current?.();
-        await call?.stopPublish(SfuModels.TrackType.SCREEN_SHARE);
-      }
-    }
-  };
+  const {
+    isScreenSharingEnabledInCall,
+    isScreenSharingAccessRequestEnabled,
+    CanDeviceScreenShare,
+    onPress,
```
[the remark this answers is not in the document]
santhoshvai: I think I better rename the button to ScreenShareToggleButton

packages/react-native-sdk/src/components/Livestream/HostLivestream/HostLivestream.tsx
```diff
@@ -72,7 +71,7 @@ export const HostLivestream = ({
   };
 
   return (
-    <SafeAreaView
```
santhoshvai, line 75: @khushal87 we should never use safeareaview inside the SDK

we have to let the user do the safe area stuff.. as these components can be inside bottom tabs.. half height of screens etc

packages/react-native-sdk/src/utils/StreamVideoRN/index.ts
santhoshvai, line 1: unrelevant change to this PR

this enables awaiting in the logout process (optionally)

PR #1310, opened 2024-04-08
feat: user feedback

su225:


packages/client/src/gen/coordinator/index.ts
```diff
@@ -2008,6 +2008,64 @@ export interface ClosedCaptionEvent {
...
+ * @export
+ * @interface CollectUserFeedbackRequest
+ */
+export interface CollectUserFeedbackRequest {
+  /**
+   *
+   * @type {object}
+   * @memberof CollectUserFeedbackRequest
+   */
+  custom?: object;
```
oliverlaz, line 2023:
```suggestion
  custom?: { [key:string]: any };
```

oliverlaz: This is a general issue that still hasn't been resolved in the OpenAPI schema. Can you please change the `object` as suggested?

oliverlaz (approved):

PR #1314, opened 2024-04-11
fix(react-native): change objectFit to contain for screensharing ParticipantView

khushal87:


packages/react-native-sdk/src/components/Call/CallLayout/CallParticipantsSpotlight.tsx
```diff
@@ -113,6 +113,7 @@ export const CallParticipantsSpotlight = ({
                   callParticipantsSpotlight.spotlightContainer,
                 ]
           }
+          objectFit="contain"
```
santhoshvai, line 116: dont we need to set as contain only for screenshare tracks? here it is also set for video tracks

PR #1330, opened 2024-04-29
fix(state): optimized Call State updates

oliverlaz:
### Overview

**Observables emitting primitive values will emit only when the value changes**
`false` -> `false` - the subscribers won't run for the second update as the value didn't change

**updateOrAddParticipant doesn't iterate the participant array twice before updating or adding a participant**
We were iterating the participant array to figure out whether to add a participant or update an existing one. Then, we create a shallow clone of the array and apply the updates. This isn't optimal.
Now, these two operations are merged into one operation.

### DIFF
--- a/packages/client/src/store/CallState.ts
+++ b/packages/client/src/store/CallState.ts
@@ -398,9 +398,8 @@ export class CallState {
      */
     const isShallowEqual = <T>(a: Array<T>, b: Array<T>): boolean => {
       if (a.length !== b.length) return false;
-      for (let i = 0; i < a.length; i++) {
-        if (b.indexOf(a[i]) === -1) return false;
-      }
+      for (const item of a) if (!b.includes(item)) return false;
+      for (const item of b) if (!a.includes(item)) return false;
       return true;
     };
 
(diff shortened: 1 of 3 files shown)


packages/client/src/store/CallState.ts
```diff
@@ -376,33 +372,60 @@ export class CallState {
...
     this.thumbnails$ = this.thumbnailsSubject.asObservable();
 
+    /**
+     * Performs shallow comparison of two arrays.
+     * Expects primitive values: [1, 2, 3] is equal to [2, 1, 3].
+     */
+    const isShallowEqual = <T>(a: Array<T>, b: Array<T>): boolean => {
+      if (a.length !== b.length) return false;
+      for (let i = 0; i < a.length; i++) {
+        if (b.indexOf(a[i]) === -1) return false;
```
myandrienko, line 402: a = [1, 2, 2]
b = [1, 2, 3]

Will return true.

I suppose in our case all values are unique, but worth looking into it.

oliverlaz: Good point. Let me see how to fix this efficiently :)

oliverlaz: Fixed with: [affbc30](https://github.com/GetStream/stream-video-js/pull/1330/commits/affbc3003123e492cbde65c735ed6c390e22c10b)

PR #1335, opened 2024-05-02
fix(devices): API to disable speaking while muted notifications

### LINKED ISSUE #1329: Disable microphone while muted

**Describe the solution you'd like**
Currently when the microphone is muted via `micrphone.disable()` it just disables the track/stream. However, the microphone is still enabled on the device and is actively listening, I believe this is to support the `SpeakingWhileMutedDetection`. It would be nice if it was possible to truly disable the microphone while muted, this is particularly useful in a PTT type of application. Without this our users see the microphone privacy/notification constantly active on their device, even though they muted (not pushing PTT).

oliverlaz:
### Overview

A joint PR that fixes a couple of related issues:
- Adds an API that disables speaking while muted notifications - fixes #1329
- The selected device can't be reliably detected when audio or video filters are used
- Audio tracks are now stopped (instead of disabled) on the Web - this should release the mic allocation
- Fixed an issue in `usePersistedDevicePreferences()` where the camera and mic were automatically enabled after a call had been left
- Pronto and Video Demo now remember the last used video filter and they enable it automatically

### DIFF
--- a/packages/client/src/devices/MicrophoneManager.ts
+++ b/packages/client/src/devices/MicrophoneManager.ts
@@ -3,6 +3,7 @@ import type { INoiseCancellation } from '@stream-io/audio-filters-web';
 import { Call } from '../Call';
 import { InputMediaDeviceManager } from './InputMediaDeviceManager';
 import { MicrophoneManagerState } from './MicrophoneManagerState';
+import { TrackDisableMode } from './InputMediaDeviceManagerState';
 import { getAudioDevices, getAudioStream } from './devices';
 import { TrackType } from '../gen/video/sfu/models/models';
 import { createSoundDetector } from '../helpers/sound-detector';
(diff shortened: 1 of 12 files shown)


packages/client/src/devices/MicrophoneManager.ts
```diff
@@ -16,37 +17,48 @@ import { createSubscription } from '../store/rxUtils';
...
-        }
-      } else {
-        await this.stopSpeakingWhileMutedDetection();
-      }
-    });
+  constructor(
+    call: Call,
+    disableMode: TrackDisableMode = isReactNative()
+      ? 'disable-tracks'
+      : 'stop-tracks',
```
oliverlaz, line 31: @szuperaz, we should check if this makes sense once you are back. This defaulted to `disable-tracks` for the mic, and `stop-tracks` for the camera. I don't know the reasons behind this.

I tested with Web, and now the mic indicator goes away once both, mic and speaking-while-muted notifications are disabled. I left the previous default for React Native.

PR #1342, opened 2024-05-07
fix: optimistically toggle device status

myandrienko:
Clicking on camera, microphone and screen share toggles gives immediate visual feedback, even if actually enabling/disabling a device takes longer.

Marked as `fix` mostly for semver reasons.


packages/react-sdk/src/hooks/useOptimisticDeviceStatus.ts
```diff
@@ -0,0 +1,32 @@
...
+  const requestPromise = useRef<Promise<void> | null>(null);
+
+  useEffect(() => {
+    console.log({ optimisticIsMute, isMute });
+    if (optimisticIsMute !== isMute && requestPromise.current === null) {
+      requestPromise.current = manager.toggle().finally(() => {
+        requestPromise.current = null;
+      });
+    }
+  }, [manager, optimisticIsMute, isMute]);
```
myandrienko, line 21: "Eventual consistency" :)

Update `optimisticIsMute` first, then deal with doing actual work in an effect.

packages/react-sdk/src/hooks/useOptimisticDeviceStatus.ts
```diff
@@ -0,0 +1,32 @@
+import { useCallback, useEffect, useRef, useState } from 'react';
+
+interface Toggleable {
+  toggle: () => Promise<void>;
+}
```
myandrienko, line 5: Unfortunately no way to use `InputMediaDeviceManager` type here because of private members and nominal typing.

oliverlaz: What would happen if `toggle()` throws an error for any reason?

myandrienko: Good question. I've added handling for that case. In that case we reset the optimistic state: https://github.com/GetStream/stream-video-js/pull/1342/commits/e485ec304e616d963e09ec16515bb8c3bec0fa2f

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -18,11 +21,7 @@ export abstract class InputMediaDeviceManager<
   /**
    * @internal
    */
-  enablePromise?: Promise<void>;
-  /**
-   * @internal
-   */
-  disablePromise?: Promise<void>;
+  statusChangePromise?: Promise<void>;
```
myandrienko, line 24: Store enable/disable promise in a single value. We can use `this.state.pendingStatus` to differentiate between the two, but usually that is not necessary.

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -63,16 +63,23 @@ export abstract class InputMediaDeviceManager<
...
-    this.enablePromise = this.unmuteStream();
-    try {
-      await this.enablePromise;
-      this.state.setStatus('enabled');
-      this.enablePromise = undefined;
-    } catch (error) {
-      this.enablePromise = undefined;
-      throw error;
-    }
+    if (this.state.pendingStatus === 'enabled') return;
```
myandrienko, line 66: 1. If there already is a pending request to enable device - do nothing, the device will end up enabled anyway.

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -63,16 +63,23 @@ export abstract class InputMediaDeviceManager<
...
+    const signal = this.nextAbortableStatusChangeRequest('enabled');
```
myandrienko, line 67: 2. Cancel any pending request that hasn't started yet.
3. Set pending status to `enabled`.

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -63,16 +63,23 @@ export abstract class InputMediaDeviceManager<
...
+      } finally {
+        if (!signal.aborted) {
+          if (!signal.aborted) this.resetStatusChangeRequest();
+        }
+      }
+    };
+    this.statusChangePromise = this.statusChangePromise
+      ? this.statusChangePromise.then(doEnable)
+      : doEnable();
+    await this.statusChangePromise;
```
myandrienko, line 82: 4. If there's already a pending request in the works, queue to run after it. Otherwise, run immediately.

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -63,16 +63,23 @@ export abstract class InputMediaDeviceManager<
...
-      this.state.setStatus('enabled');
-      this.enablePromise = undefined;
-    } catch (error) {
-      this.enablePromise = undefined;
-      throw error;
-    }
+    if (this.state.pendingStatus === 'enabled') return;
+    const signal = this.nextAbortableStatusChangeRequest('enabled');
+    const doEnable = async () => {
+      if (signal.aborted) return;
```
myandrienko, line 69: 5. Check if the request was canceled. That can happen if the request was scheduled to run, but then a new request came in and canceled it (see p. 2).

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -63,16 +63,23 @@ export abstract class InputMediaDeviceManager<
...
+      try {
+        await this.unmuteStream();
+        this.state.setStatus('enabled');
```
myandrienko, line 72: 6. Do the actual work.

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -63,16 +63,23 @@ export abstract class InputMediaDeviceManager<
...
+      } finally {
+        if (!signal.aborted) {
+          if (!signal.aborted) this.resetStatusChangeRequest();
+        }
```
myandrienko, line 76: 7. If we are the last request in the queue - clean up all the pending stuff.

packages/react-sdk/src/components/CallControls/ScreenShareButton.tsx
```diff
@@ -26,10 +26,11 @@ export const ScreenShareButton = (props: ScreenShareButtonProps) => {
...
 
-  const { screenShare, isMute: amIScreenSharing } = useScreenShareState();
-  const disableScreenShareButton = amIScreenSharing
-    ? isSomeoneScreenSharing || isScreenSharingAllowed === false
-    : false;
+  const { screenShare, optimisticIsMute } = useScreenShareState();
+  const amIScreenSharing = !optimisticIsMute;
+  const disableScreenShareButton =
+    !amIScreenSharing &&
+    (isSomeoneScreenSharing || isScreenSharingAllowed === false);
```
myandrienko, line 33: This was breaking my brain until I realized that the `amIScreenSharing` variable was named incorrectly.

santhoshvai:
Great! we would need to add this in react-native as well

cc @khushal87

myandrienko:
> Great! we would need to add this in react-native as well
> 
> cc @khushal87

Hopefully it'll be very easy for you, just use `optimisticStatus` instead of `status` for the UI 🤞

oliverlaz (approved):
Looks good to me. @szuperaz would you like to take a look as well?

szuperaz:
> Looks good to me. @szuperaz would you like to take a look as well?

I'll take a quick look now

szuperaz (approved):
Really nice solution!

PR #1316, opened 2024-04-15
feat: update open api spec DON'T MERGE

szuperaz:

PR #1343, opened 2024-05-08
feat: align with the latest openapi

myandrienko:


packages/client/src/rtc/videoLayers.ts
```diff
@@ -90,9 +90,11 @@ export const getComputedMaxBitrate = (
     const currentPixels = currentWidth * currentHeight;
     const targetPixels = targetWidth * targetHeight;
     const reductionFactor = currentPixels / targetPixels;
-    return Math.round(targetResolution.bitrate * reductionFactor);
+    // TODO: bitrate is optional
+    return Math.round(targetResolution.bitrate! * reductionFactor);
   }
-  return targetResolution.bitrate;
+  // TODO: bitrate is optional
+  return targetResolution.bitrate!;
```
myandrienko, line 97: Not sure what should be done here. Bitrate is optional in the TargetResolution type. Should we add some defaults here?

oliverlaz: Perhaps we should check this with our backend team.

packages/client/src/gen/coordinator/index.ts
```diff
@@ -1865,3011 +1909,4281 @@ export interface CallTranscriptionStoppedEvent {
...
+export interface PushNotificationSettingsInput {
   /**
    *
-   * @type {string}
-   * @memberof MemberRequest
+   * @type {NullBool}
+   * @memberof PushNotificationSettingsInput
    */
-  role?: string;
+  disabled?: NullBool;
```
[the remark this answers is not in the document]
oliverlaz: yep, looks suspicious. let's check this one with our backend team as well.

oliverlaz (commented):
@szuperaz it would be good if you check this one as well. I think there is an overlap with #1316

szuperaz:
> @szuperaz it would be good if you check this one as well. I think there is an overlap with #1316

Looks good to me, I closed the old PR in favor of this.

oliverlaz (approved):

szuperaz (approved):

PR #1351, opened 2024-05-13
feat: add full screen incoming call view when phone is locked for android

santhoshvai:
When phone is locked, now the incoming call view is directly opened on android

Screenshot_20240513-150224


sample-apps/react-native/dogfood/android/app/google-services.json
santhoshvai, line 1: unrelevant but important change.. aligns the json file with the service account present in the dashboard

khushal87:
At times I get that more than one call alert on the app as soon as it opens on the lock screen.

khushal87:
Screenshot_20240514-223332
After I unlock the device the notification doesn't go away but stays on the notification stack.

khushal87 (approved):

PR #1354, opened 2024-05-14
fix(react-native): mute and unmute device media stream optimistically

khushal87:


santhoshvai (approved):
screensharing to be done in a different PR

PR #1356, opened 2024-05-14
fix: popup-related UI updates

myandrienko:
1. Dropdown menus close when item is selected (e.g. camera device is selected)
2. Tooltip component is used to display hints for call controls

### DIFF
--- a/packages/react-sdk/src/components/Tooltip/WithTooltip.tsx
+++ b/packages/react-sdk/src/components/Tooltip/WithTooltip.tsx
@@ -1,9 +1,13 @@
-import { ComponentProps, useState } from 'react';
+import { ComponentProps, createContext, useContext, useState } from 'react';
 import { Tooltip, TooltipProps } from './Tooltip';
 import { useEnterLeaveHandlers } from './hooks';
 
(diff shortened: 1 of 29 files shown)


packages/react-sdk/src/components/Tooltip/WithTooltip.tsx
```diff
@@ -1,25 +1,29 @@
...
-  const { handleMouseEnter, handleMouseLeave, tooltipVisible } =
+  const { handleMouseEnter, handleMouseLeave, tooltipVisible, forceHide } =
     useEnterLeaveHandlers<HTMLDivElement>();
   const [tooltipAnchor, setTooltipAnchor] = useState<HTMLDivElement | null>(
     null,
   );
 
   return (
-    <>
+    <TooltipContext.Provider value={{ hideTooltip: forceHide }}>
```
myandrienko, line 26: Sometimes we want to hide tooltip even if the target is still hovered. One example is if it conflicts with an open menu.

https://github.com/GetStream/stream-video-js/assets/975978/b095b746-15ef-4932-a375-80e958997fef

PR #1365, opened 2024-05-22
chore: bump react to 18.3

myandrienko:


packages/react-native-sdk/package.json
```diff
@@ -122,7 +122,7 @@
     "@tsconfig/node14": "14.1.0",
     "@types/jest": "^29.5.1",
     "@types/lodash.merge": "^4",
-    "@types/react": "18.2.28",
+    "@types/react": "18.3.2",
```
oliverlaz, line 125: is there any reason not to use `^18.3.2` here @santhoshvai @khushal87?

packages/react-sdk/package.json
```diff
@@ -46,8 +46,8 @@
     "@rollup/plugin-typescript": "^11.1.6",
     "@stream-io/audio-filters-web": "workspace:^",
     "@stream-io/video-styling": "workspace:^",
-    "react": "^18.2.0",
-    "react-dom": "^18.2.0",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
```
oliverlaz, line 50: I just noticed this but I think we need to list `@types/react` and `@types/react-dom` in the dev dependencies. It works now because we have them hoisted to the root.

oliverlaz: Also, I'm not sure if we need `react-dom` as well, as we aren't using anything from it in this package.
What do you think?

sample-apps/react-native/dogfood/package.json
```diff
@@ -32,7 +32,7 @@
     "@stream-io/flat-list-mvcp": "^0.10.3",
     "@stream-io/react-native-webrtc": "118.0.1",
     "@stream-io/video-react-native-sdk": "workspace:^",
-    "react": "18.2.0",
+    "react": "18.3.1",
```
oliverlaz, line 35: Same here: @khushal87 @santhoshvai

sample-apps/react-native/expo-video-sample/package.json
```diff
@@ -27,7 +27,7 @@
     "expo-splash-screen": "~0.26.4",
     "expo-status-bar": "~1.11.1",
     "expo-task-manager": "~11.7.1",
-    "react": "18.2.0",
+    "react": "18.3.1",
```
oliverlaz, line 30: And here :)

packages/react-sdk/package.json
myandrienko: I agree with both points. Fixed in https://github.com/GetStream/stream-video-js/pull/1365/commits/42f550f224dd20d545dc4ec7fca30e8c7a6efc48

sample-apps/react-native/dogfood/package.json
santhoshvai: we need to keep it pinned in apps for sure, as RN version is tied to React..

for example see the [latest React conference app](https://github.com/expo/react-conf-app/blob/bb9016a22c95e11a3a8682ee9e3c6f1758f98f15/package.json#L47)

oliverlaz (approved):

PR #1372, opened 2024-05-24
fix: add LivestreamPlayer wrapper component

oliverlaz:
### Overview

Provides a new `LivestreamPlayer` wrapper component similar to the one available in iOS.

### DIFF
--- a/packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
+++ b/packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
@@ -1,0 +1,49 @@
+import { useEffect, useState } from 'react';
+import { Call } from '@stream-io/video-client';
+import { useStreamVideoClient } from '@stream-io/video-react-bindings';
+import {
+  LivestreamLayout,
+  LivestreamLayoutProps,
+  StreamCall,
+} from '../../core';
+
+export type LivestreamPlayerProps = {
+  /**
+   * The call type. Usually `livestream`.
+   */
+  callType: string;
+  /**
+   * The call ID.
+   */
+  callId: string;
+  /**
+   * The props for the {@link LivestreamLayout} component.
+   */
+  layoutProps?: LivestreamLayoutProps;
+};
+
+export const LivestreamPlayer = (props: LivestreamPlayerProps) => {
+  const { callType, callId, layoutProps } = props;
+  const client = useStreamVideoClient();
+  const [call, setCall] = useState<Call>();
+  useEffect(() => {
+    if (!client) return;
+    const myCall = client.call(callType, callId);
+    setCall(myCall);
+    myCall.join().catch((e) => {
+      console.error('Failed to join call', e);
+    });
+    return () => {
+      myCall.leave().catch((e) => {
+        console.error('Failed to leave call', e);
+      });
+      setCall(undefined);
+    };
+  }, [callId, callType, client]);
+
+  return (
+    <StreamCall call={call}>
+      <LivestreamLayout {...layoutProps} />
+    </StreamCall>
+  );
+};
--- a/packages/react-sdk/index.ts
+++ b/packages/react-sdk/index.ts
@@ -6,6 +6,7 @@ export * from '@stream-io/video-react-bindings';
 export * from './src/core';
 
 export * from './src/components';
+export * from './src/wrappers';
 export * from './src/translations';
 export {
   useHorizontalScrollPosition,
--- a/packages/react-sdk/src/wrappers/LivestreamPlayer/index.ts
+++ b/packages/react-sdk/src/wrappers/LivestreamPlayer/index.ts
@@ -1,0 +1,1 @@
+export * from './LivestreamPlayer';
--- a/packages/react-sdk/src/wrappers/index.ts
+++ b/packages/react-sdk/src/wrappers/index.ts
@@ -1,0 +1,1 @@
+export * from './LivestreamPlayer';


packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
```diff
@@ -0,0 +1,49 @@
...
+  /**
+   * The props for the {@link LivestreamLayout} component.
+   */
+  layoutProps?: LivestreamLayoutProps;
+};
+
+export const LivestreamPlayer = (props: LivestreamPlayerProps) => {
+  const { callType, callId, layoutProps } = props;
+  const client = useStreamVideoClient();
+  const [call, setCall] = useState<Call>();
```
myandrienko, line 28: BTW, I always wondered, why we don't recommend this in the tutorial:

```js
const call = useMemo(
  () => client.call(callType, callId),
  [client, callType, callId]
);
useEffect(() => {
  call.join().catch(/* ... */);
  return () => call.leave().catch(/* ... */);
}, [call]);
```

Feels more straightforward, plus saves one render.

myandrienko: Oh, lol, didn't notice @santhoshvai's comment above saying the exact same thing!

packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
```diff
@@ -0,0 +1,49 @@
...
+  useEffect(() => {
+    if (!client) return;
+    const myCall = client.call(callType, callId);
+    setCall(myCall);
+    myCall.join().catch((e) => {
+      console.error('Failed to join call', e);
+    });
+    return () => {
+      myCall.leave().catch((e) => {
```
khushal87, line 37: Sorry for neat-picking but I believe try-catch would be better here rather than .catch, since this is the pattern we follow most of the places in the code.

packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
```diff
@@ -0,0 +1,49 @@
...
+
+export const LivestreamPlayer = (props: LivestreamPlayerProps) => {
+  const { callType, callId, layoutProps } = props;
+  const client = useStreamVideoClient();
+  const [call, setCall] = useState<Call>();
+  useEffect(() => {
+    if (!client) return;
+    const myCall = client.call(callType, callId);
+    setCall(myCall);
+    myCall.join().catch((e) => {
```
khushal87, line 33: Sorry for neat-picking but I believe try-catch would be better here rather than .catch, since this is the pattern we follow most of the places in the code.

packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
oliverlaz: hey @myandrienko, @santhoshvai,

This combination fixes the problem with StrictMode but IMO it makes it a bit harder to work with the rest of the SDK as there isn't a way to "throw-away" the `call` instance anymore.

With `useState`, I can set the `call` to `undefined` once I don't need it anymore. This can't be done with `useMemo`. At least, not without re-inventing `useState` on top of `useMemo`.

`useMemo` will force users to do this:
```tsx
const MyApp = () => {
   const call = useMemo();
   useEffect(() => {}, []);
   return (
     <StreamCall call={call}>
        <MyExtraComponent />
     </StreamCall>
   );
}

const MyExtraComponent = () => {
  const { useCallCallingState } = useCallStateHooks();
  const callingState = useCallCallingState();
  
  const inCall = callingState !== CallingState.IDLE && callingState !== CallingState.LEFT;
  
  if (!inCall) return null;
  return <MyUI />;
}
```

instead of:
```tsx
const [call, setCall] = useState();
useEffect(() => {}, [...]);

if (!call) return null;
return <MyUi />;
```

Checking whether one has a valid `call` instance would be more complex as one would always have to peek into the `call.state` instead of doing simple `if (!call) {}` checks.

IMO, this is a good tradeoff for one extra render :)
What do you think about this?

packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
oliverlaz: hey @khushal87, we can't go without `.catch()` inside a `useEffect()`. This is because `useEffect` doesn't work with async code that well. Consider the following code:
```tsx
useEffect(() => {
  const effect = async () => {
    try { 
      await doStuff(); 
    } catch (e) { 
      handle(e); 
    }
  }
  effect();
}, []);
```

Although now `effect` wraps everything in a `try/catch` block, it still needs to have a `.catch()` cause to handle the possible future rejections. So, we would have to implement this effect like this:

```tsx
useEffect(() => {
  const effect = async () => {
    try { 
      await doStuff(); 
    } catch (e) { 
      handle(e); 
    }
  }
  effect().catch((err) => handle(err));
}, []);
```

Ultimately, we can simplify to:
```tsx
useEffect(() => {
  doStuff().catch((err) => handle(err));
}, []);
```

packages/react-sdk/src/wrappers/LivestreamPlayer/LivestreamPlayer.tsx
arnautov-anton: I feel like:
```tsx
const call = useMemo(() => client?.call(callType, callId), [client, callId, callType]);
```
should make the `call` variable either `Call` or `undefined` at which point you can definitelly provide a check for:
```tsx
!call
```
right? Change of state (either `client`, `callId` or `callType`) should esentially call a cleanup function with the instance of that scope and provide a new instance to begin "connecting" to. 

Unless client is always available and your current setup should be setting the `call` when the actuall call connects? Like this:

```tsx
 const [call, setCall] = useState<Call>();
  useEffect(() => {
    if (!client) return;
    const myCall = client.call(callType, callId);
    myCall.join().then(() => {
      setCall(myCall);
    }).catch((e) => {
      console.error('Failed to join call', e);
    });
    return () => {
      myCall.leave().catch((e) => {
        console.error('Failed to leave call', e);
      });
      setCall(undefined);
    };
  }, [callId, callType, client]);
```

arnautov-anton: > now what? how can I set the `call` to `undefined` without an additional state property?|

I don't think you have to do anything as the new `call`  is coming in right away - correct?
The whole setup would look something like:

```ts
const call = useMemo(() => {
  if (!client || !type || !id) return undefined;
  return client.call(type, id);
}, [client, type, id]);

useEffect(() => {
  if (!call) return;

  const joinPromise = call.join().catch(console.error);

  return () => {
    joinPromise.then(() => call.leave());
  };
});
```

arnautov-anton: Or maybe there's a some form of a limitation in terms of state that I'm forgetting but this should work just fine I think.

santhoshvai: got you here, useMemo means you can avoid a rerender, but you cant drop the call on leave easily

PR #1373, opened 2024-05-25
feat(react-native): add livestream player component

santhoshvai:


packages/react-native-sdk/src/components/Livestream/LivestreamPlayer/LivestreamPlayer.tsx
```diff
@@ -0,0 +1,59 @@
...
+  const call = useMemo(
+    () => client?.call(callType, callId),
+    [callType, callId, client],
+  );
+
+  useEffect(() => {
+    if (!call) {
+      return;
+    }
+    call.join().catch((e) => {
```
khushal87, line 40: Can we have try catch instead of .then.catch?

packages/react-native-sdk/src/components/Livestream/LivestreamPlayer/LivestreamPlayer.tsx
```diff
@@ -0,0 +1,59 @@
...
+      console.error('Failed to join call', e);
+    });
+    return () => {
+      call.leave().catch((e) => {
```
khushal87, line 44: Same here

packages/react-native-sdk/src/components/Livestream/LivestreamPlayer/LivestreamPlayer.tsx
santhoshvai: my opinion is that it doesnt matter here. Both are fine.

When we chain promises I would rather go for try/catch, for single promise, just having catch is fine in a useEffect scenario

khushal87 (approved):

PR #1378, opened 2024-05-28
fix: improved input device error handling

oliverlaz:
### Overview

Improves the error handling for the input devices. 

In some conditions, our Device Manager APIs could throw an unrecoverable error that couldn't be handled anywhere as we weren't propagating it accordingly. Another problem is async observable subscriptions. Currently, there isn't a way for someone to handle errors in those subscriptions. Until we find a better day, I've wrapped them with a try/catch block.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -388,7 +388,9 @@ export class Call {
         const currentUserId = this.currentUserId;
         if (currentUserId && blockedUserIds.includes(currentUserId)) {
           this.logger('info', 'Leaving call because of being blocked');
-          await this.leave({ reason: 'user blocked' });
+          await this.leave({ reason: 'user blocked' }).catch((err) => {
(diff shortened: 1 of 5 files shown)


szuperaz (approved):

packages/client/src/Call.ts
```diff
@@ -600,7 +602,7 @@ export class Call {
       this.clientStore.registerCall(this);
     }
 
-    this.applyDeviceConfig();
+    await this.applyDeviceConfig();
```
szuperaz, line 605: Do we need `await` here if the `applyDeviceConfig` now catches the error?

oliverlaz: I was thinking about whether to add an `await` here or not. Eventually, I added it because it keeps the linear execution flow. There wasn't any other reason for the `await`.

szuperaz: I'm fine with this change, I just was thinking if I'm missing something

PR #1379, opened 2024-05-30
feat: support new rtmp endpoints in video client

nivklein:


packages/client/src/gen/coordinator/index.ts
```diff
@@ -911,6 +911,91 @@ export interface CallParticipantResponse {
...
+ * @export
+ * @interface CallRTMPBroadcast
+ */
+export interface CallRTMPBroadcast {
+  /**
+   *
+   * @type {string}
+   * @memberof CallRTMPBroadcast
+   */
+  CallID: string;
```
szuperaz, line 925: This case is a bit weird, open api models usually follow snake case, could this be a mistake?

packages/client/src/Call.ts
```diff
@@ -634,6 +638,21 @@ export class Call {
     return response;
   };
 
+  startRTMPBroadcast = async (data: StartRTMPBroadcastsRequest) => {
+    console.log('Niv is VERY VERY 1232 awesome >>>>')
```
szuperaz, line 642: I agree, but probably the `console.log` was probably left here by mistake 🙂

PR #1388, opened 2024-06-04
feat: improve `isSupported` method for noise cancellation

szuperaz:


packages/react-sdk/src/components/NoiseCancellation/NoiseCancellationProvider.tsx
```diff
@@ -76,10 +76,33 @@ export const NoiseCancellationProvider = (
...
+          err,
+        ),
+      );
+  }, [noiseCancellation]);
+
+  useEffect(() => {
+    setIsSupported(
+      isSupportedByBrowser && hasCapability && noiseCancellationAllowed,
+    );
+  }, [isSupportedByBrowser, hasCapability, noiseCancellationAllowed]);
```
myandrienko, line 105: `isSupported` doesn't have to be state in this case. It can be just

```
const isSupported = isSupportedByBrowser && hasCapability && noiseCancellationAllowed;
```

(see https://react.dev/learn/you-might-not-need-an-effect#updating-state-based-on-props-or-state for the explanation of this pattern)

packages/react-sdk/src/components/NoiseCancellation/NoiseCancellationProvider.tsx
```diff
@@ -76,10 +76,31 @@ export const NoiseCancellationProvider = (
...
+      result
+        .then((_isSupportedByBrowser) =>
+          setIsSupportedByBrowser(_isSupportedByBrowser),
+        )
+        .catch((err) =>
+          console.error(
+            `Can't determine if noise cancellation is supported`,
+            err,
+          ),
+        );
```
myandrienko, line 98:
```suggestion
      let cancel = false;
      result
        .then((_isSupportedByBrowser) => {
          if (!cancel) {
            setIsSupportedByBrowser(_isSupportedByBrowser)
          }
        })
        .catch((err) =>
          console.error(
            `Can't determine if noise cancellation is supported`,
            err,
          ),
        );
        return () => (cancel = true);
```

myandrienko: Also, can we add a quick note somewhere in the doc comments what "undefined" means for `isSupported`?

szuperaz: Yes, updated the react guide

PR #1391, opened 2024-06-05
feat: use logger from client for react native sdk

santhoshvai:
A custom logger can be passed to the video client

In RN SDK, we did not use that logger, this PR fixes that


packages/react-native-sdk/.prettierrc
santhoshvai, line 1: unrelated, but vscode randomly switches to root prettier rc if we dont add this.

khushal87 (approved):

PR #1392, opened 2024-06-05
feat: add concurrency helpers

myandrienko:
Adds three concurrency helpers:
1. `withoutConcurrency` - accepts async functions and makes sure they run one after another, and never in parallel.
2. `withCancelation` - does the same thing, but also cancels all other async functions when a new function is scheduled. Useful for actions that override each other, like enabling and disabling camera.
3. `createSafeAsyncSubscription` adds an async listener to an Rx observable, and makes sure listeners never run in parallel, even if updates come in quicker than handlers can run.

As an example, those utilities are adopted in two places, where ad-hoc solutions were previously added:
https://github.com/GetStream/stream-video-js/pull/1371
https://github.com/GetStream/stream-video-js/pull/1342


packages/client/src/store/rxUtils.ts
```diff
@@ -72,3 +73,24 @@ export const createSubscription = <T>(
...
+
+/**
+ * Creates a subscription and returns a function to unsubscribe. Makes sure that
+ * only one async handler runs at the same time. If updates come in quicker than
+ * it takes for the current handler to finish, other handlers will wait.
+ *
+ * @param observable the observable to subscribe to.
+ * @param handler the async handler to call when the observable emits a value.
+ */
+export const createSafeAsyncSubscription = <T>(
```
szuperaz, line 85: Is this method used somewhere?

szuperaz (commented):
Looks good to me

packages/client/src/store/rxUtils.ts
myandrienko: Not right now. I wanted to use in places like this https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/Call.ts#L388-L397 where it's not safe to call `call.leave()` in a callback that's now awaited. But I don't want to cram to many changes in one PR.

szuperaz (approved):

PR #1394, opened 2024-06-07
fix: improve browser permission handling

myandrienko:
Adds a new `BrowserPermission` class that handles common tasks around browser permissions (like camera and microphone access):

1. Keeps track of the current permission state
2. Allows prompting for permission
3. Makes sure we don't prompt for permission too often

To make sure we don't prompt for permission too often, we separate all actions that require a permission into two groups. Low priority actions (like listing audio/video devices) are not allowed to prompt for permission more than once. High priority actions (like enabling an audio/video stream) can prompt for permissions again, it the prompt was previously dismissed.

The `BrowserPermission#prompt` method also doesn't allow running two prompts for the same permission at once.

Most of these changes are relevant to Chrome (Chromium), which is the only browser that allows dismissing permission prompts.

This PR also restructures `get[Audio/Video]Devices` observables, so that they react to the change in browser permissions.


packages/client/src/devices/BrowserPermission.ts
```diff
@@ -0,0 +1,152 @@
...
+import { getLogger } from '../logger';
+import { disposeOfMediaStream } from './devices';
+import { withoutConcurrency } from '../helpers/concurrency';
+
+interface BrowserPermissionConfig {
+  constraints: DisplayMediaStreamOptions;
+  queryName: PermissionName;
+}
+
+export class BrowserPermission {
```
myandrienko, line 12: 🍓 Start review here.

packages/client/src/devices/BrowserPermission.ts
```diff
@@ -0,0 +1,152 @@
...
+  async prompt({
+    forcePrompt = false,
+    throwOnNotAllowed = false,
+  }: { forcePrompt?: boolean; throwOnNotAllowed?: boolean } = {}) {
+    await withoutConcurrency(
+      `permission-prompt-${this.permission.queryName}`,
+      async () => {
+        if (
+          (await this.getState()) !== 'prompt' ||
+          (this.wasPrompted && !forcePrompt)
```
myandrienko, line 75: BrowserPermission will not allow re-prompting for a permission unless explicitly allowed to do so.

packages/client/src/devices/BrowserPermission.ts
```diff
@@ -0,0 +1,152 @@
...
+
+          return isGranted;
+        }
+
+        try {
+          this.wasPrompted = true;
+          const stream = await navigator.mediaDevices.getUserMedia(
+            this.permission.constraints,
+          );
+          disposeOfMediaStream(stream);
```
myandrienko, line 94: To prompt, we `getUserMedia` and then immediately dispose the media stream.

packages/client/src/devices/BrowserPermission.ts
```diff
@@ -0,0 +1,152 @@
...
+          });
+          throw e;
+        }
+      },
+    );
+  }
+
+  listen(cb: (state: PermissionState) => void) {
+    this.listeners.add(cb);
+    if (this.state) cb(this.state);
```
myandrienko, line 121: Helpful feature: the listener will immediately fire if the permission state is already known.

packages/client/src/devices/BrowserPermission.ts
```diff
@@ -0,0 +1,152 @@
...
+      (handler) => this.listen(handler),
+      (handler, unlisten) => unlisten(),
+    ).pipe(
+      // In some browsers, the 'change' event doesn't reliably emit and hence,
+      // permissionState stays in 'prompt' state forever.
+      // Typically, this happens when a user grants one-time permission.
+      // Instead of checking if a permission is granted, we check if it isn't denied
+      map((state) => state !== 'denied'),
+    );
+  }
```
myandrienko, line 136: Creates an observable from a listener on `permissionStaus.onchange`.

packages/client/src/devices/devices.ts
```diff
@@ -16,49 +19,28 @@ import { getLogger } from '../logger';
...
-      if (needsGetUserMedia) {
-        let mediaStream: MediaStream | undefined;
-        try {
-          mediaStream = await navigator.mediaDevices.getUserMedia(constraints);
-          devices = await navigator.mediaDevices.enumerateDevices();
-        } finally {
-          if (mediaStream) disposeOfMediaStream(mediaStream);
-        }
+      if (shouldPromptForBrowserPermission) {
+        await permission.prompt({ throwOnNotAllowed: true });
```
myandrienko, line 32: Enumerating devices is not an important enough action to allow re-prompting for a permission, so we use the default here (`forcePrompt: false`).

packages/client/src/devices/devices.ts
```diff
@@ -87,97 +69,92 @@ const videoDeviceConstraints = {
...
-    map((values) => values.filter((d) => d.kind === 'audioinput')),
+export const getAudioDevices = lazy(() => {
+  return merge(
+    getDeviceChangeObserver(),
+    getAudioBrowserPermission().asObservable(),
+  ).pipe(
+    startWith(undefined),
+    concatMap(() => getDevices(getAudioBrowserPermission(), 'audioinput')),
+    shareReplay(1),
   );
```
myandrienko, line 119: An observable with a list of devices will now emit when permission changes.

packages/client/src/devices/devices.ts
```diff
@@ -197,7 +174,20 @@ export const getAudioStream = async (
       ...trackConstraints,
     },
   };
-  return getStream(constraints);
+
+  try {
+    await getAudioBrowserPermission().prompt({
+      throwOnNotAllowed: true,
+      forcePrompt: true,
```
myandrienko, line 181: Enabling camera/microphone is an important enough operation to allow prompting for the permission again if the prompt was previously dismissed.

szuperaz (approved):
Sorry for taking so long with the review, it looks good to me. I think it might be good to ask for a review and test from the RN team as well.

packages/client/src/helpers/lazy.ts
```diff
@@ -0,0 +1,15 @@
+const uninitialized = Symbol();
```
arnautov-anton, line 1:
```suggestion
const uninitialized = Symbol('uninitialized');
```

packages/client/src/helpers/lazy.ts
```diff
@@ -0,0 +1,15 @@
...
+
+/**
+ * Lazyly creates a value using a provided factory
```
arnautov-anton, line 4:
```suggestion
 * Lazily creates a value using a provided factory
```

PR #1399, opened 2024-06-11
fix: infinitely adjustable blur filter

myandrienko:
Previously, we had three hard-coded settings for the background blur filter: low, medium and high.

We now add an option to provide an arbitrary numeric value which will determine the filter strength (it's used a standard deviation value for the gaussian blur filter).

Hard-coded values are also adjusted.

### DIFF
--- a/packages/react-sdk/docusaurus/docs/React/10-advanced/06-apply-video-filters.mdx
+++ b/packages/react-sdk/docusaurus/docs/React/10-advanced/06-apply-video-filters.mdx
@@ -105,6 +105,10 @@ export const MyBackgroundFilterSettings = () => {
 };
 ```
 
+:::note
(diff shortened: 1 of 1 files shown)


packages/video-filters-web/src/webgl2/backgroundBlurStage.ts
```diff
@@ -62,12 +62,22 @@ function buildBlurPass(
...
-        : [0.627027027, 0.3445945946, 0.2216216216, 0.0540540541, 0.0162162162];
+  const sigma =
+    typeof blurLevel === 'number'
+      ? blurLevel
+      : blurLevel === 'low'
+        ? 2
+        : blurLevel === 'medium'
+          ? 4
+          : 6;
+  const windowSize = Math.max(1, Math.floor(sigma * 3));
```
myandrienko, line 73: "Three sigma" is a big enough window for a good quality blur.

PR #1404, opened 2024-06-17
docs: update the chat with video guide

oliverlaz:
### Overview

Updates the code example to include all channel members as call members as well (including self!).


szuperaz (approved):
I think on the long run it would also make sense to change this tutorial to use join links instead of ringing calls?

oliverlaz:
> I think on the long run it would also make sense to change this tutorial to use join links instead of ringing calls?

Yeah, I think this whole tutorial needs to be rewritten. We wrote this one quite early and since then, a lot of things are improved in general.

PR #1413, opened 2024-06-20
fix: stop background filter when video track stops

myandrienko:
Currently the background filter keeps working on the (black) frames after camera is disabled and video track is stopped. That's because the filter doesn't know that the source track is stopped.

This PR adds an optional cleanup function that can be specified when registering a filter. This cleanup function is called in two cases:

1. The filter is unregistered
2. The source track is muted

So now background filters specify a cleanup function that remove the canvas element, thus stopping the rendering pipeline.


oliverlaz (approved):
Looks good to me. Can you update the documentation as well?
- https://getstream.io/video/docs/react/guides/camera-and-microphone/#camera-and-microphone-filters
- https://getstream.io/video/docs/javascript/guides/camera-and-microphone/#camera-and-microphone-filters

myandrienko:
Closing this in favor of #1415.

PR #1415, opened 2024-06-21
fix: refactor background filters

myandrienko:


packages/react-sdk/src/components/BackgroundFilters/BackgroundFilters.tsx
```diff
@@ -197,166 +197,124 @@ export const BackgroundFiltersProvider = (
...
 
-    const filter = canvasRef.captureStream();
-    resolveFilter(filter);
-    return () => {
-      disposeOfMediaStream(filter);
-    };
-  }, [canvasRef]);
+  const start = useCallback(
+    (ms: MediaStream) => {
+      return new Promise<[MediaStream, () => void]>((resolve, reject) => {
```
myandrienko, line 239: That's the whole point of this PR: to rewrite background filter pipeline as a single linear function, instead of an effects-based logic.

packages/react-sdk/src/components/BackgroundFilters/BackgroundFilters.tsx
```diff
@@ -197,166 +197,124 @@ export const BackgroundFiltersProvider = (
...
+          const outputStream = canvasEl.captureStream();
+
+          resolve([
+            outputStream,
+            () => {
+              renderer.dispose();
+              videoEl.srcObject = null;
+              disposeOfMediaStream(outputStream);
+            },
+          ]);
```
myandrienko, line 278: With the new filter API, filter returns a stream and a cleanup function. The cleanup function will be called when the filter is unregistered or when the source stream is muted.

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -142,20 +151,29 @@ export abstract class InputMediaDeviceManager<
...
+    return {
+      ready,
+      unregister: () =>
+        withoutConcurrency(this.filterRegistrationConcurrencyTag, async () => {
+          const cleanup = entry[1];
+          cleanup?.();
+          this.filters = this.filters.filter((f) => f !== entry);
+          await this.applySettingsToStream();
+        }),
     };
```
myandrienko, line 177: We move the logic preventing unregistering before finishing registering into the client. Note how both registering and unregistering logic is wrapped in `withoutConcurrency`.

This can be further optimized. Most of the times we unregister one filter to immediately register another. We can introduce a method called `replaceFilter` that will only call `applySettingsToStream` once. That way we won't get flickering when switching blur levels or background images.

myandrienko: I don't like the API with filters returning [stream, cleanup] tuple, leaving it in this draft, but will probably change before merging.

packages/client/src/devices/MicrophoneManager.ts
```diff
@@ -139,8 +139,10 @@ export class MicrophoneManager extends InputMediaDeviceManager<MicrophoneManager
         },
       );
 
-      this.noiseCancellationRegistration = this.registerFilter(
-        noiseCancellation.toFilter(),
+      const { ready: noiseCancellationRegistration, unregister } =
+        this.registerFilter(noiseCancellation.toFilter());
+      this.noiseCancellationRegistration = noiseCancellationRegistration.then(
+        () => unregister,
```
myandrienko, line 145: I had to change things around the noise cancellation filter a bit, but I haven't looked to deep into it. Probably the code can be simplified in this file.

packages/react-sdk/src/components/BackgroundFilters/BackgroundFilters.tsx
```diff
@@ -194,167 +213,145 @@ export const BackgroundFiltersProvider = (
...
+            reject(new Error('No video tracks in input media stream'));
+          }
+
+          const trackSettings = track.getSettings();
+          flushSync(() =>
+            setVideoSize({
+              width: trackSettings.width ?? 0,
+              height: trackSettings.height ?? 0,
+            }),
+          );
```
oliverlaz, line 293: Do we really need to do `flushSync`? Can we do this differently?

packages/react-sdk/src/components/BackgroundFilters/BackgroundFilters.tsx
```diff
@@ -194,167 +213,145 @@ export const BackgroundFiltersProvider = (
...
+    <div className="str-video__background-filters">
       <video
         className={clsx(
           'str-video__background-filters__video',
-          height > width && 'str-video__background-filters__video--tall',
+          videoSize.height > videoSize.width &&
+            'str-video__background-filters__video--tall',
         )}
-        ref={setVideoRef}
-        autoPlay
```
oliverlaz, line 299: Did we intentionally remove `autoPlay` because we want to control this via `ref.play()`?

myandrienko: Yes, I just didn't think it's necessary since we always call `ref.play()`. Are there any advantages to having `autoPlay`?

oliverlaz (approved):

packages/react-sdk/src/components/BackgroundFilters/BackgroundFilters.tsx
oliverlaz: no, I think it is good like this. `autoPlay` should immediately kick-start the playback once `srcObject` is set and there is enough metadata loaded.

PR #1419, opened 2024-06-25
fix: unpublish must also release track while using react native

santhoshvai:
While looking at the webrtc code for react native, I realised that there is a release function for media tracks. Which does the garbage collection natively. And we did not use it on unpublish. This PR adds that.

### DIFF
--- a/packages/client/src/rtc/Publisher.ts
+++ b/packages/client/src/rtc/Publisher.ts
@@ -341,16 +341,23 @@ export class Publisher {
     const transceiver = this.pc
       .getTransceivers()
       .find((t) => t === this.transceiverRegistry[trackType] && t.sender.track);
+    const track = transceiver?.sender.track;
     if (
       transceiver &&
-      transceiver.sender.track &&
-      (stopTrack
-        ? transceiver.sender.track.readyState === 'live'
-        : transceiver.sender.track.enabled)
+      track &&
+      (stopTrack ? track.readyState === 'live' : track.enabled)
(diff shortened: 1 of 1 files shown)


packages/client/src/rtc/Publisher.ts
```diff
@@ -341,16 +341,23 @@ export class Publisher {
...
-      stopTrack
-        ? transceiver.sender.track.stop()
-        : (transceiver.sender.track.enabled = false);
+      if (stopTrack) {
+        track.stop();
+        // @ts-expect-error release() is present in react-native-webrtc
+        // and must be called to dispose the track in the stream
+        if (typeof track.release === 'function') {
+          // @ts-expect-error
+          track.release();
```
oliverlaz, line 356: We have a similar code here: https://github.com/GetStream/stream-video-js/blob/ed37b7c1516847aa5809eae57b0f7855452ae470/packages/client/src/devices/InputMediaDeviceManager.ts#L209-L228

is this redundant or we have to handle this at both places?

oliverlaz: Follow-up to our Slack discussion with @santhoshvai 
- the new code is added primarily for screen share tracks, as they don't go through InputMediaDeviceManager
- it is safe to call `.release()` multiple times, react-native-webrtc handles this well internally

oliverlaz (approved):

santhoshvai:
closing this PR in favor of #1425

PR #1425, opened 2024-06-27
feat(react-native): enable android 14 compatibility for screensharing foreground service

santhoshvai:
* fixes memory leak of screensharing track
* ensures android 14 targetSdk compatibility, but still missing for ringing so not yet updated that part
* expo plugin updated


packages/react-native-sdk/src/hooks/useScreenShareButton.ts
```diff
@@ -0,0 +1,125 @@
...
+import React, { useRef, useEffect } from 'react';
+import { Platform, findNodeHandle, NativeModules } from 'react-native';
+import { usePrevious } from '../utils/hooks';
+import { useIsIosScreenshareBroadcastStarted } from './useIsIosScreenshareBroadcastStarted';
+
+// ios >= 14.0 or android - platform restrictions
+const CanDeviceScreenShare =
+  (Platform.OS === 'ios' &&
+    // @ts-ignore
+    Number.parseInt(Platform.Version.split('.')[0], 10) >= 14) ||
```
oliverlaz, line 16:
```suggestion
    Number.parseInt(Platform.Version.split('.')[0], 10) >= 14, 10) ||
```

Please always make sure to pass `radix` to the `parseInt` function as otherwise weird things can happen:
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt#radix

packages/react-native-sdk/src/hooks/useScreenShareButton.ts
```diff
@@ -0,0 +1,125 @@
...
+    iosScreenShareStartedFromSystem
+  );
+
+  const localParticipant = useLocalParticipant();
+  const hasPublishedScreenShare =
+    localParticipant && hasScreenShare(localParticipant);
+
+  // listens to iOS screen share broadcast started event from the system
+  useEffect(() => {
+    const run = async () => {
```
oliverlaz, line 67: As this function is marked as `async` perhaps you want to `await` some async functions inside? Perhaps `call.screenShare.enable()` and `call.screenShare.disable()`?

packages/react-native-sdk/src/hooks/useScreenShareButton.ts
```diff
@@ -0,0 +1,125 @@
...
+        const reactTag = findNodeHandle(screenCapturePickerViewiOSRef.current);
+        await NativeModules.ScreenCapturePickerViewManager.show(reactTag);
+        // After this the iOS screen share broadcast started/stopped event will be triggered
+        // and the useEffect listener will handle the rest
+      } else {
+        try {
+          await call?.screenShare.enable();
+          onScreenShareStartedHandler?.();
+        } catch (e) {
+          // ignored.. user didnt allow the screen share in the popup
```
oliverlaz, line 112: Please don't hide exceptions. It makes debugging harder than necessary. Can we have at least a log statement here?

packages/react-native-sdk/expo-config-plugin/src/withAndroidManifest.ts
```diff
@@ -15,17 +15,17 @@ type ManifestService = Unpacked<
...
-        android:foregroundServiceType="mediaProjection|microphone" />
+        android:foregroundServiceType="mediaProjection" />
  */
   let foregroundServiceType = 'microphone';
-  if (enableScreenshare) {
-    foregroundServiceType = 'mediaProjection|' + foregroundServiceType;
-  }
+  // if (enableScreenshare) {
+  //   foregroundServiceType = 'mediaProjection|' + foregroundServiceType;
+  // }
```
oliverlaz, line 28: Can we remove this code if we don't need it anymore?

packages/react-native-sdk/src/hooks/useScreenShareButton.ts
oliverlaz: btw, awesome stuff - switching to `call.screenShare` APIs 🚀

packages/react-native-sdk/src/hooks/useScreenShareButton.ts
santhoshvai: thanks

PR #1433, opened 2024-07-04
fix: allow reusing call instances after leaving

myandrienko:


packages/client/src/Call.ts
```diff
@@ -483,74 +502,74 @@ export class Call {
...
-    this.publisher = undefined;
+      this.sfuStatsReporter?.stop();
+      this.sfuStatsReporter = undefined;
 
-    this.sfuClient?.close(StreamSfuClient.NORMAL_CLOSURE, reason);
-    this.sfuClient = undefined;
+      this.subscriber?.close();
+      this.subscriber = undefined;
 
-    this.dispatcher.offAll();
```
myandrienko, line 529: 🍓 Not removing listeners from dispatcher.

That seems to be safe, since every object keeping a reference to a dispatcher (except for the call itself) is nullified on leave. So the call instance will be GC-ed properly even without `offAll`.

packages/client/src/Call.ts
```diff
@@ -276,35 +279,51 @@ export class Call {
...
-    );
-
     this.camera = new CameraManager(this);
     this.microphone = new MicrophoneManager(this);
     this.speaker = new SpeakerManager(this);
     this.screenShare = new ScreenShareManager(this);
   }
 
+  private async setup() {
+    await withoutConcurrency(this.joinLeaveConcurrencyTag, async () => {
```
myandrienko, line 289: Prevents a race condition with `leave`. It's important to wait for `leave` to finish before we set everything back up again. See the test case here: https://github.com/GetStream/stream-video-js/pull/1433/files#diff-912706ed8d680e55d746f9765aa5090d64771196a9894262bed3e0b73d9a8c97R92-R110

oliverlaz (approved):

PR #1434, opened 2024-07-04
feat(svc-codec): VP9 and AV1 support

oliverlaz:
### Overview

Enables support for SVC codecs (VP9 and AV1).

Adds new API that allows codec overriding:
- `call.updatePublishOptions({ preferredCodec: 'vp9' })`

### Implementation notes:
- the SDK will now check whether the platform supports the preferred codec, otherwise, it will fall back to a known supported codec
  -  for testing purposes, use `call.updatePublishOptions({ forceCodec: 'vp9' })`
The SDK ships a predefined bitrate lookup table now; in the future, this mapping should come from our backend. The bitrates used here enable ~30-50% less network usage while maintaining the same video quality (h264, vp9).
- when SVC codec is used, the SDK now supports the dynamic update of the `scalabilityMode`
- in Simulcast, the SDK can turn off any layer when instructed by the SFU - this allows further CPU/bandwidth reductions
- default codecs:
  - Firefox - always uses `vp8` as it has intermittent issues when other codecs are used
  - Safari - always uses `h264` as it is hardware accelerated and yields best performance
  - RN iOS and iPadOS - always uses `h264` as it is hardware-accelerated
  - RN Android - uses `vp8` by default

https://github.com/GetStream/protocol/pull/776/


packages/react-native-sdk/README.md
```diff
@@ -136,10 +136,10 @@ Stream's video roadmap and changelog are available [here](https://github.com/Get
 - [x] Speaking while muted
 - [x] Demo app on play-store and app-store
 - [x] Transcriptions
-- [ ] Speaker management (needs docs)
-- [ ] PiP on iOS
-- [ ] Audio & Video filters
-- [ ] CPU usage improvement
+- [x] Speaker management
+- [x] PiP on iOS
+- [x] Audio & Video filters
```
santhoshvai, line 141: I think we should mark audio filters still as TODO, that's the noise cancellation, no support yet on RN

oliverlaz: Ok, I'll break it in two

PR #1437, opened 2024-07-11
fix: default to h264 in Safari

oliverlaz:
### Overview

Enables H264 by default in Safari. 
This codec is hardware-accelerated and improves the general encoding performance on Apple devices when Safari is used.


szuperaz (approved):

myandrienko (approved):

santhoshvai:
@oliverlaz I added this long back for iPAD is time to keep it for iOS too? 

I think it makes sense to wait for until we prefer most supported profile for h264

https://github.com/GetStream/stream-video-js/blob/c99fff33ba2e6cd9b26fe040b1dc627015b14812/packages/client/src/rtc/Publisher.ts#L277-L284

oliverlaz:
Closing in favor of: #1434

PR #1439, opened 2024-07-11
feat: Reconnects v2

oliverlaz:
### Overview

Implements the Reconnects v2 flow. The implementation doesn't change the public API (not breaking); internally, it does change many things. The Reconnects v2 consists of three reconnect strategies:
- **FAST**: typically used when switching networks (wifi->5g or similar). This strategy takes the SFU-provided threshold, which normally is set to 25 seconds. The client attempts to connect to the same SFU and performs an ICE restart on the existing peer connections.
- **REJOIN**: used when FAST isn't applicable, failed or the SFU told us to do it. Drops all existing peer connections, closes the web socket to the SFU, fetches new credentials from the Coordinator, and finally establishes a connection to the SFU returned by the coordinator. The important detail here is, that we rejoin with a new `session_id` parameter. This strategy is analogous to a browser tab refresh.
- **MIGRATE**: when the SFU reaches its operational capacity or when it gets shut down due to maintenance or upgrade, it will ask connected clients to migrate away to another, healthy SFU. The migration is marked as complete, once the `participantMigrationComplete` event is emitted. If the migration fails, the client automatically attempts REJOIN.
- **DISCONNECT**: The SFU politely asks you to leave the call

### Implementation notes dump for easier review ###
- the original reconnection flow is completely removed and replaced with the new one
- we don't rely on WS status codes anymore, as they have proven to be unreliable across platforms/browsers
- with the new flow, the SFU can ask the client to perform a reconnect or a migration (signaled via the `error` event). This is super-helpful for cases where the auth token has expired and a new one should be fetched, the SFU is being updated, PeerConnections are broken and can't recover due to various reasons
- the client, when disconnecting from an SFU, now sends a reason why. We store this information and it should appear in the call stats
- SFU Client: to prevent race conditions on the server, the SFU client now waits until `JoinResponse` is received before it attempts to perform any RPC request. Also, these requests are now retried until they are successful, the SFU tells us not to do it anymore or the call has been left
- Concurrent event processing: we now guard from processing SFU WS events concurrently. ICE restarts, Subscriber offers, etc... are now applied in the correct order
- Orphaned tracks: as MediaTracks are sent on the PeerConnection, but the participant info is sent through a WebSocket, it could happen they arrive in the wrong order. Now, we keep a reference to these tracks in the call state and bind them to the appropriate participant once the info arrives
- Video FPS: all three track qualities (f, h, q) now have 30fps, providing a smoother experience
- The `network.changed` event is introduced, it emits every time the device detects being offline/online. It is decoupled from the existing `connection.changed` event. `connection.changed` represents two things: the device being online/offline or the coordinator socket is healthy or not.
- logging: the SFU client, peer connections, dispatcher, and the signal web socket now have a shared `logTag`. It makes the debugging easier.
- Coordinator unrecoverable error: when rejoining after a network loss, the call settings might have changed and the user might not have permission to join the call anymore or the call was marked as ended. The coordinator signals this with `unrecoverable: true` in the REST API response. The reconnect loop breaks and signals `RECONNECTION_FAILED`.


packages/client/src/Call.ts
```diff
@@ -1565,12 +1626,12 @@ export class Call {
...
-  private assertCallJoined = () => {
+  private waitUntilCallJoined = () => {
     return new Promise<void>((resolve) => {
       this.state.callingState$
         .pipe(
           takeWhile((state) => state !== CallingState.JOINED, true),
-          filter((s) => s === CallingState.JOINED),
+          filter((state) => state === CallingState.JOINED),
         )
         .subscribe(() => resolve());
```
arnautov-anton, line 1636:
```tsx
this.state.callingState$
    .pipe(
      takeWhile((state) => state !== CallingState.JOINED),
    )
    .subscribe({ complete: resolve });
```

oliverlaz: Thank you! I'll adjust accordingly!

PR #1451, opened 2024-07-30
fix: ringing state issues when call was already ended

santhoshvai:
see comments in the files

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -416,11 +416,43 @@ export class Call {
       // "ringing" mode effects and event handlers
       createSubscription(this.ringingSubject, (isRinging) => {
         if (!isRinging) return;
-        this.scheduleAutoDrop();
-        if (this.state.callingState === CallingState.IDLE) {
-          this.state.setCallingState(CallingState.RINGING);
+        const callSession = this.state.session;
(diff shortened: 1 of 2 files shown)


packages/client/src/Call.ts
santhoshvai, line 1: now checks conditions when the call should leave out ringing state to idle

packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
santhoshvai, line 1: fixes a bug where we didnt stop foreground service of incoming call notifications when call is joined

packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
```diff
@@ -90,9 +85,7 @@ export const useAndroidKeepCallAliveEffect = () => {
           );
           if (activeCallNotification) {
             // this means that we have a incoming call notification shown as foreground service and we must stop it
-            if (isAndroid7OrBelow) {
-              notifee.stopForegroundService();
-            }
+            notifee.stopForegroundService();
```
oliverlaz, line 88: Shouldn't we catch the possible rejections here?

oliverlaz (approved):

PR #1456, opened 2024-08-06
fix(react-native)!: make notifee to be optional

santhoshvai:
- [x] update codebase of the SDK
- [x] documentation changes
- [ ] test on android 24 emulator without notifee installed

### DIFF
--- a/packages/react-native-sdk/docusaurus/docs/reactnative/06-advanced/08-screensharing/02-setup/01-react-native.mdx
+++ b/packages/react-native-sdk/docusaurus/docs/reactnative/06-advanced/08-screensharing/02-setup/01-react-native.mdx
@@ -19,7 +19,6 @@ To be able to use the foreground service, the permission must be declared in the
 <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
 <!-- We declare the permissions to for using foreground service -->
 <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
-<uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE" />
 // highlight-next-line
 <uses-permission android:name="android.permission.FOREGROUND_SERVICE_MEDIA_PROJECTION" />
 ```
--- a/packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
+++ b/packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
@@ -1,15 +1,16 @@
 import { useCall, useCallStateHooks } from '@stream-io/video-react-bindings';
 import { useEffect, useRef } from 'react';
-import notifee, { AuthorizationStatus } from '@notifee/react-native';
 import { StreamVideoRN } from '../utils';
 import { Platform } from 'react-native';
 import { CallingState, getLogger } from '@stream-io/video-client';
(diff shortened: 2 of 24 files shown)


packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
santhoshvai, line 1: basic premise here is to not throw an error for keeping call alive purposes if notifee is installed.. 

basically in android < 26, user gets a log that notifee is not installed but wont cause the app to crash

packages/react-native-sdk/src/utils/push/android.ts
santhoshvai, line 1: basic premise here is to throw an error for push notifications if notifee is installed..

packages/react-native-sdk/src/utils/StreamVideoRN/types.ts
santhoshvai, line 1: I dont know if this would work.. I am using type of an optional  dependency

I will test and update here

packages/react-native-sdk/src/utils/push/libs/notifee.ts
```diff
@@ -0,0 +1,33 @@
...
+export function getNotifeeLibNoThrowForKeepCallAlive() {
+  if (!notifeeLib) {
+    const logger = getLogger(['getNotifeeLibNoThrow']);
+    logger(
+      'info',
+      `${'@notifee/react-native library not installed. It is required to keep call alive in the background for Android < 26. '}${INSTALLATION_INSTRUCTION}`
+    );
+  }
+  return notifeeLib;
+}
```
oliverlaz, line 33: Shall we merge these methods to something more generic?
```
export function getNotifeeLib({ required = false }) { }
```

oliverlaz (approved):

packages/react-native-sdk/src/utils/push/libs/notifee.ts
santhoshvai: @oliverlaz I separated it for the clear log messages. Otherwise yeah the generic one makes sense. 

Also this method is not exposed to integrators and heance internal to the SDK. So the more explicit the better maintenance in my view.

packages/react-native-sdk/docusaurus/docs/reactnative/06-advanced/08-screensharing/02-setup/01-react-native.mdx
```diff
@@ -19,7 +19,6 @@ To be able to use the foreground service, the permission must be declared in the
 <uses-permission android:name="android.permission.POST_NOTIFICATIONS" />
 <!-- We declare the permissions to for using foreground service -->
 <uses-permission android:name="android.permission.FOREGROUND_SERVICE" />
-<uses-permission android:name="android.permission.FOREGROUND_SERVICE_MICROPHONE" />
```
khushal87, line 22: Just a double-check. This isn't needed anymore?

khushal87 (approved):

packages/react-native-sdk/docusaurus/docs/reactnative/06-advanced/08-screensharing/02-setup/01-react-native.mdx
santhoshvai: yes it isnt

PR #1461, opened 2024-08-13
create a instance getter - wip

santhoshvai:


packages/client/src/StreamVideoClient.ts
```diff
@@ -120,13 +122,51 @@ export class StreamVideoClient {
...
           this.logger('error', 'Failed to connect', err);
         });
       }
     }
   }
 
+  public static getOrCreateInstance(args: {
+    apiKey: string;
+    user: User;
+    tokenProvider: TokenProvider;
```
santhoshvai, line 145: I made user and tokenProvider to be required here.. that makes it simpler


client has to be tied to user-id so User has to be required

alternative to not making tokenProvider required is strange to me as User always ties with token when we connect the user

packages/react-native-sdk/src/utils/push/android.ts
santhoshvai, line 1: unrelevant, but I noticed that we could avoid an API call here

packages/client/src/devices/BrowserPermission.ts
santhoshvai, line 1: irrelevant, but annoying log on RN

packages/client/src/coordinator/connection/token_manager.ts
santhoshvai, line 1: we dont use secrets anymore, its moved to node-sdk

oliverlaz: Our Chat QA tests still use this form to subscribe to WS events. But it is ok, SDK integrators shouldn't be using this.

packages/client/src/StreamVideoClient.ts
```diff
@@ -120,13 +122,62 @@ export class StreamVideoClient {
...
       if (user) {
+        let id = user.id;
+        if (user.type === 'anonymous') {
+          id = '!anon';
+        }
+        if (id) {
+          if (StreamVideoClient._instanceMap.has(id)) {
+            this.logger(
+              'warn',
+              `A StreamVideoClient already exists for ${user.type === 'anonymous' ? 'an anyonymous user' : 'the given user ID'}; Prefer using getOrCreateInstance method`,
```
oliverlaz, line 133:
```suggestion
              `A StreamVideoClient already exists for ${user.type === 'anonymous' ? 'an anyonymous user' : id}; Prefer using getOrCreateInstance method`,
```

packages/client/src/StreamVideoClient.ts
```diff
@@ -120,13 +122,63 @@ export class StreamVideoClient {
...
+      if (args.user.type !== 'anonymous' && args.user.type !== 'guest') {
+        throw new Error(
+          'TokenProvider or token is required for a user that is not a guest or anonymous',
+        );
+      }
+    }
+    let instance = StreamVideoClient._instanceMap.get(user.id);
+    if (!instance) {
+      instance = new StreamVideoClient({ ...args, user });
+      StreamVideoClient._instanceMap.set(user.id, instance);
```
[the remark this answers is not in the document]
santhoshvai: good point i ll remove the `set` here

packages/client/src/StreamVideoClient.ts
```diff
@@ -120,13 +122,63 @@ export class StreamVideoClient {
...
+    } else {
+      if (instance.streamClient.key !== args.apiKey) {
+        throw new Error(
+          'It is not supported to have a video client with different API key for an existing user',
+        );
+      }
```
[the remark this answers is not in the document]
santhoshvai: discussed now, will remove it

oliverlaz (approved):

PR #1468, opened 2024-08-27
docs(react-native): broken links in documentation

isekovanic:
This PR fixes a bunch of broken links within the React Native SDK documentation.

As a follow-up to this, we should consolidate the page structure to be the same as for the other SDKs (the `react` one for example). That will come in a followup PR.

### DIFF
--- a/packages/react-native-sdk/docusaurus/docs/reactnative/03-core/02-joining-creating-calls.mdx
+++ b/packages/react-native-sdk/docusaurus/docs/reactnative/03-core/02-joining-creating-calls.mdx
@@ -14,7 +14,7 @@ This guide shows how to create, join, leave, and end calls and ring calls.
 
 You can create a call by specifying its `callType` and `callId`:
 
-The [Call Type](../../core/call-types) controls which features are enabled, and sets up permissions.
+The [Call Type](../../core/configuring-call-types) controls which features are enabled, and sets up permissions.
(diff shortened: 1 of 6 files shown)


packages/react-native-sdk/docusaurus/docs/reactnative/03-core/02-joining-creating-calls.mdx
```diff
@@ -14,7 +14,7 @@ This guide shows how to create, join, leave, and end calls and ring calls.
 
 You can create a call by specifying its `callType` and `callId`:
 
-The [Call Type](../../guides/configuring-call-types) controls which features are enabled, and sets up permissions.
```
isekovanic, line 17: `configuring-call-types` does not exist as a file, I assumed this was referring to `call-types` in general.

isekovanic: Ah, I just noticed the `id` at the top of every `mdx` file, I assume that's what links these ?

oliverlaz: this is a shared page, it lives in our docusaurus repo: https://github.com/GetStream/stream-chat-docusaurus-cli/blob/production/shared/video/_call-types.mdx

Feel free to take the link from the React docs.

PR #1472, opened 2024-08-30
chore: update node-sdk, add this to pronto for token generation

szuperaz:
I wanted to do this for a long time now, this way we can catch any issues with token creation in the node-sdk

### DIFF
--- a/sample-apps/react/react-dogfood/helpers/jwt.ts
+++ b/sample-apps/react/react-dogfood/helpers/jwt.ts
@@ -1,4 +1,5 @@
-import { decodeBase64, JWTUserToken } from 'stream-chat';
+import { StreamClient } from '@stream-io/node-sdk';
+import { decodeBase64 } from 'stream-chat';
 
 /**
  * The maximum validity of a token, in seconds.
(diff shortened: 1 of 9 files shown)


sample-apps/react/react-dogfood/helpers/jwt.ts
```diff
@@ -9,34 +10,29 @@ export const maxTokenValidityInSeconds: number =
...
-  const payload: Record<string, unknown> = {
+  return client.generateUserToken({
     iss: 'https://pronto.getstream.io',
     sub: `user/${userId}`,
     // subtract 5 seconds, sometimes the coordinator fails with
     // "token used before issued at (iat)" error
     iat: Math.round(Date.now() / 1000) - 5,
     ...rest,
     user_id: userId,
-    exp: expiration,
```
oliverlaz, line 36: I think `exp` is the JWT standard field for token validity check. Is this something that has changed on our backend too?

szuperaz: No, but the node-sdk will set `exp` from `validity_in_seconds`, this is there to make token generator easier for integrators, no need to deal with timestamps

oliverlaz: ah, I see! Thanks!

oliverlaz (approved):

PR #1479, opened 2024-09-06
fix: beep only when participantCount is less than five

oliverlaz:
### Overview

Beeping is uncomfortable in large calls when many people are joining/leaving the call.
This PR enables beeping only when the participant count is less than 5.


szuperaz (approved):
haha, I love that our official term for this is `beep`, I never noticed

PR #1493, opened 2024-09-20
fix: overridable bitrate and bitrate downscale factor

oliverlaz:
### Overview

Adds experimental APIs for overriding codec, bitrate, and bitrate downscale factor.
Currently, we keep this API internal and undocumented until we get more feedback and more insights from our testing.

In Pronto, these parameters can be overriden via a query parameter:

`?video_codec=h264&bitrate=1200000&bitrate_factor=3` 
- will use h264, with 1.2mbps bitrate, and every lower layer will use 3 times less bitrate than the previous one.

### DIFF
--- a/sample-apps/react/react-dogfood/components/MeetingUI.tsx
+++ b/sample-apps/react/react-dogfood/components/MeetingUI.tsx
@@ -49,6 +49,10 @@ export const MeetingUI = ({ chatClient, mode }: MeetingUIProps) => {
 
   const isProntoEnvironment = useIsProntoEnvironment();
   const videoCodecOverride = router.query['video_codec'] as string | undefined;
+  const bitrateOverride = router.query['bitrate'] as string | undefined;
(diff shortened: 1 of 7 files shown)


sample-apps/react/react-dogfood/components/MeetingUI.tsx
```diff
@@ -59,15 +63,22 @@ export const MeetingUI = ({ chatClient, mode }: MeetingUIProps) => {
         const preferredCodec = videoCodecOverride || prontoDefaultCodec;
 
         const videoSettings = call.state.settings?.video;
-        const frameHeight = videoSettings?.target_resolution.height || 1080;
+        const frameHeight =
+          call.camera.getCaptureResolution()?.width ??
```
myandrienko, line 67: Height?

oliverlaz: true :D

myandrienko (approved):

PR #1495, opened 2024-09-24
fix: do not always error out api calls when web socket initially failed

santhoshvai:
## Problem

Currently for all axios requests, we wait for `connectionIdPromise`. This promise is rejected or resolved based on the first WS connect attempt. If subsequent reconnects were successful but the first connect was rejected, this promise would still stay rejected. 

## Solution

To allow for reconnects to resolve, say for example, in 3rd try the connection got open, we can wait for the `connectionOpen` promise to resolve. To allow for 2 or 3 retries, we use `_waitForHealthy` function here. Alternatively we  could await only on `connectionOpen` promise without allowing for timeouts to wait for reconnections also.

### DIFF
--- a/packages/client/src/coordinator/connection/client.ts
+++ b/packages/client/src/coordinator/connection/client.ts
@@ -365,12 +365,7 @@ export class StreamClient {
       return Promise.resolve();
     }
 
-    this.connectionIdPromise = new Promise<string | undefined>(
-      (resolve, reject) => {
-        this.resolveConnectionId = resolve;
-        this.rejectConnectionId = reject;
-      },
-    );
+    this._setupConnectionIdPromise();
 
     this.clientID = `${this.userID}--${randomId()}`;
     this.wsPromise = this.connect();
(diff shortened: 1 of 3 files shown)


packages/client/src/coordinator/connection/client.ts
```diff
@@ -540,8 +540,15 @@ export class StreamClient {
       await Promise.all([
         this.tokenManager.tokenReady(),
         this.guestUserCreatePromise,
-        this.connectionIdPromise,
       ]);
+      try {
+        await this.connectionIdPromise;
```
myandrienko, line 545: In this case, do we even need `connectionIdPromise`? Can we just always do `waitForHealthy` instead?

myandrienko (approved):

szuperaz (approved):

packages/client/src/coordinator/connection/client.ts
santhoshvai: good point here, I think we dont need it too, I ll remove it

santhoshvai: @myandrienko not really for anonymous users we have this code inside the client

```
    // some endpoints require a connection_id to be resolved.
    // as anonymous users aren't allowed to open WS connections, we just
    // resolve the connection_id here.
    this.resolveConnectionId?.();
```

so need to keep this

PR #1499, opened 2024-09-26
feat: improvements for the theming scheme of the SDK

kristian-mkd:
## 🎯 Goal

- consolidate and improve the theming options of the SDK

## 🛠 Implementation details

- theme colors names and options are improved
- new default variants added (fontSizes, spacingSizes)
- alternative colorPalette theming option is deleted

## 🎨 UI Changes

<details>
<summary>iOS</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

<details>
<summary>Android</summary>

<table>
    <thead>
        <tr>
            <td>Before</td>
            <td>After</td>
        </tr>
    </thead>
    <tbody>
        <tr>
            <td>
                
            </td>
            <td>
                
            </td>
        </tr>
    </tbody>
</table>
</details>

## ☑️ Checklist

- [ ] I have signed the [Stream CLA](https://docs.google.com/forms/d/e/1FAIpQLScFKsKkAJI7mhCr7K9rEIOpqIDThrWxuvxnwUq2XkHyG154vQ/viewform) (required)
- [ ] PR targets the `develop` branch
- [ ] Documentation is updated
- [ ] New code is tested in main example apps, including all possible scenarios
  - [ ] SampleApp iOS and Android
  - [ ] Expo iOS and Android

### DIFF
--- a/packages/react-native-sdk/src/components/Call/CallControls/CallControlsButton.tsx
+++ b/packages/react-native-sdk/src/components/Call/CallControls/CallControlsButton.tsx
@@ -7,7 +7,7 @@ import {
   View,
 } from 'react-native';
 import { useTheme } from '../../../contexts/ThemeContext';
-import { Theme } from '../../../theme/theme';
+import { Theme, defaultTheme } from '../../../theme/theme';
 
 interface CallControlsButtonProps {
   /**
@@ -73,7 +73,7 @@ export const CallControlsButton = (
       height: size || buttonSizes.sm,
       width: size || buttonSizes.sm,
       borderRadius: (size || buttonSizes.sm) / 2,
-      borderColor: colors.content_bg,
+      borderColor: defaultTheme.darkColors.border,
     },
     styleProp?.container ?? null,
     container,
--- a/packages/react-native-sdk/src/components/Call/CallControls/ToggleAudioPublishingButton.tsx
+++ b/packages/react-native-sdk/src/components/Call/CallControls/ToggleAudioPublishingButton.tsx
@@ -41,7 +41,7 @@ export const ToggleAudioPublishingButton = ({
     <Restricted requiredGrants={[OwnCapability.SEND_AUDIO]}>
       <CallControlsButton
         onPress={onPress}
-        color={!optimisticIsMute ? colors.static_white : colors.overlay_dark}
+        color={!optimisticIsMute ? colors.static_white : colors.overlay_medium}
         style={toggleAudioPublishingButton}
       >
         {!optimisticIsMute ? (
--- a/packages/react-native-sdk/src/components/Call/CallControls/ToggleCameraFaceButton.tsx
+++ b/packages/react-native-sdk/src/components/Call/CallControls/ToggleCameraFaceButton.tsx
@@ -47,7 +47,9 @@ export const ToggleCameraFaceButton = ({
     <Restricted requiredGrants={[OwnCapability.SEND_VIDEO]}>
       <CallControlsButton
         onPress={onPress}
-        color={direction === 'back' ? colors.overlay_dark : colors.static_white}
+        color={
+          direction === 'back' ? colors.overlay_medium : colors.static_white
+        }
         disabled={optimisticIsMute}
         style={toggleCameraFaceButton}
       >
--- a/packages/react-native-sdk/src/components/Call/CallControls/internal/ReactionsPicker.tsx
+++ b/packages/react-native-sdk/src/components/Call/CallControls/internal/ReactionsPicker.tsx
@@ -12,14 +12,15 @@ import { ComponentTestIds } from '../../../../constants/TestIds';
 import { useTheme } from '../../../../contexts/ThemeContext';
 import { ReactionsButtonProps } from '../ReactionsButton';
 import { defaultEmojiReactions } from '../../../../constants';
+import { defaultTheme } from '../../../..';
 
 type ReactionPickerProps = Pick<ReactionsButtonProps, 'supportedReactions'> & {
   reactionsButtonLayoutRectangle?: LayoutRectangle;
   onRequestedClose: () => void;
 };
 
-const TOP_PADDING = 4;
-const REACTION_MARGIN_BOTTOM = 4;
+const TOP_PADDING = defaultTheme.variants.spacingSizes.xs;
+const REACTION_MARGIN_BOTTOM = defaultTheme.variants.spacingSizes.xs;
 
 export const ReactionsPicker = ({
   supportedReactions = defaultEmojiReactions,
@@ -121,7 +122,7 @@ export const ReactionsPicker = ({
               reactionItemStyle,
               {
                 // temporary background color until we have theming
-                backgroundColor: colors.overlay,
+                backgroundColor: colors.overlay_light,
               },
               reactionsPicker.reactionItem,
             ]}
--- a/packages/react-native-sdk/src/components/Livestream/LivestreamControls/LivestreamAudioControlButton.tsx
+++ b/packages/react-native-sdk/src/components/Livestream/LivestreamControls/LivestreamAudioControlButton.tsx
@@ -28,7 +28,7 @@ export const LivestreamAudioControlButton = () => {
       style={[
         styles.container,
         {
-          backgroundColor: colors.dark_gray,
+          backgroundColor: colors.background,
           height: buttonSizes.xs,
           width: buttonSizes.xs,
         },
--- a/packages/react-native-sdk/src/theme/colors.ts
+++ b/packages/react-native-sdk/src/theme/colors.ts
@@ -10,40 +10,42 @@ const colors = {
   primary: palette.blue500,
   error: palette.red400,
   info: palette.green500,
-  static_black: palette.grey950,
-  static_white: palette.grey50,
-  static_overlay: palette.grey950 + opacityToHex(0.85),
-  static_grey: palette.grey700,
   disabled: palette.grey600,
+
+  background: palette.grey800,
+  border: palette.grey300,
+  bar: palette.grey50,
   text_low_emphasis: palette.grey500,
   text_high_emphasis: palette.grey950,
-  controls_bg: palette.grey50,
-  borders: palette.grey300,
-  overlay: palette.grey950 + opacityToHex(0.4),
-  overlay_dark: palette.grey950 + opacityToHex(0.6),
-  bars: palette.grey50,
-  content_bg: palette.grey950 + opacityToHex(0.05),
-  dark_gray: palette.grey800,
+
+  overlay_light: palette.grey950 + opacityToHex(0.4),
+  overlay_medium: palette.grey950 + opacityToHex(0.6),
+  overlay_heavy: palette.grey950 + opacityToHex(0.85),
+
+  static_white: palette.grey50,
+  static_grey: palette.grey700,
+  static_black: palette.grey950,
 };
 
-const darkThemeColors = {
+const darkColors = {
   primary: palette.blue500,
   error: palette.red400,
   info: palette.green500,
-  static_black: palette.grey950,
-  static_white: palette.grey50,
-  static_overlay: palette.grey950 + opacityToHex(0.85),
-  static_grey: palette.grey700,
   disabled: palette.grey600,
+
+  background: palette.grey800,
+  border: palette.grey950 + opacityToHex(0.05),
+  bar: palette.grey900,
   text_low_emphasis: palette.grey500,
   text_high_emphasis: palette.grey50,
-  controls_bg: palette.grey900,
-  borders: palette.grey700,
-  overlay: palette.grey950 + opacityToHex(0.4),
-  overlay_dark: palette.grey50 + opacityToHex(0.6),
-  bars: palette.grey900,
-  content_bg: palette.grey950 + opacityToHex(0.05),
-  dark_gray: palette.grey800,
+
+  overlay_light: palette.grey950 + opacityToHex(0.4),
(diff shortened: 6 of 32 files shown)


packages/react-native-sdk/src/theme/theme.ts
```diff
@@ -585,4 +610,32 @@ export const defaultTheme: Theme = {
     buttonIcon: {},
     buttonText: {},
   },
+  getValue: function (component: string, prop: string) {
```
kristian-mkd, line 613: @santhoshvai @oliverlaz What do you think about this option having a `getValue` method that will check if there is value for given prop for the given component and if there isn't to just fallback to some theme defaults value. It's usage would be: `backgroundColor: theme.getValue('button', 'backgroundColor'),`.

sample-apps/react-native/dogfood/src/components/Button.tsx
```diff
@@ -24,6 +24,9 @@ export const Button = ({
   titleStyle,
   ...rest
 }: ButtonPropTypes) => {
+  const { theme } = useTheme();
+  const styles = getStyles(theme);
```
kristian-mkd, line 28: @santhoshvai what do you think of this approach passing the theme to `getStyles` so that all `appTheme` usages in `dogfood` app are replaced with proper usage of the theme so that we can more easily switch theme on app level dinamically? If this is ok, I can replace all `appTheme` usages in the `dogfood` app.

packages/react-native-sdk/src/theme/theme.ts
oliverlaz: It makes sense. I would even shorten the name of the method to `theme.get(path)`.

Also, we can support `theme.get('button.backgroundColor')`.

packages/react-native-sdk/src/theme/theme.ts
```diff
@@ -585,4 +610,32 @@ export const defaultTheme: Theme = {
...
+      case 'fontWeight':
+        return this.defaults.fontWeight;
+      case 'borderRadius':
+        return this.defaults.borderRadius;
+      case 'borderColor':
+        return this.defaults.borderColor;
+      case 'borderWidth':
+        return this.defaults.borderWidth;
+      default:
+        throw new Error(`Invalid component or prop: ${component}.${prop}`);
```
oliverlaz, line 638: I'm not sure if we should throw here. Missing theme property shouldn't crash the application. I think we should just log an error.

Also, as this is a forever-growing list, I think we can implement it like this:
```ts
if (this[component] && this[component][prop]) {
  return this[component][prop];
}

const defaultValue = this.defaults[prop];
if (!defaultValue) {
  // log
}
return defaultValue;
```

sample-apps/react-native/dogfood/src/components/Button.tsx
oliverlaz: Chiming-in here: I think it would be better if `getStyles` is turned into `useStyles()` that internally memoizes the returned stylesheet. Something like this:

```ts
const useStyles = () => {
  const theme = useTheme();
  return useMemo(() => {
    return StyleSheet.create({
      button: {
        //backgroundColor: theme.colors.primary,
        backgroundColor: theme.getValue('button', 'backgroundColor'),
        justifyContent: 'center',
        borderRadius: 8,
        height: BUTTON_HEIGHT,
        paddingHorizontal: theme.variants.spacingSizes.md,
      },
      buttonText: {
        color: theme.colors.static_white,
        fontWeight: theme.typefaces.heading6.fontWeight,
        textAlign: 'center',
        fontSize: 17,
      },
      disabledButtonStyle: {
        backgroundColor: theme.colors.disabled,
      },
    );
  }, [theme]);
}
```

This hook should be implemented for every component that uses these theme variables.

packages/react-native-sdk/src/components/Call/CallControls/CallControlsButton.tsx
```diff
@@ -73,7 +74,7 @@ export const CallControlsButton = (
       height: size || buttonSizes.sm,
       width: size || buttonSizes.sm,
       borderRadius: (size || buttonSizes.sm) / 2,
-      borderColor: colors.content_bg,
+      borderColor: colorPalette.dark.border,
```
oliverlaz, line 77: Please read this from `useTheme()`

packages/react-native-sdk/src/components/Call/CallControls/CallControlsButton.tsx
```diff
@@ -8,6 +8,7 @@ import {
 } from 'react-native';
 import { useTheme } from '../../../contexts/ThemeContext';
 import { Theme } from '../../../theme/theme';
+import { colorPalette } from '../../..';
```
oliverlaz, line 11: This should be an illegal import as it is source of theme bugs :)

sample-apps/react-native/dogfood/src/components/ParticipantActions.tsx
```diff
@@ -1,6 +1,6 @@
 import {
   Avatar,
-  colorPallet,
+  colorPalette,
```
oliverlaz, line 3: As this is anyway a breaking change (renamed 😅 ), can we remove this export completely? `useTheme()` should be the way to go here.

packages/react-native-sdk/src/components/Call/CallControls/CallControlsButton.tsx
santhoshvai: we should never import color palette inside our SDK components

But we should export colorPalette to integrators as they may want to use the default theme colors for their own components outside of `useTheme`

packages/react-native-sdk/src/components/Call/CallControls/ToggleCameraFaceButton.tsx
```diff
@@ -47,7 +47,9 @@ export const ToggleCameraFaceButton = ({
     <Restricted requiredGrants={[OwnCapability.SEND_VIDEO]}>
       <CallControlsButton
         onPress={onPress}
-        color={direction === 'back' ? colors.overlay_dark : colors.static_white}
+        color={
+          direction === 'back' ? colors.overlay_medium : colors.static_white
```
santhoshvai, line 51: we should avoid any strong color reference here like 

```
colors.static_white
```

rather it should be generic like `colors.cardBackground` etc

packages/react-native-sdk/src/components/Call/CallControls/internal/ReactionsPicker.tsx
```diff
@@ -12,14 +12,15 @@ import { ComponentTestIds } from '../../../../constants/TestIds';
...
 
 type ReactionPickerProps = Pick<ReactionsButtonProps, 'supportedReactions'> & {
   reactionsButtonLayoutRectangle?: LayoutRectangle;
   onRequestedClose: () => void;
 };
 
-const TOP_PADDING = 4;
-const REACTION_MARGIN_BOTTOM = 4;
+const TOP_PADDING = defaultTheme.variants.spacingSizes.xs;
+const REACTION_MARGIN_BOTTOM = defaultTheme.variants.spacingSizes.xs;
```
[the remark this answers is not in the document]
santhoshvai: +1

packages/react-native-sdk/src/components/Call/CallControls/ToggleAudioPublishingButton.tsx
```diff
@@ -41,7 +41,7 @@ export const ToggleAudioPublishingButton = ({
     <Restricted requiredGrants={[OwnCapability.SEND_AUDIO]}>
       <CallControlsButton
         onPress={onPress}
-        color={!optimisticIsMute ? colors.static_white : colors.overlay_dark}
+        color={!optimisticIsMute ? colors.static_white : colors.overlay_medium}
```
santhoshvai, line 44: Instead of `overlay_medium` can we make it like `secondaryBackground` ? good to check Figma here, maybe they use a naming scheme

packages/react-native-sdk/src/theme/colors.ts
```diff
@@ -6,44 +6,51 @@ const opacityToHex = (opacity: number) => {
...
-  content_bg: palette.grey950 + opacityToHex(0.05),
-  dark_gray: palette.grey800,
+
+  overlay_light: palette.grey950 + opacityToHex(0.4),
+  overlay_medium: palette.grey950 + opacityToHex(0.6),
+  overlay_heavy: palette.grey950 + opacityToHex(0.85),
+
+  static_white: palette.grey50,
+  static_grey: palette.grey700,
+  static_black: palette.grey950,
```
santhoshvai, line 48: All these should turn to generic names in my view, we should never be using a name called as `static_black` it should be something like `accent` or `primary` etc

packages/react-native-sdk/src/theme/theme.ts
santhoshvai: The function is fine,

but its super confusing right now that I should pass `getValue("", "color")` to get the generic values.. 

also, I have a larger question, is this function necessary? Its already easier in my view

```
const theme = useTheme()
const defaultBg = theme.defaults. backgroundColor
const buttonBg = theme.button.backgroundColor
```

sample-apps/react-native/dogfood/src/components/Button.tsx
santhoshvai: Side note: StyleSheet.create is for compile time type checking.Also you can look at Stylesheet.compose if you want to merge styles with compile time type checking.

packages/react-native-sdk/src/theme/colors.ts
oliverlaz: this is a good point. we don't use color names in the theme variables for React:
- https://github.com/GetStream/stream-video-js/blob/main/packages/styling/src/_global-theme-variables.scss

packages/react-native-sdk/src/theme/theme.ts
kristian-mkd: @santhoshvai I see this function as alternative helper to your approach, so I think we can add it as additional way to read from theme and fallback to defaults.

packages/react-native-sdk/src/theme/colors.ts
kristian-mkd: Yes, you are right, we should not mention color names in the theme so I am renaming them like this:
```ts
  // old static_(white, gray, black)
  base1: palette.grey50,
  base2: palette.grey300,
  base3: palette.grey700,
  base4: palette.grey950,

  // old overlay_* or similar less frequently and wrongly used bar, border etc colors
  background1: palette.grey950 + opacityToHex(0.05),
  background2: palette.grey800,
  background3: palette.grey950 + opacityToHex(0.4), 
  background4: palette.grey950 + opacityToHex(0.6), 
  background5: palette.grey900,
  background6: palette.grey950 + opacityToHex(0.85),
```

packages/react-native-sdk/src/components/Livestream/LivestreamControls/LivestreamAudioControlButton.tsx
```diff
@@ -46,9 +46,9 @@ export const LivestreamAudioControlButton = () => {
         ]}
       >
         {!optimisticIsMute ? (
-          <Mic color={colors.static_white} />
```
kristian-mkd, line 49: Changing this on purpose with different color as in the other use cases with Mic on/off.

packages/react-native-sdk/src/contexts/ThemeContext.tsx
```diff
@@ -29,11 +29,10 @@ export type ThemeContextValue = {
 
 export const mergeThemes = (params: MergedThemesParams) => {
   const { style, theme } = params;
-  const finalTheme = (
-    !theme || Object.keys(theme).length === 0
-      ? JSON.parse(JSON.stringify(defaultTheme))
```
kristian-mkd, line 34: Had to switch to `Object.assign()` because the old solution was dropping the methods defined in the theme object. @khushal87 thanks again for helping me debug this. 🚀

oliverlaz: hey @kristian-mkd, keep in mind that `Object.assign` does a shallow merge. Perhaps you want to use `lodash/deepMerge`.

kristian-mkd: I think I can use `lodash/deepClone` first and then `lodash/merge`, something like this:
```ts
  const finalTheme = isThemeEmpty ? cloneDeep(defaultTheme) : cloneDeep(theme);
  if (style) {
    merge(finalTheme, style);
  }
```

packages/react-native-sdk/src/theme/theme.ts
```diff
@@ -335,6 +361,17 @@ export const defaultTheme: Theme = {
       fontWeight: '400',
     },
   },
+  defaults: {
+    color: colors.primary,
+    backgroundColor: colors.background2,
+    margin: 10,
+    padding: 10,
+    fontSize: 16,
```
santhoshvai, line 369: Please do make sure that all margins, padding, sizes also move to theme when we move to design v2

packages/react-native-sdk/src/contexts/ThemeContext.tsx
santhoshvai: this change is also necessary when you remove the method

santhoshvai (approved):

PR #1506, opened 2024-10-03
feat: add a prop to control mirroring of local participant video

myandrienko:
We get a lot of questions about disabling mirroring of local participant video. It's currently possible to do by overriding the `str-video__video--mirror` CSS class, but it seems that things will be easier if we have a prop to control this behavior.

This PR adds an optional `mirrorLocalParticipantVideo` prop to all built-in layout components. Setting it to `false` disable local participant video mirroring.

### DIFF
--- a/packages/react-sdk/docusaurus/docs/React/03-ui-components/core/call-layout.mdx
+++ b/packages/react-sdk/docusaurus/docs/React/03-ui-components/core/call-layout.mdx
@@ -77,26 +77,28 @@ const MyApp = () => {
 
 #### Props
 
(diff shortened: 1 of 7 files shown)


packages/react-sdk/docusaurus/docs/React/03-ui-components/core/call-layout.mdx
oliverlaz, line 1: While you are here, could you please add a page for the rest of the Core components (Video, ParticipantAudio, etc...)?

oliverlaz (approved):

PR #1508, opened 2024-10-03
feat(closed captions): Integration in the SDKs

myandrienko:
### Overview

Implements the spec: https://www.notion.so/stream-wiki/Closed-Captions-Spec-and-UI-Cookbook-10b6a5d7f9f68083a6a8cc255d93dac5

### Pronto

Adds status indicators for closed captions and transcription:

Screenshot 2024-12-19 at 17 54 23


kristian-mkd (approved):
@oliverlaz's closed captions changes and cookbook, LGTM 🚀

packages/client/src/store/CallState.ts
```diff
@@ -1299,4 +1393,50 @@ export class CallState {
       this.setCurrentValue(this.ownCapabilitiesSubject, event.own_capabilities);
     }
   };
+
+  private updateClosedCaptions = (event: ClosedCaptionEvent) => {
```
myandrienko, line 1397: Suggestion: some event handlers we keep outside of the state class, in `src/events`. Since CallState is getting rather big, should we consider extracting this handler there as well?

packages/react-native-sdk/docusaurus/docs/reactnative/05-ui-cookbook/21-closed-captions.mdx
```diff
@@ -0,0 +1,103 @@
...
+```tsx
+import { useCallStateHooks } from '@stream-io/video-react-native-sdk';
+
+export const ClosedCaptions = () => {
+  const { useCallClosedCaptions } = useCallStateHooks();
+  const closedCaptions = useCallClosedCaptions();
+  return (
+    <View style={styles.rootContainer}>
+      {closedCaptions.map(({ speaker_name, text }) => (
+        <View style={styles.closedCaptionItem} key={`${speaker_name}/${text}`}>
```
myandrienko, line 75: Let's use speaker name + start time as key, since that's what we deduplicate on. Otherwise this key may not be unique.

oliverlaz: Fixed with: [6901fda](https://github.com/GetStream/stream-video-js/pull/1508/commits/6901fda8bbe93dcbe934902eeca9a6b3eb625a67)

packages/client/src/store/CallState.ts
oliverlaz: It used to be like that in the past. We have moved to a single class-does-it-all approach to be consistent with the other SDKs. So, at this moment, I wouldn't be breaking this class into pieces :)

oliverlaz (approved):

PR #1523, opened 2024-10-17
feat: add methods to query default device

myandrienko:


myandrienko:
Closed in favor of #1541.

PR #1528, opened 2024-10-18
feat: add call duration in the call state

kristian-mkd:
Adds `duration` to call state used for tracking time duration since the call started. 
Useful for implementation of timers in the video call screens (ex. Call Duration Badge component in client aplication.


packages/client/src/store/CallState.ts
```diff
@@ -390,6 +396,10 @@ export class CallState {
     this.participantCount$ = duc(this.participantCountSubject);
     this.recording$ = duc(this.recordingSubject);
     this.transcribing$ = duc(this.transcribingSubject);
+    this.duration$ = duc(this.durationSubject);
+    setInterval(() => {
+      this.setDuration((d) => d + 1);
+    }, 1000);
```
myandrienko, line 402: If we set an interval like this, we need to also clear it somewhere, otherwise it's a memory leak.

You can also consider setting interval in Rx (with `timer` or `interval`) instead.

kristian-mkd: Good catch about the memory leak 👍 I wasn't able to find good place to unsubscribe to this `setInterval` (no such thing as destructors in js) so I opted for different solution, moving this out of the call state.

Also the new hook takes `callStarted` timestamp as well, so on rejoining call or other future call joiners will see the proper time duration of a call.

@oliverlaz @myandrienko can we proceed with this solution or do you have a better idea?

oliverlaz: @kristian-mkd, this should live in `callState` as otherwise, it would be unavailable for Plain-JS.
In our CC PR, I've introduced `CallState.#dispose()` method, you can add it here too, and I'll resolve the conflicts.
- https://github.com/GetStream/stream-video-js/pull/1508/files#diff-e07349e380403998673da01d8b5dc1ebc2e7955bd23a6cf82409c21b02d8dda9R494-R503

Additionally, the interval should start automatically when
- joined (init with `session.live_started_at`)
- call.session_started event is emitted
- call.session_ended should stop it

kristian-mkd: @oliverlaz I have pushed the suggested changes. Please verify them, if all good I will update the docs.

packages/react-native-sdk/docusaurus/docs/reactnative/03-core/03-call-and-participant-state.mdx
```diff
@@ -69,6 +69,7 @@ Here is an excerpt of the available call state hooks:
 | `useCallCreatedAt`                   | The time the call was created.                                                                                |
 | `useCallCreatedBy`                   | The user that created the call.                                                                               |
 | `useCallCustomData`                  | The custom data attached to the call.                                                                         |
+| `useCallDuration`                    | The call duration since the start of the call in seconds.                                                     |
```
oliverlaz, line 72: It would be good to add this to the React docs too. I know it is annoying to have the same content spread in three places :)

sample-apps/react/react-dogfood/components/ActiveCallHeader.tsx
```diff
@@ -38,26 +38,24 @@ const LatencyIndicator = () => {
...
-      const hours = format.substring(11, 13);
-      const minutes = format.substring(14, 16);
-      const seconds = format.substring(17, 19);
-      const time = `${hours !== '00' ? hours + ':' : ''}${minutes}:${seconds}`;
-      setElapsed(time);
-    }, 1000);
-    return () => clearInterval(interval);
-  }, [startedAtDate]);
+    setElapsed(formatTime(duration));
+  }, [duration]);
```
oliverlaz, line 58: I think we can remove the `useEffect` and `useState` here, and apply the formatting directly:

```ts
const elapsed = formatTime(duration);
```

oliverlaz:
Closing stale PR.

PR #1533, opened 2024-10-23
feat: report input devices in call stats

oliverlaz:
### Overview

Adds input device reporting in the call stats.

### DIFF
--- a/packages/client/src/stats/SfuStatsReporter.ts
+++ b/packages/client/src/stats/SfuStatsReporter.ts
@@ -142,6 +142,7 @@ export class SfuStatsReporter {
   stop = () => {
     this.subscriptions.forEach((unsubscribe) => unsubscribe());
     this.subscriptions = [];
+    this.inputDevices.clear();
     clearInterval(this.intervalId);
     this.intervalId = undefined;
   };
--- a/packages/client/src/stats/SfuStatsReporter.ts
+++ b/packages/client/src/stats/SfuStatsReporter.ts
@@ -5,10 +5,7 @@ import { getLogger } from '../logger';
 import { Publisher, Subscriber } from '../rtc';
 import { flatten, getSdkName, getSdkVersion } from './utils';
 import { getWebRTCInfo, LocalClientDetailsType } from '../client-details';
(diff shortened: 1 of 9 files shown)


packages/client/src/stats/SfuStatsReporter.ts
```diff
@@ -20,34 +28,86 @@ export class SfuStatsReporter {
...
+            : ownCapabilities.includes(OwnCapability.SEND_VIDEO);
+        if (!hasPermission || !hasCapability) {
+          this.inputDevices.set(kind, {
+            currentDevice: '',
+            availableDevices: [],
+            isPermitted: false,
+          });
+          return;
+        }
+        const listDevicesSubscription = createSubscription(
```
myandrienko, line 94: I'm a bit worried about creating a subscription inside a subscription, especially since we won't clean up "inner" subscriptions until `stop` is called.

Maybe we can use something like `iif` to subscribe to `listDevices()` only if we have both permission and capabilities?

packages/client/src/stats/SfuStatsReporter.ts
```diff
@@ -60,10 +120,16 @@ export class SfuStatsReporter {
       webrtcVersion: this.webRTCVersion,
       subscriberStats,
       publisherStats,
+      audioDevices: this.inputDevices.get('mic'),
+      videoDevices: this.inputDevices.get('camera'),
+      deviceState: { oneofKind: undefined },
```
myandrienko, line 125: I've tried using Compute Pressure API in Chrome, I'll push this to your branch :)

packages/client/src/stats/SfuStatsReporter.ts
oliverlaz: yep, it can happen. rarely, but it can still happen. let me try to fix this.

oliverlaz: @myandrienko please check https://github.com/GetStream/stream-video-js/pull/1533/commits/b2b6e6cee9ffc066524c7df3969c7623f872ac66

PR #1535, opened 2024-10-25
feat(design-v2): call recording flow

kristian-mkd:
This PR:
- completes the Call recording flow
- adds the Recording modal component
- moves the `useToggleCallRecording` to `useCallStateHooks` so it can be used for React Native as well

ios-after

ios-after

ios-after


packages/react-bindings/src/hooks/callStateHooks.ts
```diff
@@ -479,3 +479,43 @@ export const useIncomingVideoSettings = () => {
...
 };
+
+/**
+ * Custom hook for toggling call recording in a video call.
+ *
+ * This hook provides functionality to start and stop call recording,
+ * along with state management for tracking the recording status
+ * and the loading indicator while awaiting a response.
+ */
+export const useToggleCallRecording = () => {
```
oliverlaz, line 490: I like the idea but let's move these utility hooks to some other file. This hook isn't a `callState` hook.

oliverlaz: I think there are a few more utility hooks like this in both SDKs that can be moved to bindings.

kristian-mkd: I have created new file in the hooks folder named `callUtilsHooks.ts` and moved the `useToggleCallRecording` hook there.

PR #1541, opened 2024-10-29
fix: various device selector issues

myandrienko:
This PR fixes a bunch of device selector issues.

## Invalid default device indicator

In some of our device selector components, we assumed that the first device in the list was the default. This isn't always true. So now, if we don't know what the default device is, we don't mark any device as selected, which is more "honest" behavior.

image

This PR also ensures that device selectors always fall back to the "Default" option if the current device is not found in the device list (e.g. because of stale persisted device preferences) or the device list is empty (e.g. if there's no browser permission).

image

image

## Older Firefox cannot query permissions

Up until version 131 (incl.) there was no way to query camera and microphone permissions with `navigator.permissions.query`. We now assume that permission status is `prompt` in this case, and just try getting user media. If it works, permission is granted, if not - it’s denied.

## Firefox doesn't show device labels without `getUserMedia`

When enumerating devices, Firefox will hide device labels unless there’s been an active user media stream on the page. So even when camera and microphone permissions are `granted`, device labels are still hidden until there’s been a successful call to `getUserMedia`. Now we force device list updates after every successful `getUserMedia` call.

## Safari doesn't fire permission change events reliably

Safari doesn't fire camera and microphone permission change events for reliably. For example, the event doesn't fire if the user denies camera/microphone access in the initial prompt. As a result, we didn't reliably display missing permission indicators in Safari.

We now update browser permission state not only from permission change events, but also based on `NotAllowedError` and successful attempts to obtain the user's media stream.

### DIFF
--- a/packages/client/src/devices/devices.ts
+++ b/packages/client/src/devices/devices.ts
@@ -32,7 +32,9 @@ const getDevices = (permission: BrowserPermission, kind: MediaDeviceKind) => {
         await permission.prompt({ throwOnNotAllowed: true });
         devices = await navigator.mediaDevices.enumerateDevices();
       }
-      return devices.filter((d) => d.kind === kind);
+      return devices.filter(
+        (device) => device.kind === kind && device.label !== '',
+      );
     })(),
   );
 };
@@ -125,7 +127,7 @@ export const getAudioDevices = lazy(() => {
  * if devices are added/removed the list is updated, and if the permission is revoked,
  * the observable errors.
  */
-export const getVideoDevices = () => {
+export const getVideoDevices = lazy(() => {
   return merge(
     getDeviceChangeObserver(),
     getVideoBrowserPermission().asObservable(),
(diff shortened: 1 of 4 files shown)


packages/client/src/devices/devices.ts
```diff
@@ -151,10 +156,17 @@ export const getAudioOutputDevices = () => {
...
 
 const getStream = async (constraints: MediaStreamConstraints) => {
-  return await navigator.mediaDevices.getUserMedia(constraints);
+  const stream = await navigator.mediaDevices.getUserMedia(constraints);
+  // On Firefox, there's no reliable way to listen to device permission changes,
+  // so it's difficult to update the device list once camera or microphone access
+  // is allowed by the user. However, if we were able to get user media, we can be
+  // sure permission was granted, so we fake the devicechange event to force device
+  // lists to update.
+  navigator.mediaDevices.dispatchEvent(new Event('devicechange'));
```
oliverlaz, line 168: Shall we limit this dispatch to Firefox only?

oliverlaz (approved):

packages/client/src/devices/devices.ts
myandrienko: Done in [8ba1e8c](https://github.com/GetStream/stream-video-js/pull/1541/commits/8ba1e8ce11a2b6a00330f2c6dc257cd6985c7576). I also updated the comment, since it was misleading about the reasons to force this event.

PR #1542, opened 2024-10-29
feat: make push notification listeners more flexible for easier debugging

### LINKED ISSUE #1447: React native - StreamVideoRN.setPushConfig() setBackgroundMessageHandler conflict

**Describe the bug**
In my app I use react-native-firebase to handle notifications logic 
in my index.js  I declare setBackgroundMessageHandler to handle background notification logic 
But when I initialize StreamVideoRN.setPushConfig() it also creates its own setBackgroundMessageHandler, so my previous setBackgroundMessageHandler index.js logic does not work anymore.
Is it possible to preserve my original logic in setBackgroundMessageHandler and also use StreamVideoRN setBackgroundMessageHandler?

**Version**

- SDK version: 0.9.2

**Env:**
Production

**React Native**

- OS: android, ios
- emulator and real device

santhoshvai:
- [x] Removed notification listeners inside the SDK, but export them as utility functions
- [x] Documentation guides how to add own listeners using the utility functions. This fixes #1447
- [x] Dogfood apps are updated
- [x] Added doc on how to add incoming/outgoing call UIs on app root level
- [x] throw error when deprecated navigation functions for ringing calls are used
- [x] add section about how to add sound to ringing calls on foreground
- [ ] make a 1.3 release


packages/react-native-sdk/src/utils/StreamVideoRN/index.ts
```diff
@@ -67,10 +65,6 @@ export class StreamVideoRN {
       return;
     }
     this.config.push = pushConfig;
-    // After getting the config we should setup callkeep events, firebase handler asap to handle incoming calls from a dead state
-    setupFirebaseHandlerAndroid(pushConfig);
-    // setup ios handler for non-voip push notifications asap
-    setupRemoteNotificationsHandleriOS(pushConfig);
```
santhoshvai, line 73: main effect is here..

SDK will not automatically add listeners, integrators need to add them

oliverlaz (approved):

PR #1546, opened 2024-10-31
fix: camera not enabled on foreground notifications

santhoshvai:


packages/client/src/devices/InputMediaDeviceManager.ts
santhoshvai, line 1: this is the fix in client for resume not working if disable was called before enable completed

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -78,6 +78,7 @@ export abstract class InputMediaDeviceManager<
       return;
     }
 
+    this.state.prevStatus = this.state.status;
```
santhoshvai, line 81: this change is not really necessary, but spotted that prevState was never set when doing enable()

we dont have a usecase for prevState being disable and currentstate being enable yet

oliverlaz (approved):
looks good to me, but I'd like @myandrienko to also take a look.

packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -74,10 +74,10 @@ export abstract class InputMediaDeviceManager<
    * Starts stream.
    */
   async enable() {
+    this.state.prevStatus = this.state.status;
```
myandrienko, line 77: I think, you can replace that with `this.state.prevStatus = this.state.optmisticStatus`. Then the check in lines 101-108 will not be required.

I'm not 100% sure though, and the PR LGTM anyway.

myandrienko (approved):

PR #1554, opened 2024-11-04
fix: camera flip did not work in react-native

santhoshvai:
fixes regression caused by #1521

### DIFF
--- a/packages/client/src/devices/CameraManager.ts
+++ b/packages/client/src/devices/CameraManager.ts
@@ -8,8 +8,6 @@ import { PreferredCodec } from '../types';
 import { isMobile } from '../compatibility';
 import { isReactNative } from '../helpers/platforms';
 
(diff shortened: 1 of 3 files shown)


myandrienko (approved):

packages/client/src/devices/CameraManager.ts
```diff
@@ -6,6 +6,9 @@ import { getVideoDevices, getVideoStream } from './devices';
 import { TrackType } from '../gen/video/sfu/models/models';
 import { PreferredCodec } from '../types';
 import { isMobile } from '../compatibility';
+import { isReactNative } from '../helpers/platforms';
+
+const isDirectionSupportedByDevice = isMobile() || isReactNative();
```
oliverlaz, line 11: Any reason why this variable is declared top-level?

oliverlaz (approved):

PR #1555, opened 2024-11-04
fix: force baseline profiles for h264 and vp9

oliverlaz:
### Overview

Forces usage of baseline codec profiles for h264 and vp9. 
Furthermore, the codec preferences are grouped now - there won't be `h264, vp9, vp8, h264...` chains anymore.


oliverlaz:
Replaced with #1527

PR #1556, opened 2024-11-05
fix(design-v2): color consolidation and other fixes

kristian-mkd:


packages/react-native-sdk/src/icons/Mic.tsx
```diff
@@ -8,7 +8,7 @@ type Props = {
 };
 
 export const Mic = ({ color, size }: Props) => (
-  <Svg viewBox={`0 0 ${size} ${size}`} width={size} height={size}>
+  <Svg viewBox={'0 0 27 27'} width={size} height={size}>
     <Path
```
santhoshvai, line 12: why change this??

santhoshvai: size will not be changeable then

kristian-mkd: I noticed that best way to adjust the icon size without distortion is through dynamic width={size} height={size}, and keeping the viewBox with the fixed dimensions as it was exported from Figma.

PR #1557, opened 2024-11-05
fix: use worker to prevent timer throttling

myandrienko:


packages/client/src/coordinator/connection/types.ts
```diff
@@ -146,6 +146,12 @@ export type StreamClientOptions = Partial<AxiosRequestConfig> & {
    * In Node.js environment, you can use the `ws` package.
    */
   WebSocketImpl?: typeof WebSocket;
+
+  /**
+   * Create Web Worker to initiate timer events like health checks. Can possibly prevent
+   * timer throttling issues in inactive browser tabs.
+   */
+  expertimental_enableTimerWorker?: boolean;
```
myandrienko, line 154: Feature toggle

oliverlaz (approved):

PR #1585, opened 2024-11-21
fix: unhandled promise rejections during reconnect

myandrienko:
For an explanation of where unhandled rejections are coming from, see https://github.com/GetStream/stream-video-js/pull/1585/files#diff-420f6ddab47c1be72fd9ce8c99e1fa2b9f5f0495b7c367546ee0ff634beaed81


packages/client/src/StreamVideoClient.ts
```diff
@@ -316,19 +315,15 @@ export class StreamVideoClient {
...
     const userId = this.streamClient.user?.id;
     const apiKey = this.streamClient.key;
     const disconnectUser = () => this.streamClient.disconnectUser(timeout);
-    this.disconnectionPromise = this.connectionPromise
-      ? this.connectionPromise.then(() => disconnectUser())
-      : disconnectUser();
-    this.disconnectionPromise.finally(
-      () => (this.disconnectionPromise = undefined),
+    await withoutConcurrency(this.connectionConcurrencyTag, () =>
+      disconnectUser(),
```
myandrienko, line 325: We had a lot of this `disconnectionPromise`/`connectionPromise` chaining to make sure the two don't run at the same time. Replacing with `withoutConcurrency` for simplicity. Also allows us to get rid of two stored promises.

packages/client/src/StreamVideoClient.ts
```diff
@@ -303,8 +304,6 @@ export class StreamVideoClient {
         this.writeableStateStore.registerCall(theCall);
       }),
     );
-
-    return connectUserResponse;
```
oliverlaz, line 307: This is potentially breaking. Can you revert this change back?

myandrienko: Yes, it's a mistake that it was removed. Thanks.

PR #1594, opened 2024-11-25
feat(rn-dogfood): add local sfu support

santhoshvai:


sample-apps/react-native/dogfood/fastlane/Fastfile
santhoshvai, line 1: test for faster testflight reviews - unrelated

oliverlaz (approved):

PR #1600, opened 2024-11-26
fix: handle timeout on SFU WS connections

myandrienko:
### Fixes hanging join requests when SFU WS connection cannot be established

Previously, `signalReady` for SFU WS never rejected if WS connection couldn't be established. We now reject it after a timeout, so that the `call.join()` doesn't hang forever.

We reuse the `joinTimeout` option as a sensible default here.

### Adds support for disconnection timeout setting

When going through the reconnection flow, we now take into account the new `disconnectionTimeoutSeconds` setting, so we don't retry reconnecting indefinitely.


packages/client/src/Call.ts
```diff
@@ -1050,6 +1052,9 @@ export class Call {
    */
   private handleSfuSignalClose = (sfuClient: StreamSfuClient) => {
     this.logger('debug', '[Reconnect] SFU signal connection closed');
+    // SFU WS closed before we finished current join, no need to schedule reconnect
+    // because join operation will fail
+    if (this.state.callingState === CallingState.JOINING) return;
```
myandrienko, line 1057: When joining the call, we already wait for the SFU WS connection to be healthy. If that doesn't happen, we shouldn't try reconnecting, and should throw from `join` instead.

When in-call (`callingState === JOINED`) we should still schedule a reconnect.

oliverlaz (approved):

PR #1611, opened 2024-12-04
fix: multiple rare ringing issues in react-native

santhoshvai:
1. When permission was denied in RN. It was not properly handled. 

Now after this change, if permissions are not granted in RN, our client properly logs like below 

```
ERROR: [devices]: Failed to get audio stream {"constraints": {"audio": {"autoGainControl": true, "deviceId": undefined, "echoCancellation": true, "noiseSuppression": true}}, "error": {"constraintName": undefined, "message": "Permission denied.", "name": "SecurityError"}}
 ERROR: [permissions]: Failed to getUserMedia {"error": {"constraintName": undefined, "message": "Permission denied.", "name": "SecurityError"}, "permission": {"constraints": {"video": [Object]}, "queryName": "camera"}}
```

2. `DOMException ` is not present in RN, so must check for it 

3. Foreground notification can rarely make app go to PiP automatically, fixed in this

4.  hot reloads could add multiple listeners, causing push to be processed twice.. fixed here

5. App going to background, could disable unnecessarily, fixed here

6. Removed the camera handling in incmoing call component to fix race on foreground notification

7. updated the firebase setup with both samples inside stream-video project

8. Apply  importance: 4 to ringing notifications, support request.. channel was already set to this, so doesnt change anything actually


myandrienko (approved):

packages/client/src/devices/BrowserPermission.ts
```diff
@@ -73,10 +73,16 @@ export class BrowserPermission {
...
+            if (typeof DOMException !== 'undefined') {
+              throw new DOMException(
+                'Permission was not granted previously, and prompting again is not allowed',
+                'NotAllowedError',
+              );
+            } else {
+              throw new Error(
+                'Permission was not granted previously, and prompting again is not allowed',
+              );
+            }
```
oliverlaz, line 85: Does it matter if we throw a `DOMException` or a regular `Error`? Can we go with `throw new Error` and not have multiple code branches to achieve the same thing?

santhoshvai: @myandrienko can you clarify here, DOMException does something special on browsers?

santhoshvai: removed it, as per MDN, it was for exception thrown by a web API, so not really necessary here

packages/client/src/devices/devices.ts
```diff
@@ -12,6 +12,7 @@ import { getLogger } from '../logger';
 import { BrowserPermission } from './BrowserPermission';
 import { lazy } from '../helpers/lazy';
 import { isFirefox } from '../helpers/browsers';
+import { isReactNative } from '../helpers/platforms';
```
oliverlaz, line 15:
```suggestion
```

unused.

oliverlaz (approved):

PR #1612, opened 2024-12-05
feat: add an option to filter participants in layouts

myandrienko:
1. Adds a prop to `SpeakerLayout` and `PaginatedGridLayout` to filter participants based on a user predicate.
2. ~~On top of that, add a config option to the composite app to filter participants based on role.~~


oliverlaz (approved):

myandrienko:
After some internal discussion, I'm dropping the update to the composite app from this PR. Users can use the `filterParticipants` prop in their custom recording application.

PR #1615, opened 2024-12-09
fix: avoid call.get in all call.ring events

santhoshvai:
As per spec

```
When the event is received, you need to ensure there is a WS connection open. If the connection needs to be opened, you also need to call call.get to fetch the latest state of the call object
```

Avoids confusion on support where users logging callingstate inform that they get left before ringing state

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -26,6 +26,7 @@ import type {
   AcceptCallResponse,
   BlockUserRequest,
   BlockUserResponse,
+  CallRingEvent,
   CollectUserFeedbackRequest,
   CollectUserFeedbackResponse,
   Credentials,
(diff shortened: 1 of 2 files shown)


packages/client/src/Call.ts
```diff
@@ -595,6 +596,30 @@ export class Call {
...
+    // as the creator does not get the ring event
+    // so update the member list accordingly
+    const creator = this.state.members.find(
+      (m) => m.user.id === event.call.created_by.id,
+    );
+    if (!creator) {
+      this.state.setMembers(event.members);
+    } else {
+      this.state.setMembers([creator, ...event.members]);
+    }
```
santhoshvai, line 615: Alternatively we could skip setting members altogether because call.created event already would have set members

oliverlaz (approved):

PR #1617, opened 2024-12-11
fix: pre-built timer worker

myandrienko:
Dropping the `experimental_` prefix since this feature was confirmed working.

Since the timer worker (introduced in #1557) is not going to be updated too often, we can skip an additional built step and just check in the prebuilt worker.

The "source" file `worker.ts` and the script to generate worker are still there, but they are not used during build.

### DIFF
--- a/packages/client/src/StreamVideoClient.ts
+++ b/packages/client/src/StreamVideoClient.ts
@@ -89,12 +89,11 @@ export class StreamVideoClient {
     if (typeof apiKeyOrArgs === 'string') {
       logLevel = opts?.logLevel || logLevel;
       logger = opts?.logger || logger;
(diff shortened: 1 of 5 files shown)


packages/client/src/StreamVideoClient.ts
```diff
@@ -89,12 +89,11 @@ export class StreamVideoClient {
     if (typeof apiKeyOrArgs === 'string') {
       logLevel = opts?.logLevel || logLevel;
       logger = opts?.logger || logger;
-      if (opts?.expertimental_enableTimerWorker) enableTimerWorker();
+      if (opts?.enableTimerWorker) enableTimerWorker();
```
myandrienko, line 92: Oops, it used to say "expertimental" the whole time!..

PR #1619, opened 2024-12-11
fix: prevent auto-dropping already accepted or rejected calls

myandrienko:
This PR compliments https://github.com/GetStream/chat/pull/7669 and prevents auto-rejection of a ringing call after it was accepted or rejected by the current user (on another device or in another tab).

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -342,6 +342,22 @@ export class Call {
     );
 
     this.leaveCallHooks.add(
+      createSubscription(this.state.session$, (session) => {
+        if (!this.ringing) return;
+
+        const receiverId = this.clientStore.connectedUser?.id;
+        if (!receiverId) return;
+
+        const isAcceptedByMe = Boolean(session?.accepted_by[receiverId]);
+        const isRejectedByMe = Boolean(session?.rejected_by[receiverId]);
+
+        if (isAcceptedByMe || isRejectedByMe) {
+          this.cancelAutoDrop();
+        }
+      }),
+    );
+
+    this.leaveCallHooks.add(
       // watch for auto drop cancellation
       createSubscription(this.state.callingState$, (callingState) => {
         if (!this.ringing) return;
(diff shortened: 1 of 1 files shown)


packages/client/src/Call.ts
```diff
@@ -2000,28 +2016,33 @@ export class Call {
    * Applicable only for ringing calls.
    */
   private scheduleAutoDrop = () => {
-    clearTimeout(this.dropTimeout);
-    this.leaveCallHooks.add(
-      createSubscription(this.state.settings$, (settings) => {
```
myandrienko, line 2005: Previously, a subscription on call settings was set up when scheduling auto-drop. I don't see a reason to set up the subscription here.

Call settings should be up-to-date at this point; and in case call settings were somehow update mid-ringing, this subscription didn't take updated timeouts into account correctly anyway.

So, to simplify things, we are just setting up a timeout using current call settings.

packages/client/src/Call.ts
```diff
@@ -342,16 +342,18 @@ export class Call {
...
-      // watch for auto drop cancellation
-      createSubscription(this.state.callingState$, (callingState) => {
+      // cancel auto-drop when call is
+      createSubscription(this.state.session$, (session) => {
         if (!this.ringing) return;
-        if (
-          callingState === CallingState.JOINED ||
-          callingState === CallingState.JOINING ||
-          callingState === CallingState.LEFT
-        ) {
```
myandrienko, line 352: Cancelling auto-drop on calling state change is no longer required, since it's covered by two other checks:
1. Auto-drop is cancelled if the call is either accepted or rejected by the current user.
2. Auto-drop is a no-op if calling state changed to anything other than `RINGING`.

packages/client/src/Call.ts
```diff
@@ -2000,28 +2002,36 @@ export class Call {
...
+      ? settings.ring.auto_cancel_timeout_ms
+      : settings.ring.incoming_call_timeout_ms;
+
+    // 0 means no auto-drop
+    if (timeoutInMs <= 0) return;
+
+    this.dropTimeout = setTimeout(() => {
+      // the call might have stopped ringing by this point,
+      // e.g. it was already accepted and joined
+      if (this.state.callingState !== CallingState.RINGING) return;
```
myandrienko, line 2022: An additional guard to prevent dropping a call that is no longer `RINGING` (e.g. was accepted or rejected on the current device/tab).

oliverlaz (approved):

PR #1621, opened 2024-12-12
fix: adjust dynascale debouncing for upscaling and downscaling

myandrienko:
We currently debounce requests to update participant subscriptions when a bound video element size changes with a delay of 1200ms. That's too long of delay, but there are also other considerations.

Not all size changes are equal: when upscaling, low quality video in a larger video element is very noticeable; high quality video in a smaller video element for a second or two is not a big deal.

Proposal:
1. When a video element upscales significantly (currently that means that either width or height went up more than 20%), update participant subscription with minimal debouncing.
2. When a video element upscales a little, debounce more aggressively, to guard against quick-firing resize events.
3. When a video element downscales, it's also fine to debounce more aggressively. That will guard us against switching to lower quality when several people are talking over other and the dominant speaker changes quickly.


packages/client/src/helpers/DynascaleManager.ts
```diff
@@ -413,7 +427,7 @@ export class DynascaleManager {
           .subscribe((isPublishing) => {
             if (isPublishing) {
               // the participant just started to publish a track
-              requestTrackWithDimensions(DebounceType.FAST, {
+              requestTrackWithDimensions(DebounceType.IMMEDIATE, {
                 width: videoElement.clientWidth,
                 height: videoElement.clientHeight,
               });
```
myandrienko, line 433: For this case (subscribing after a participant started to publish video), do we even need debouncing? I think it would be nice to just subscribe immediately.

oliverlaz (approved):

PR #1630, opened 2024-12-24
fix(rn-video): upgrade to expo ver 52

kristian-mkd:


packages/react-native-sdk/src/utils/push/utils.ts
```diff
@@ -3,6 +3,10 @@ import { FirebaseMessagingTypes } from './libs/firebaseMessaging';
 import { ExpoNotification } from './libs/expoNotifications';
 import { NonRingingPushEvent } from '../StreamVideoRN/types';
 import { PushNotificationiOSType } from './libs/iosPushNotification';
+import {
+  NotificationTrigger,
+  PushNotificationTrigger,
+} from 'expo-notifications';
```
santhoshvai, line 9: 🛑 this should not be done, we should not be importing types for an optional dependency

santhoshvai (approved):

oliverlaz (approved):

PR #1641, opened 2025-01-09
fix(rn-sdk): fixes the camera status on app restore from background

kristian-mkd:
## Overview

[Liner issue](https://linear.app/stream/issue/RN-97/camera-automatically-turns-on-when-restoring-app-from-background)

Steps to reproduce:
During a call user disables camera and puts the app in background mode. When opens the app the camera is enabled (instead of disabled as previously set by the user)

**The fix:** 
Allow the `resume()` method to enable the camera only if camera was previously disable due to AppState.


packages/react-native-sdk/src/providers/StreamCall/AppStateListener.tsx
```diff
@@ -67,7 +68,10 @@ export const AppStateListener = () => {
             call?.camera?.enable();
           });
         } else {
-          call?.camera?.resume();
+          if (cameraDisabledByAppState) {
```
oliverlaz, line 71:
```suggestion
          if (cameraDisabledByAppState.current) {
```

santhoshvai (approved):

PR #1643, opened 2025-01-15
feat: android 12+ pip uses setAutoEnterEnabled api

santhoshvai:
## Main Goal

Android 12+ now uses setAutoEnterEnabled api instead of being dependent on onUserHint, which is buggy on these platforms when any external dialog like permission dialog comes.

This adds the seamless animation transition support for Android whereever it is available.

## Side optimisiations

* Added a a native workaround for PiP views from react native not being updated on Android 14 and above 🎉
* In JS part, moved PiP to one global state in StreamCall to avoid multiple state updates

## Android 15

https://github.com/user-attachments/assets/1d9d9021-d294-4650-b6c7-b39a0dc56cdf


oliverlaz (approved):

kristian-mkd (approved):

packages/react-native-sdk/expo-config-plugin/src/withMainActivity.ts
```diff
@@ -48,33 +47,36 @@ const withStreamVideoReactNativeSDKMainActivity: ConfigPlugin<ConfigProps> = (
...
       statementToInsert = `         
-      override fun onPictureInPictureModeChanged(isInPictureInPictureMode: Boolean) {
-          super.onPictureInPictureModeChanged(isInPictureInPictureMode)
-          if (Build.VERSION.SDK_INT >= Build.VERSION_CODES.O && lifecycle.currentState == Lifecycle.State.CREATED) {
-              // when user clicks on Close button of PIP
-              finishAndRemoveTask()
-          } else {
-              StreamVideoReactNative.onPictureInPictureModeChanged(isInPictureInPictureMode)
-          }
+      fun onPictureInPictureModeChanged(isInPictureInPictureMode: Boolean, newConfig: Configuration) {
```
kristian-mkd, line 72: This change causes this build failure on my side:
```
> Task :app:compileDebugKotlin FAILED
e: file:///Users/mkristian/Projects/stream/stream-video-js/sample-apps/react-native/expo-video-sample/android/app/src/main/java/io/getstream/expovideosample/MainActivity.kt:72:5 'onPictureInPictureModeChanged' hides member of supertype 'ReactActivity' and needs 'override' modifier

FAILURE: Build failed with an exception.
```
I have tried adding the `override` back and that fixes the issue.

PR #1644, opened 2025-01-15
feat: [VID-160] call inspector page

myandrienko:
## Call inspector

This is the call inspector page (`/inspect`):

image

It shows device encoding and decoding capabilities (inspector does its best to show capabilities in human-readable format), as well as available media input devices.

When connected to a call, it shows a little bit more.

### Joining a call

To join an existing call, enter call type, id, API key and user token, and press <kbd>Enter</kbd>.

Pronto has a _Go to Inspector_ option in the dev menu:

image

Finally, you can press _Join demo call_, and a new test call will be created and joined on the current pronto environment. After you leave the test call, its credentials will remain entered, so you can reconnect.

### In call

When in-call, a few more options become available.

image

Header shows call cid and current calling state. A little call connection log appears. If there were any problems while joining the call they will be displayed in the log.

Media input device dashes become interactive, allowing to select and toggle devices.

And a couple more dashes appear:

image

_Connectivity_ dash shows current connection state to Coordinator and SFU. Little heartbeats show when status was last polled. There's a button to force close connection - it can be used to emulate network drop or server error.

When publishing, it also features a transport diagram that visualizes the selected candidate pair for the RTC connection. _Re-ICE_ button forces forces renegotiation.

_Codecs in use_ dash list codecs that are currently used in the call. That depends on other call participants, so there's a button to force subscribing to video from all call participants.

### Markdown report

At any point you can press the little clipboard button in the top right to copy full report contents in Markdown format.

### Client and call instance

Current client and call instances are exposed via the `_inspector` global variable, so you can inspect it using developer console.

## Device test page

The device test page (`/test`) is absolutely the same as inspector, except for one thing: it doesn't allow joining an arbitrary call, and always connects to the demo call as soon as possible. Sharing a link to this page requires less explanation: the user sees immediately sees all the data, without additional steps.


packages/client/src/Call.ts
```diff
@@ -2473,4 +2473,12 @@ export class Call {
...
     this.disconnectionTimeoutSeconds = timeoutSeconds;
   };
+
+  getConnectionString = (): string => {
+    const cid = this.cid;
+    const apiKey = this.streamClient.key;
+    const token = this.streamClient.tokenManager.getToken();
+
+    return `${cid}@${apiKey}:${token}`;
+  };
```
oliverlaz, line 2483: I have mixed feelings about this method. It is too specific to our inspector page implementation.
I think we should lift it to the NextJS app.

myandrienko: Fixed in https://github.com/GetStream/stream-video-js/pull/1644/commits/498dccb2a6dd4dbe6947220e6d4e59aa66808788.

oliverlaz (approved):

PR #1654, opened 2025-01-23
fix: leave ringing call if accepted or rejected elsewhere

myandrienko:
Immediately leave ringing call if it was accepted by the same user elsewhere.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -128,7 +128,7 @@ import {
   SpeakerManager,
 } from './devices';
 import { getSdkSignature } from './stats/utils';
-import { withoutConcurrency } from './helpers/concurrency';
+import { hasPending, withoutConcurrency } from './helpers/concurrency';
 import { ensureExhausted } from './helpers/ensureExhausted';
 import { pushToIfMissing } from './helpers/array';
 import {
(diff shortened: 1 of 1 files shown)


santhoshvai:
maybe we should add more here? say rejected at another device 

https://github.com/GetStream/stream-video-js/blob/051dda26383ada3215cfe5f1ee89e2959e4186f7/packages/client/src/Call.ts#L398-L416

packages/client/src/Call.ts
```diff
@@ -373,6 +373,14 @@ export class Call {
         if (isAcceptedByMe || isRejectedByMe) {
           this.cancelAutoDrop();
         }
+
+        if (
+          isAcceptedByMe &&
+          this.state.callingState === CallingState.RINGING
+        ) {
+          // call accepted by user in another session
+          this.leave();
```
oliverlaz, line 382: Check @santhoshvai's comment. I think that's a good point. And also, I think we should have a `.catch()` here to prevent potential Unhandled Promise Rejections

myandrienko: Agreed, we should extend this behavior to rejected calls as well.

myandrienko: @santhoshvai The code looked at is responsible for changing calling state when the call starts ringing. When the call starts ringing, but it is already rejected by the user, it should indeed state idle (as if it didn't ring at all).

But we should handle rejecting call after it started ringing, that's true.

oliverlaz (approved):

santhoshvai (approved):

PR #1670, opened 2025-02-03
fix: closed captions and speaker border bugfixes

kristian-mkd:
## Overview
- adds default transparent border to the speaker, so when speaker is active the blue border does not resize the video of the current speaker which previously led to a lot of unnecessary SFU updateSubscriptions requests 
- fixes the closed captions formatting when there are long messages
- fixes the bottom snap point of the controls drawer in the dogfood app

ios-after


santhoshvai (approved):

packages/react-native-sdk/src/components/Participant/ParticipantView/ParticipantView.tsx
```diff
@@ -174,6 +174,8 @@ const useStyles = () => {
           overflow: 'hidden',
           justifyContent: 'flex-end',
           borderRadius: theme.variants.borderRadiusSizes.md,
+          borderWidth: 2,
```
santhoshvai, line 177: if border width is added, no need to have a seperated style for highlightedContainer.. that must be removed

PR #1671, opened 2025-02-04
fix: do not stop track on camera flip

santhoshvai:


packages/client/src/devices/CameraManager.ts
```diff
@@ -34,10 +34,31 @@ export class CameraManager extends InputMediaDeviceManager<CameraManagerState> {
...
+            facingMode: direction === 'front' ? 'user' : 'environment',
+          };
+          videoTrack.applyConstraints(constraints);
+        }
+        this.state.setDirection(direction);
+      } else {
+        // web mobile
+        this.state.setDirection(direction);
+        // Providing both device id and direction doesn't work, so we deselect the device
+        this.state.setDevice(undefined);
```
myandrienko, line 56: I think this should go outside of this `if`.

packages/client/src/devices/CameraManager.ts
```diff
@@ -34,10 +34,31 @@ export class CameraManager extends InputMediaDeviceManager<CameraManagerState> {
...
+      if (isReactNative()) {
+        const videoTrack = this.getTracks()[0];
+        if (!videoTrack) return;
+        // @ts-expect-error _switchCamera() is only present in react-native-webrtc 124 and below
+        if (typeof videoTrack._switchCamera === 'function') {
+          // @ts-expect-error for older versions of react-native-webrtc support
+          videoTrack._switchCamera();
+        } else {
+          const constraints = {
+            facingMode: direction === 'front' ? 'user' : 'environment',
```
myandrienko, line 47: Should we add `...this.constraints` here? Otherwise we lose target resolution constraint.

packages/client/src/devices/CameraManager.ts
```diff
@@ -34,10 +34,31 @@ export class CameraManager extends InputMediaDeviceManager<CameraManagerState> {
...
+          };
+          videoTrack.applyConstraints(constraints);
+        }
+        this.state.setDirection(direction);
```
myandrienko, line 51: Since `setMediaStream` is not called in this case, there's a slight change of behavior: deviceId is not set to an actual deviceId from track settings. I don't know if it's important for RN though.

santhoshvai: it is not important for RN, the change happens under the hood

myandrienko (approved):

PR #1678, opened 2025-02-10
fix: warn natively if notifee is not configured for keep call alive

### LINKED ISSUE #1587: App gettting crashed on android on final step

- [ ] `@stream-io/video-react-native-sdk`

**Describe the bug**
```
  useEffect(() => {
    const _call = client?.call('default', callId);
    _call
      ?.join({create: true})
      .then(() => {
        console.log("Call created successfully");
        
        setCall(_call) //App gets crashed when I set _call in state
      })
  }, [client, callId]);
```

I tried following the guideline for react native but in the end when I set call in state. App is getting crashed.

**To Reproduce**
Attaching a repo which might help in reproduction:
https://github.com/aliwaqar981/GetStreamRePro

 I tried to create reproduction in separate project. It was working correctly there. My new project was on react-native 0.76.2. I'm thinking maybe that's the reason but I'm not sure.

**Expected behavior**
Video call should get starts after running above code snippet.

**Screenshots**

**Version**

- SDK version
- "@stream-io/video-react-native-sdk": "^1.3.5",
- @stream-io/react-native-webrtc: ^124.0.2

**Env:**

Fill out what's applicable

**React Native**
react-native: 0.74.3

- OS: android
- Device or emulator: Vivo v21e

**Node JS client**

- Node version 20.12.0

**Additional context**
It seems some compatibility issue of sdk with react-native 0.74.3 because it worked on 0.76.2.

santhoshvai:
fixes #1587

### DIFF
--- a/packages/react-native-sdk/android/src/main/java/com/streamvideo/reactnative/util/CallAlivePermissionsHelper.kt
+++ b/packages/react-native-sdk/android/src/main/java/com/streamvideo/reactnative/util/CallAlivePermissionsHelper.kt
@@ -1,0 +1,50 @@
+package com.streamvideo.reactnative.util
+
+import android.content.pm.PackageInfo
+import android.content.pm.PackageManager
+import android.util.Log
+import com.facebook.react.bridge.ReactApplicationContext
+
+
+object CallAlivePermissionsHelper {
+    private const val NAME = "StreamVideoReactNative"
+
+    fun hasForegroundServicePermissionsDeclared(context: ReactApplicationContext): Boolean {
+        val packageManager = context.packageManager
+        val packageName = context.packageName
+
+        try {
+            val packageInfo: PackageInfo =
+                packageManager.getPackageInfo(packageName, PackageManager.GET_PERMISSIONS)
+            val requestedPermissions = packageInfo.requestedPermissions
+
+            if (requestedPermissions != null) {
+                val requiredPermissions = arrayOf(
+                    "android.permission.FOREGROUND_SERVICE",
+                    "android.permission.FOREGROUND_SERVICE_CAMERA",
+                    "android.permission.FOREGROUND_SERVICE_MICROPHONE",
+                    "android.permission.FOREGROUND_SERVICE_CONNECTED_DEVICE",
+                    "android.permission.FOREGROUND_SERVICE_DATA_SYNC"
+                )
+
+                val missingPermissions =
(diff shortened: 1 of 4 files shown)


packages/react-native-sdk/android/src/main/java/com/streamvideo/reactnative/util/CallAlivePermissionsHelper.kt
```diff
@@ -0,0 +1,50 @@
...
+                        "Missing ForegroundServicePermissions: ${missingPermissions.joinToString(", ")}"
+                    )
+                    return false
+                } else {
+                    return true
+                }
+            }
+        } catch (e: PackageManager.NameNotFoundException) {
+            // Package not found (shouldn't happen for the current app)
+            e.printStackTrace()
```
oliverlaz, line 45: This is a bad practice. How can we better handle this exception?

packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
```diff
@@ -11,12 +16,25 @@ import {
...
-    });
-  });
+  if (!notifeeLib) return;
+  NativeModules.StreamVideoReactNative.isCallAliveConfigured().then(
+    (isConfigured: boolean) => {
+      if (!isConfigured) {
+        const logger = getLogger(['setForegroundService method']);
+        logger(
+          'info',
+          'Skipping KeepCallAlive foreground service setup as it is not configured properly'
```
oliverlaz, line 26:
```suggestion
          'KeepCallAlive is not configured. Skipping foreground service setup.'
```

Maybe this is intentional. `properly` sounds like we want customers to always enable it.

packages/react-native-sdk/android/src/main/java/com/streamvideo/reactnative/util/CallAlivePermissionsHelper.kt
santhoshvai: made it better by adding a better logger

but this is not bad practice I think. Atleast I do not know. This catch block was autogenerated by android studio. So I do not think so 😆

packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
santhoshvai: changed

oliverlaz (approved):

PR #1685, opened 2025-02-14
feat: support static token and token provider at the same time

oliverlaz:
### Overview

Until now, providing both, `token` and `tokenProvider` wasn't truly supported, and when this was the case, the `client` always used the `static` token.

This PR allows the client to connect faster and save one roundtrip to the customer's servers.

The logic is the following:
- if only `token` is provided - use the `token` (existing behavior)
- if only `tokenProvider` is provided - use the `tokenProvider` (existing behavior)
- if both `token` and `tokenProvider` are provided - use the `token` initially. Once it expires, use the `tokenProvider` (new behavior).

The new behavior aligns with the logic that the other SDKs have, and it potentially reduces the time it takes to establish a WS connection to our systems.

In addition, we now validate internally that the token provided by the token provider is issued for the correct `user_id.` This should prevent potential integration issues.

This PR also refactors and cleans the code used to construct a new `StreamVideoClient` and also adds test cases for the critical parts of it.

Also, a few dependencies are updated to their latest versions (`axios`, `sdp-transform`, `ua-parser-js`, `vitest`, etc...).


santhoshvai (approved):

packages/client/src/coordinator/connection/token_manager.ts
```diff
@@ -36,8 +35,9 @@ export class TokenManager {
     user: UserWithId,
     isAnonymous: boolean,
   ) => {
-    this.validateToken(tokenOrProvider, user, isAnonymous);
     this.user = user;
+    this.isAnonymous = isAnonymous;
```
myandrienko, line 39: Could be not an issue at all, but could be dangerous. Previously, `isAnonymous` was a local variable, so even if `setTokenOrProvider` was not awaited - it would kind of work. Now `isAnonymous` can be mutated in the middle of loading token.

myandrienko (approved):

packages/client/src/coordinator/connection/token_manager.ts
myandrienko: On a second thought, nothing here works if `setTokenOrProvider` is not awaited, so it's not a regression.

oliverlaz: yeah, things are super racy here and also intertwined

PR #1686, opened 2025-02-18
fix: do not remove voip listeners for wrong user

santhoshvai:
### Overview

Consider this scenario where user A logs out and then user B logs in

1. User B logs in, new event listeners are attached
2. StreamVideo useEffect for A unmounts listeners are removed
3. Now there is no more event listeners

This is fixed by this PR

Additionally: also added a method for clearing logout callbacks

### DIFF
--- a/packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
+++ b/packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
@@ -24,6 +24,12 @@ import { RxUtils, StreamVideoClient, getLogger } from '@stream-io/video-client';
 
 const logger = getLogger(['useIosVoipPushEventsSetupEffect']);
 
+/* VoipPushNotificationLib has support for only one listener type at a time
+ hence to support login and logout scenario of multiple users we keep of the last count of the listener that was added
+ This helps in not removing the listeners when a new user logs in and overrides the last listener
+*/
+let lastListenerCount = 0;
+
 function setLogoutCallback(
   client: StreamVideoClient,
   token: string,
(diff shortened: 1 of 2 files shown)


packages/react-native-sdk/src/utils/StreamVideoRN/index.ts
```diff
@@ -102,6 +102,10 @@ export class StreamVideoRN {
     return Promise.resolve();
   }
 
+  static clearPushLogoutCallbacks() {
```
kristian-mkd, line 105: Is this method planned for some future use? Or is it forgotten to be used at the moment?

packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
```diff
@@ -160,12 +166,19 @@ export const useIosVoipPushEventsSetupEffect = () => {
...
+    const currentListenerCount = lastListenerCount;
+
     return () => {
-      logger(
-        'debug',
-        'Voip event listeners are removed for user: ' +
-          client.streamClient._user?.id
-      );
+      const userId = client.streamClient._user?.id;
+      if (currentListenerCount !== lastListenerCount) {
```
kristian-mkd, line 174: Is it possible to have some userId comparison here instead of the count check? To store the `lastLoggedInUserId` and `currentLoggedInUserId` and distinct them like that?

packages/react-native-sdk/src/utils/StreamVideoRN/index.ts
santhoshvai: for use by integrators on user switch  not from sdk

packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
santhoshvai: no, user could be undefined on the client level, just count suffices as the library overrides the listener

PR #1687, opened 2025-02-18
fix: early completion call on iOS remote notifications

santhoshvai:
### Overview

rarely didDisplayIncomingCall comes after onNotificationReceived completes, this is problematic as per the voipNotification library. We must complete only at the last after all are processed it seems. 

But I dont know what the problem is though. 

### Test

Just ensure that JS calls the completion function in all three scenarios

1. Foregrounded app
2. Backgrounded app
3. Killed off app


packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
santhoshvai, line 1: this is unrelevant to this PR, but aims to fix a bug here, useEffect now captures the last count by reference so that there is no chance of a state value

oliverlaz (approved):

kristian-mkd (approved):

PR #1695, opened 2025-02-25
fix: expo build issues

santhoshvai:
### Overview

Expo by default uses commonJS, this is because of app.plugin.js support. The recent upgrade of builder bob in the SDK without any other changes, broke the common js build. 

### Implementation notes
 * Builder bob now has its own babel config for easiness and consitency, we use that now
 * builder bob has custom script support now. We use that for expo config plugins
 * `verbatimModuleSyntax: true` added now as recommended by builderbob. This means we have to add explicit type imports.
 *  monospace root detection disabled for expo cli inisde the sample to make the build work. TODO: future we must remove the hositinglimit and support expo default root detection.
 * ESM support breaks the flow syntax used in old RN versions. So we have it disabled for now. We can enable once our minimum RN moves to where flow was removed.
 * we use `expo-module-scripts` cli now for the config plugin. And extend their default configs.
 
 ### Test

* expo sample and dogfood must run without build issues


packages/react-native-sdk/package.json
```diff
@@ -132,8 +131,9 @@
     "@types/react-native": "^0.70.4",
     "@types/react-native-incall-manager": "^3.2.1",
     "@types/react-test-renderer": "^18",
-    "expo": "50.0.19",
-    "expo-build-properties": "^0.12.3",
+    "expo": "52.0.37",
+    "expo-build-properties": "^0.13.2",
+    "expo-module-scripts": "^4.0.4",
```
santhoshvai, line 136: updated for addition of module scripts

kristian-mkd (approved):

oliverlaz (approved):

PR #1702, opened 2025-02-26
fix: do not stop incall manager on change of audio settings

santhoshvai:
If call settings are changed on the fly when a call is already joined, then incall manager is started and stopped again. 

InCallmanger mode with "audio" is almost never used by our integrators. Speaker is the default preferred output in any scenario. We should not route to earpiece in video disabled calls by default too. I do not know why this was added.
 
This behaviour also aligns with other SDKs and also fixes the bug on iOS that audio is lost after start->stop->start

### DIFF
--- a/packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
+++ b/packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
@@ -85,9 +85,10 @@ export type CallContentProps = Pick<
      */
     disablePictureInPicture?: boolean;
     /**
-     * Props to set the audio mode for the InCallManager
-     * If media type is video, audio is routed by default to speaker otherwise it is routed to earpiece
-     * Changing the mode on the fly is not supported. Manually invoke `InCallManager.start({ media })` to achieve this.
+     * Props to set the audio mode for the InCallManager.
+     * If media type is video, audio is routed by default to speaker, otherwise it is routed to earpiece.
+     * Changing the mode on the fly is not supported.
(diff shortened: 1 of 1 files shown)


packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
```diff
@@ -84,6 +84,14 @@ export type CallContentProps = Pick<
...
      */
     disablePictureInPicture?: boolean;
+    /**
+     * Props to set the audio mode for the InCallManager.
+     * If media type is video, audio is routed by default to speaker, otherwise it is routed to earpiece.
+     * Changing the mode on the fly is not supported.
+     * Manually invoke `InCallManager.start({ media })` to achieve this.
+     * @default 'video'
+     */
+    inCallManagerAudioMode?: 'video' | 'audio';
```
oliverlaz, line 94: I'd rename this prop to something else. Something like `defaultInCallManagerAudioMode` or `initialInCallManagerAudioMode` to signify that it works only during the initial render. Otherwise, customers might try changing it at runtime and not get the desired behavior.

kristian-mkd (approved):

packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
santhoshvai: good point, updated the name to initialInCallManagerAudioMode

oliverlaz (approved):

PR #1739, opened 2025-03-28
chore: rn-178 refactor theme provider

kristian-mkd:
### Overview

https://linear.app/stream/issue/RN-178/rename-themeprovider-to-streamtheme

### Implementation notes
Adds `StreamTheme` alias for the `ThemeProvider`.

### DIFF
--- a/packages/react-native-sdk/src/providers/StreamVideo.tsx
+++ b/packages/react-native-sdk/src/providers/StreamVideo.tsx
@@ -52,25 +52,20 @@ export const StreamVideo = (
     });
   }, [client]);
 
-  const videoProvider = (
+  return (
     <StreamVideoProvider
       client={client}
       language={language}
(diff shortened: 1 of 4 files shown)


packages/react-native-sdk/src/providers/StreamVideo.tsx
```diff
@@ -52,21 +52,25 @@ export const StreamVideo = (
...
-        </StreamVideoStoreProvider>
-      </ThemeProvider>
+      <StreamVideoStoreProvider>
+        <PushRegister />
+        {children}
+      </StreamVideoStoreProvider>
     </StreamVideoProvider>
   );
+
+  if (style) {
```
kristian-mkd, line 69: To provide backward compatibility and the keep the old solution as well passing the theme as style prop.

santhoshvai (approved):

PR #1746, opened 2025-04-03
fix: Document PiP support on Pronto

myandrienko:
### Overview

Implements pop-out call view for Pronto using the Document Picture-in-Picture API in supported browsers.

Screen Shot 2025-04-03 at 17 53 53 PM

### Implementation notes

The hook `usePipWindow` in react-dogfood app is responsible for managing PiP window state. The call is rendered into the window using a cross-document React portal. Stylesheets are copied over from the parent document.

This hook also sets up the `enterpictureinpicture` action handler with the Media Session API. That way PiP activates automatically on tab switch. If you find this annoying, revoke "Automatic Picture-in-Picture" permission.

The new `PipLayout` call layout is added. It renders as many participants as it can fit in single-column layout. The layout has two parts:

1. `PipLayout.Pip` should be rendered in the PiP window. It renders participant video.
2. `PipLayout.Host` should be rendered in the parent window. It renders participant audio. Participant audio might not play in PiP window because it doesn't have user interactions.


sample-apps/react/react-dogfood/hooks/usePipWindow.tsx
```diff
@@ -0,0 +1,117 @@
...
+    documentPictureInPicture: {
+      requestWindow: (options?: {
+        width: number;
+        height: number;
+      }) => Promise<Window>;
+    };
+  }
+}
+
+export function usePipWindow(key: string) {
```
oliverlaz, line 21: On second thought... why not make this hook part of the SDK itself?
We can extend it with extra config params and we'll be good to go.

packages/react-sdk/src/core/components/CallLayout/PipLayout.tsx
```diff
@@ -0,0 +1,103 @@
...
+          VideoPlaceholder={VideoPlaceholder}
+          PictureInPicturePlaceholder={PictureInPicturePlaceholder}
+          ParticipantViewUI={ParticipantViewUI}
+        />
+      ))}
+    </div>
+  );
+};
+
+PipLayout.displayName = 'PaginatedGridLayout';
```
oliverlaz, line 103:
```suggestion
PipLayout.displayName = 'PiPLayout';
```

packages/react-sdk/src/core/components/CallLayout/PipLayout.tsx
```diff
@@ -0,0 +1,103 @@
...
+   * <PaginatedGridLayout
+   *   filterParticipants={{
+   *     $or: [
+   *       { roles: { $contains: 'student' } },
+   *       { isPinned: true },
+   *     ],
+   *   }}
+   * />
+   */
+  filterParticipants?: ParticipantPredicate | ParticipantFilter;
```
oliverlaz, line 42:
```suggestion
  /**
   * Predicate to filter call participants or a filter object.
   * @example
   * // With a predicate:
   * <PipLayout
   *   filterParticipants={p => p.roles.includes('student')}
   * />
   * @example
   * // With a filter object:
   * <PipLayout
   *   filterParticipants={{
   *     $or: [
   *       { roles: { $contains: 'student' } },
   *       { isPinned: true },
   *     ],
   *   }}
   * />
   */
  filterParticipants?: ParticipantPredicate | ParticipantFilter;
```

packages/react-sdk/src/core/components/CallLayout/PipLayout.tsx
```diff
@@ -0,0 +1,103 @@
...
+          key={participant.sessionId}
+          participant={participant}
+          muteAudio
+          mirror={mirror}
+          VideoPlaceholder={VideoPlaceholder}
+          PictureInPicturePlaceholder={PictureInPicturePlaceholder}
+          ParticipantViewUI={ParticipantViewUI}
+        />
+      ))}
+    </div>
```
oliverlaz, line 99: How would this work in calls with a lot of participants? Shall we utilize `PaginatedGrid` here?

myandrienko: I went for just truncating participant list to however many participants fit in the window. I found that paginating interface is too much for a small popup window.

sample-apps/react/react-dogfood/hooks/usePipWindow.tsx
myandrienko: We decided to keep it app-level for now: various preparation steps for the PiP window, like setting up layout, adding analytics, copying over stylesheets - might vary too much for a one-size-fits-all solution.

PR #1752, opened 2025-04-08
fix: send device token when switching user

oliverlaz:
### Overview

When switching a `user` by reusing the same `client` instance, we should also send the existing device token to our backend as otherwise, the device remains registered to the old `user` until the new one restarts the app.

This PR also fixes another issue - we shouldn't send device tokens to anonymous users.

Ref: https://linear.app/stream/issue/RN-189/video-update-device-token-when-switching-a-user

### DIFF
--- a/packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
+++ b/packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
@@ -50,31 +50,36 @@ export const useIosVoipPushEventsSetupEffect = () => {
 
   // this effect is used to send the unsent token to stream
   useEffect(() => {
-    const pushConfig = StreamVideoRN.getConfig().push;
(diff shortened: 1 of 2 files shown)


packages/react-native-sdk/src/hooks/push/useIosVoipPushEventsSetupEffect.ts
```diff
@@ -92,15 +97,15 @@ export const useIosVoipPushEventsSetupEffect = () => {
...
     if (!pushConfig.android.incomingCallChannel) {
       // TODO: remove this check and find a better way once we have telecom integration for android
-      getLogger(['useIosVoipPushEventsSetupEffect'])(
+      logger(
         'debug',
         'android incomingCallChannel is not defined, so skipping the useIosVoipPushEventsSetupEffect',
       );
       return;
     }
     if (!pushConfig.android.incomingCallChannel) {
```
kristian-mkd, line 106: this is duplicate, can be deleted. Copilot noticed it.

kristian-mkd (approved):

PR #1772, opened 2025-04-28
feat(react-native): take screenshot of a participant's video stream

kristian-mkd:
### 📝 Implementation notes

```
  const { takeScreenshot } = useScreenshot();
  const base64PngImage = await takeScreenshot(participant, 'videoTrack');
  
  // then usage
  <Image
    source={{ uri: `data:image/png;base64,${base64PngImage}` }}
    resizeMode="contain"
  />
```

🎫 Ticket: https://linear.app/stream/issue/RN-75


packages/react-native-sdk/src/contexts/internal/SnapshotContext.tsx
```diff
@@ -0,0 +1,152 @@
...
+const SnapshotContext = createContext<SnapshotContextType | undefined>(
+  undefined,
+);
+
+// Reference map type
+type RefMap = Map<string, RefObject<any>>;
+
+export const SnapshotProvider = ({ children }: React.PropsWithChildren<{}>) => {
+  // Use a ref to store the map of participant IDs to their view refs
+  const participantRefs = useRef<RefMap>(new Map());
```
oliverlaz, line 44: This creates a `new Map()` on every render and puts pressure on the GC.
Do we have to use a ref here?

packages/react-native-sdk/src/components/Participant/ParticipantView/VideoRenderer.tsx
```diff
@@ -86,6 +91,29 @@ export const VideoRenderer = ({
...
+  React.useEffect(() => {
+    if (
+      Platform.OS === 'ios' &&
+      registerSnapshot &&
+      viewRef.current &&
+      canShowVideo
+    ) {
+      registerSnapshot(participant, trackType, viewRef);
+    } else {
+      deregisterSnapshot(participant, trackType);
```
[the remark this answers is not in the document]
santhoshvai: I agree, let me change it

packages/react-native-sdk/src/contexts/internal/SnapshotContext.tsx
santhoshvai: unnecessary and its now removed

oliverlaz (approved):

PR #1774, opened 2025-04-28
fix: use more granular permission state for stats reporter

myandrienko:
### 💡 Overview

Some time ago we introduced device reporting in call stats (https://github.com/GetStream/stream-video-js/pull/1533).

The unexpected side effect of this is that stats reporter now tries to list available devices every time the call is joined, even if camera and mic are off by default, causing permission prompt to appear.

This PR fixes that: stats reporter will never cause a permission prompt to appear.

### 📝 Implementation notes

The condition for listing devices was:
1. User has capability to send audio or video
2. Device permission was not explicitly denied (this condition comes from the `hasBrowserPermission$` observable, which is only false if permission state is `denied`)

This PR adds `browserPermissionState$` observable to input device manager, which provides more granular visibility into current permission state. Most importantly, it has different values for `prompt` and `granted` states.

Then we can rely on `browserPermissionState$` in stats reporter, and only report devices when permission is already granted, avoiding prompts.

📑 Docs: https://github.com/GetStream/docs-content/pull/311

### DIFF
--- a/packages/client/src/devices/BrowserPermission.ts
+++ b/packages/client/src/devices/BrowserPermission.ts
@@ -139,6 +139,10 @@ export class BrowserPermission {
     );
   }
 
+  asStateObservable() {
+    return this.getStateObservable();
+  }
+
   getIsPromptingObservable() {
     return this.getStateObservable().pipe(
       map((state) => state === 'prompting'),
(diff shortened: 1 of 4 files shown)


oliverlaz (approved):
nice! 👍

PR #1776, opened 2025-04-30
fix: use timeout reason when auto-dropping calls (instead of decline)

szuperaz:
### 💡 Overview

When auto dropping calls the client sent `decline` instead of `timeout` reason

### 📝 Implementation notes

The `reason` field previously was misused, instead of sending it coordinator, we sent it as a log message to SFU. Now `reason` is sent to coordinator, and a new field is introduced - `message` - to send to SFU.

I checked and all UI SDKs use the `reason` field correctly - so no changes were required there.

🎫 Ticket: https://linear.app/stream/issue/REACT-363/fix-call-rejection-reasons-in-js-client-for-timeout-scenarios

📑 Docs: N/A

### DIFF
--- a/packages/client/src/events/call.ts
+++ b/packages/client/src/events/call.ts
@@ -58,12 +58,16 @@ export const watchCallRejected = (call: Call) => {
         .every((m) => rejectedBy[m.user_id]);
       if (everyoneElseRejected) {
         call.logger('info', 'everyone rejected, leaving the call');
-        await call.leave({ reason: 'ring: everyone rejected' });
+        await call.leave({
+          reject: true,
+          reason: 'cancel',
+          message: 'ring: everyone rejected',
+        });
       }
     } else {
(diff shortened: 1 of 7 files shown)


packages/client/src/events/call.ts
```diff
@@ -58,12 +58,16 @@ export const watchCallRejected = (call: Call) => {
         .every((m) => rejectedBy[m.user_id]);
       if (everyoneElseRejected) {
         call.logger('info', 'everyone rejected, leaving the call');
-        await call.leave({ reason: 'ring: everyone rejected' });
+        await call.leave({
+          reject: true,
+          reason: 'cancel',
+          message: 'ring: everyone rejected',
```
szuperaz, line 64: This was working fine even before my fix (due to https://github.com/GetStream/stream-video-js/blob/main/packages/client/src/Call.ts#L583), but this makes it more explicit

oliverlaz (approved):

PR #1780, opened 2025-05-06
feat: rn livestream improvements

kristian-mkd:
### 💡 Overview

Livestream player improvements in the SDK and dogfood app based on this https://www.notion.so/stream-wiki/Livestream-Player-improvements-1d26a5d7f9f68016b057d007548c4179

### 📝 List of improvements

- Refactors the Livestream Player component so that can be used standalone for viewer livestream experience 
- Adds new `ViewerLobby` component to the SDK that provides provides info about:
  - remaining time to start of the livestream
  - already joined participants
  - handles possible call related errors
  - call ended state
  - handles joining a call ahead of time (support for 'asap' or 'live' join behaviours)
  - presentation of call recordings
- Updates the Livestream viewer controls
  - play/pause of the video stream
  - toggling the controls pane
- adds QR scanning option for getting call id in the dogfood app

🎫 Ticket: https://linear.app/stream/issue/RN-200/rn-livestream-improvements

📑 Docs: https://github.com/GetStream/docs-content/pull/344

ios
ios
ios
ios
ios
ios


packages/react-native-sdk/src/components/Livestream/LivestreamPlayer/LivestreamEnded.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const showRecordings =
+    recordingsResponse && recordingsResponse.recordings.length > 0;
+
+  return (
+    <View style={styles.container}>
+      <Text style={styles.title}>The livestream has ended.</Text>
+
+      {showRecordings && (
+        <>
+          <Text style={styles.subtitle}>Watch recordings:</Text>
```
oliverlaz, line 57: These labels need to be translatable by integrators.

packages/react-native-sdk/src/components/Livestream/ViewerLivestream/ViewerLivestream.tsx
```diff
@@ -76,6 +79,7 @@ export const ViewerLivestream = ({
 
   const [topViewHeight, setTopViewHeight] = React.useState<number>();
   const [controlsHeight, setControlsHeight] = React.useState<number>();
+  const [callJoined, setCallJoined] = useState<boolean>(false);
```
oliverlaz, line 82: This is a redundant value that can be derived from `callingState`.

packages/react-native-sdk/src/components/Livestream/ViewerLivestream/ViewerLivestream.tsx
```diff
@@ -92,9 +96,55 @@ export const ViewerLivestream = ({
...
+        return;
+      }
+      await call?.join();
+      setCallJoined(true);
+    } catch (error) {
+      console.error('Failed to join call', error);
+    }
+  };
+
+  const callingState = call?.state.callingState;
```
oliverlaz, line 126:
```suggestion
  const callingState = useCallCallingState();
```
Accessing the state `getter` directly isn't reactive and will lead to many bugs. Please use the hook.

packages/react-native-sdk/src/components/Livestream/ViewerLivestream/ViewerLobby.tsx
```diff
@@ -0,0 +1,168 @@
...
+  const { useCallStartsAt, useParticipants } = useCallStateHooks();
+  const call = useCall();
+  const startsAt = useCallStartsAt();
+  const [error, setError] = useState<Error | undefined>(undefined);
+  const [countdown, setCountdown] = React.useState('');
+  const participants = useParticipants();
+
+  useEffect(() => {
+    setCallJoined(false);
+  }, [setCallJoined]);
```
oliverlaz, line 34: Most likely unnecessary `useEffect`. This data is already part of `call.state.callingState`.
Also, if `props.setCallJoined` function is declared inline, this effect will cause top level state to always be reset to `false`.

packages/react-native-sdk/src/components/Livestream/ViewerLivestream/ViewerLobby.tsx
```diff
@@ -0,0 +1,168 @@
...
+    );
+  }
+
+  return (
+    <View style={styles.container}>
+      <Text style={styles.text}>
+        {isLive
+          ? t('Stream is ready!')
+          : startsAt
+            ? `Livestream will start in:`
```
oliverlaz, line 118: This line needs to be translatable too.

sample-apps/react-native/dogfood/ios/Podfile.lock
```diff
@@ -2100,6 +2100,13 @@ PODS:
     - ReactCommon/turbomodule/core
     - stream-react-native-webrtc
     - Yoga
+  - VisionCamera (4.6.4):
+    - VisionCamera/Core (= 4.6.4)
+    - VisionCamera/React (= 4.6.4)
+  - VisionCamera/Core (4.6.4)
+  - VisionCamera/React (4.6.4):
+    - React-Core
+  - WebRTC-SDK (125.6422.07)
```
oliverlaz, line 2109: Why do we have `WebRTC-SDK (125.6422.07)`? This should now be replaced with `StreamWebRTC (125.6422.070)`.

kristian-mkd: merge conflict error, I will fix it, thanks!

packages/react-native-sdk/src/components/Livestream/LivestreamPlayer/LivestreamEnded.tsx
```diff
@@ -0,0 +1,117 @@
...
+  const call = useCall();
+  const [recordingsResponse, setRecordingsResponse] = useState<
+    ListRecordingsResponse | undefined
+  >(undefined);
+
+  const styles = useStyles();
+
+  useEffect(() => {
+    const fetchRecordings = async () => {
+      if (recordingsResponse === null) {
```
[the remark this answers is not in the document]
kristian-mkd: You are right, this wasn't executed due to the `===` check which i replaced it with `==` check so that undefined is taken into account as well.

packages/react-native-sdk/src/components/Livestream/ViewerLivestream/ViewerLobby.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+export const ViewerLobby = ({ isLive, handleJoinCall }: LobbyProps) => {
+  const styles = useStyles();
+  const { theme } = useTheme();
+  const { t } = useI18n();
+  const { useCallStartsAt, useParticipants } = useCallStateHooks();
+  const call = useCall();
+  const startsAt = useCallStartsAt();
+  const [error, setError] = useState<Error | undefined>(undefined);
+  const [countdown, setCountdown] = React.useState('');
```
myandrienko, line 24: Nit: it's better to also calculate initial value for countdown, to have it on screen immediately rather then after an effect.

packages/react-native-sdk/src/components/Livestream/ViewerLivestream/ViewerLobby.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const participants = useParticipants();
+
+  // Automatically join call when isLive becomes true
+  useEffect(() => {
+    if (isLive && handleJoinCall) {
+      handleJoinCall();
```
myandrienko, line 30: I think it's better to move this logic to `ViewerLivestream`. Right now you're rendering lobby first, then checking if the call is live, and if it's live you're joining the call and unmounting Lobby as a result. That seems a bit convoluted :)

Additionally, when we `call.join()` in an effect we also want to `call.leave()` in effect cleanup. Right now I think you're getting a warning about joining the call twice in strict mode.

sample-apps/react-native/dogfood/src/screens/LiveStream/JoinLiveStream.tsx
```diff
@@ -47,13 +50,28 @@ export const JoinLiveStream = ({
...
+    navigation.navigate('QRScanner', {
+      onScan: (id) => {
+        setCallId(id);
+      },
+    });
+  };
+
+  useEffect(() => {
+    if (scannedCallId) {
+      setCallId(scannedCallId);
```
myandrienko, line 63: You can do this outside of an effect, you'll avoid one commit that way.

https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes

kristian-mkd: As we agreed refactoring like this:
```ts
  if (callId !== scannedCallId && scannedCallId) {
    setCallId(scannedCallId);
  }
```

packages/react-native-sdk/src/components/Livestream/ViewerLivestream/ViewerLivestream.tsx
```diff
@@ -82,6 +92,41 @@ export const ViewerLivestream = ({
     hasVideo(currentSpeaker) &&
     currentSpeaker;
 
+  const useCanJoinEarly = () => {
```
myandrienko, line 95: Oh, it's not a good idea to put custom hook inside the component. Better to move it outside.

myandrienko (approved):

PR #1808, opened 2025-05-27
fix: inconsistent device state if applySettingsToStream fails

myandrienko:
### 💡 Overview

`applySettingsToStream` didn't update device status when muting/unmuting, which made it possible to leave the device in an inconsistent state.

### 📝 Implementation notes

E.g.:

1. Device is enabled
2. `applySettingsToStream` is called
3. Device is muted
4. Unmuting device fails

Device status is not `enabled`, but it's not really enabled.

Since our UI components rely on `optimisticStatus`, this change will not cause visible intermediate UI updates.


packages/client/src/devices/InputMediaDeviceManager.ts
```diff
@@ -235,10 +235,23 @@ export abstract class InputMediaDeviceManager<
...
       if (this.enabled) {
-        await this.muteStream();
-        await this.unmuteStream();
+        try {
+          await this.muteStream();
+          this.state.setStatus('disabled');
+
+          if (signal.aborted) {
+            return;
+          }
```
santhoshvai, line 246: whats the usecase of this? just to understand here.. do we ever abort it anywhere?

myandrienko: Because it's wrapped in `withCancellation`, it can be aborted by any subsequent `withCancellation` call with the same tag - in this case, by calling `enable()` or `disable()`.

oliverlaz (approved):

PR #1811, opened 2025-05-30
fix: multiple FAST reconnect attempts and improved ICE restarts

oliverlaz:
### 💡 Overview

Improves the reconnection flow and recovers from ICE connectivity loss more reliably.

📝 Spec: https://www.notion.so/stream-wiki/Improved-Reconnects-and-ICE-connection-handling-2186a5d7f9f680c29236c2c37cfa11a3
🎫 Ticket: https://linear.app/stream/issue/REACT-433/improved-reconnects-and-ice-restarts


packages/client/src/Call.ts
```diff
@@ -1423,13 +1429,37 @@ export class Call {
...
+            attempt >= 3 || // after 3 failed attempts
+            !(this.publisher?.isHealthy() ?? true) || // if the publisher is not healthy
+            !(this.subscriber?.isHealthy() ?? true); // if the subscriber is not healthy
+
+          attempt++;
+
+          const nextStrategy = shouldRejoin
+            ? WebsocketReconnectStrategy.REJOIN
+            : WebsocketReconnectStrategy.FAST;
+          this.reconnectStrategy = nextStrategy;
```
myandrienko, line 1454: Is it ok that we are possibly switching from MIGRATE to something else here? And if yes, can we rename `shouldMigrate` to something else, e.g. `wasMigrating`?

packages/client/src/rtc/BasePeerConnection.ts
```diff
@@ -252,21 +275,37 @@ export abstract class BasePeerConnection {
     const state = this.pc.iceConnectionState;
     this.logger('debug', `ICE connection state changed`, state);
 
-    if (this.state.callingState === CallingState.OFFLINE) return;
-    if (this.state.callingState === CallingState.RECONNECTING) return;
+    const { callingState } = this.state;
+    if (callingState === CallingState.OFFLINE) return;
+    if (callingState === CallingState.RECONNECTING) return;
```
myandrienko, line 325: We have this check (`callingState === CallingState.RECONNECTING`) is several places, but as far as I can see it's not very useful: it's set in `reconnect()`, but then it's immediately replaced with `JOINING` in `doJoin()`.

Is there some particular case this check guards against?

packages/client/src/rtc/BasePeerConnection.ts
```diff
@@ -252,21 +275,37 @@ export abstract class BasePeerConnection {
...
+      // as the browser might recover the connection in the meantime
+      this.logger('warn', 'disconnected connection, scheduling restartICE()');
+      clearTimeout(this.iceRestartTimeout);
+      this.iceRestartTimeout = setTimeout(() => {
+        const currentState = this.pc.iceConnectionState;
+        if (currentState === 'disconnected' || currentState === 'failed') {
+          return tryRestartIce();
+        }
+        this.logger('info', 'connection recovered, canceled restartICE()');
+      }, this.iceRestartDelay);
```
myandrienko: Is it worth it adding this timeout here, when the reconnect loop already has a timeout between attempts as well?

packages/client/src/rtc/BasePeerConnection.ts
```diff
@@ -252,21 +275,37 @@ export abstract class BasePeerConnection {
...
+
+    if (state === 'failed') {
+      // in the `failed` state, we try to restart ICE immediately
+      this.logger('warn', 'restartICE() due to failed ICE connection');
+      tryRestartIce();
+    } else if (state === 'disconnected') {
+      // in the ` disconnected ` state, we schedule a restartICE() after a delay
+      // as the browser might recover the connection in the meantime
+      this.logger('warn', 'disconnected connection, scheduling restartICE()');
+      clearTimeout(this.iceRestartTimeout);
```
myandrienko: We probably should also clear timeout when state changes to a "good" state (connected, completed) - to prevent accidentally dropping a healthy connection.

packages/client/src/Call.ts
oliverlaz: Good point. I'll rename the variable as it is misleading. In case MIGRATE fails, we should perform REJOIN always.

packages/client/src/rtc/BasePeerConnection.ts
oliverlaz: It was recently introduced. There was one issue that might not be present anymore, as we now delay the ice restart.
It is there to prevent immediate "restartICE" while reconnecting, as it will likely fail on the SFU side. Although it doesn't do any harm, I'll double-check if we can remove this line.

packages/client/src/rtc/BasePeerConnection.ts
oliverlaz: We have a safeguard inside the "setTimeout" task, but yes, we can cancel the timeout.

packages/client/src/rtc/BasePeerConnection.ts
oliverlaz: This timeout is here to shield from peer connection failures. Reconnects are triggered exclusively when the signaling WS breaks or the SFU tells us to do so. Peer Connections typically recover after an ICE restart.

PR #1821, opened 2025-06-11
feat: add audio only ringing notification support for iOS 

santhoshvai:
### 💡 Overview

Callkit gives a way to indicate audio only calls. This is supported by our backend. 

In this PR, we add it to expo plugin and dogfood app

### 📝 Implementation notes

This pull request introduces changes to handle the presence of video in incoming calls more dynamically by checking the `video` field in the incoming stream data. The updates ensure that the `hasVideo` parameter is set based on the value of the `video` field instead of being hardcoded.

* **Swift updates in `withAppDelegate.ts` and `AppDelegate.swift`:**
  - Added logic to parse the `video` field from the incoming stream data and determine the `hasVideo` value dynamically. 
  - Updated `RNCallKeep.reportNewIncomingCall` to use the dynamically determined `hasVideo` value instead of a hardcoded `true`. 

* **Objective-C updates in `withAppDelegate.ts`:**
  - Introduced logic to evaluate the `video` field and set the `hasVideo` value accordingly. 
  - Modified `RNCallKeep.reportNewIncomingCall` to utilize the dynamically calculated `hasVideo` value. (

### DIFF
--- a/packages/react-native-sdk/expo-config-plugin/src/withAppDelegate.ts
+++ b/packages/react-native-sdk/expo-config-plugin/src/withAppDelegate.ts
@@ -474,6 +474,8 @@ function addDidReceiveIncomingPushCallbackSwift(contents: string) {
     }
     
     let uuid = UUID().uuidString
+    let videoIncluded = stream["video"] as? String
+    let hasVideo = videoIncluded == "false" ? false : true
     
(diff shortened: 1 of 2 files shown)


oliverlaz (approved):
@santhoshvai, let's make sure we have this mentioned in the docs too.

PR #1823, opened 2025-06-11
wip: add dialer / ringing page

myandrienko:
### 📝 Implementation notes

### DIFF
--- a/sample-apps/react/react-dogfood/components/Dialer/DialerPage.tsx
+++ b/sample-apps/react/react-dogfood/components/Dialer/DialerPage.tsx
@@ -1,0 +1,184 @@
+import {
+  Call,
+  Notification,
+  RingingCall,
+  StreamCall,
+  StreamVideo,
+  StreamVideoClient,
+} from '@stream-io/video-react-sdk';
+import {
+  ChangeEvent,
+  FormEvent,
+  KeyboardEvent,
+  useCallback,
+  useEffect,
+  useState,
+} from 'react';
+import { useSettings } from '../../context/SettingsContext';
+import { customSentryLogger } from '../../helpers/logger';
+import { ServerSideCredentialsProps } from '../../lib/getServerSideCredentialsProps';
+import { DefaultAppHeader } from '../DefaultAppHeader';
+import { useRouter } from 'next/router';
+import { meetingId } from '../../lib/idGenerators';
+
+export const DialerPage = ({
+  apiKey,
+  user,
+  userToken,
+}: ServerSideCredentialsProps) => {
+  const {
+    settings: { language, fallbackLanguage },
+  } = useSettings();
+  const [error, setError] = useState<Error | undefined>();
+  const [loading, setLoading] = useState(false);
+  const [videoClient, setVideoClient] = useState<StreamVideoClient>();
+  const router = useRouter();
+  const callType = (router.query['type'] as string) || 'default';
+  const [userIds, setUserIds] = useState(['']);
+  const [ringingCall, setRingingCall] = useState<Call | undefined>(undefined);
+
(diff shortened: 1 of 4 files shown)


packages/client/src/Call.ts
```diff
@@ -559,10 +559,15 @@ export class Call {
...
+      throw new Error('Cannot leave call that has already been left.');
+    }
+
     await withoutConcurrency(this.joinLeaveConcurrencyTag, async () => {
       const callingState = this.state.callingState;
+
       if (callingState === CallingState.LEFT) {
-        throw new Error('Cannot leave call that has already been left.');
+        return;
       }
```
myandrienko, line 571: There used to be a possible conflict when auto drop happened roughly at the same time for caller and callee.

1. Caller auto-drops call. Caller checks calling status (it's ringing), `call.leave()` is scheduled.
2. Callee auto-drops drops, call is rejected.
3. Caller sees that everybody rejected the call. Caller checks calling status (it's ringing, because leave call is async), `call.leave()` is scheduled.
4. Second `call.leave()` throws.

We now only check for an obvious user error when scheduling `call.leave()`, but we don't throw if the call was left between scheduling and actually leaving.

packages/client/src/Call.ts
```diff
@@ -712,7 +717,6 @@ export class Call {
     this.state.setOwnCapabilities(response.own_capabilities);
 
     if (params?.ring) {
-      // the call response can indicate where the call is still ringing or not
```
myandrienko, line 715: This comment was confusing (and untrue) so I removed it.

packages/client/src/StreamVideoClient.ts
```diff
@@ -361,13 +361,23 @@ export class StreamVideoClient {
...
+    return (
+      call ??
+      new Call({
+        streamClient: this.streamClient,
+        id: id,
+        type: type,
+        clientStore: this.writeableStateStore,
+      })
+    );
   };
```
myandrienko, line 381: When working with ringing calls in particular, it's nice to be able to access the same call instance easily from different places. That's especially important since the `ringing` state is lost when creating a new call instance.

We already have the registry of all watched calls, so we can always look up call instances there. I didn't make it the default behavior since it may have unforeseen consequences.

sample-apps/react/react-dogfood/components/MeetingUI.tsx
```diff
@@ -83,7 +83,9 @@ export const MeetingUI = ({ chatClient, mode }: MeetingUIProps) => {
             ? parseInt(maxSimulcastLayers, 10)
             : undefined,
         });
-        await call.join({ create: true });
+        if (call.state.callingState !== CallingState.JOINED) {
+          await call.join({ create: true });
+        }
```
myandrienko, line 88: This allows using the `/join/[...cid]` page to connect to accepted ringing calls without joining the call twice.

sample-apps/react/react-dogfood/helpers/client.ts
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Lazily initializes video client. Credentials are captured on the first
+ * call, and ignored for subsequent calls.
+ */
+export const getClient = (
+  creds: { apiKey: string; user?: User; userToken?: string },
+  environment: AppEnvironment,
+) => {
+  if (!client) {
+    client = new StreamVideoClient({
```
myandrienko, line 24: To be able to reuse call instances (as described in https://github.com/GetStream/stream-video-js/pull/1823#discussion_r2143339218), we must have a single client instance throughout the application. So I refactored client initialization a little bit, and now the single client instance is stored globally, not in each page's local state.

PR #1840, opened 2025-06-30
feat: move audio route manager inside SDK

santhoshvai:
### 💡 Overview

For both Android and iOS, we remove the in-call manager dependency and make a similar module inside the SDK

Main support added is Audio Device Switching

Android             |  iOS
:-------------------------:|:-------------------------:
share_7361676061915555108|  image

### 📝 Implementation notes

Too large to explain, but main premise you can think of is, incallmanager implementation but inside SDK


packages/react-native-sdk/ios/StreamVideoReactNative.m
santhoshvai, line 1: no more objective C code basically, this file is only a mapper for swift here on

packages/react-native-sdk/src/components/Livestream/LivestreamLayout/LivestreamLayout.tsx
santhoshvai, line 1: not relevant to this PR, but this change is for video dimensions for livestream from #1882

packages/noise-cancellation-react-native/android/build.gradle
santhoshvai, line 1: unrelevant to this PR, but i found this small optimisation while doing webrtc snapsots myself

oliverlaz (approved):

PR #1841, opened 2025-06-30
feat: Inbound Video Pause

oliverlaz:
### 💡 Overview

Adds SDK support for notifying about forced remote (subscribed) video mute in poor network conditions.

### 📝 Implementation notes
When our SFU bandwidth estimator detects low bandwidth, it may decide to mute certain (or all) incoming video tracks.
On the recipient side, this will appear as a frozen frame. To avoid this bad UX, now we'll show the VideoPlaceholder component every time we get information that a certain track is paused.

Demo: https://getstream.slack.com/archives/C080UH1HMUK/p1751297669629389?thread_ts=1751013657.438619&cid=C080UH1HMUK

🔗 Ref: https://github.com/GetStream/protocol/pull/1174
🎫 Ticket: https://linear.app/stream/issue/REACT-485/inbound-video-pause
📑 Docs: https://github.com/GetStream/docs-content/pull/426

### DIFF
--- a/sample-apps/react-native/dogfood/ios/Podfile.lock
+++ b/sample-apps/react-native/dogfood/ios/Podfile.lock
@@ -2294,7 +2294,7 @@ PODS:
   - stream-react-native-webrtc (125.3.1):
     - React-Core
     - StreamWebRTC (~> 125.6422.070)
-  - stream-video-react-native (1.18.0):
+  - stream-video-react-native (1.18.1):
     - DoubleConversion
     - glog
     - hermes-engine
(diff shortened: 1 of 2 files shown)


packages/react-sdk/src/core/components/ParticipantView/DefaultParticipantViewUI.tsx
```diff
@@ -143,6 +149,12 @@ export const ParticipantDetails = ({
           {indicatorsVisible && !hasVideoTrack && (
             <span className="str-video__participant-details__name--video-muted" />
           )}
+          {indicatorsVisible && isTrackPaused && (
+            <span
+              title={t('Track paused due to insufficient bandwidth')}
```
myandrienko: I don't think users think about participant video as "tracks", maybe it should say "video paused" instead?

packages/client/src/events/__tests__/internal.test.ts
```diff
@@ -131,4 +134,79 @@ describe('internal events', () => {
       { userId: 'u2', sessionId: 'session-2', pin: undefined },
     ]);
   });
+
+  it('handles InboundStateNotification', () => {
+    const state = new CallState();
+    state.setSortParticipantsBy(noopComparator());
+    state.setParticipants([
+      // @ts-expect-error incomplete data
+      { sessionId: 'session-1', pausedTracks: [] },
```
myandrienko: It seems that `pausedTracks` is not initialized with `[]` in participant state, so it will be a more "honest" test if you don't initialize it here as well.

packages/react-sdk/src/core/components/ParticipantView/DefaultParticipantViewUI.tsx
oliverlaz: Fair point. I'll update the message

packages/client/src/events/__tests__/internal.test.ts
oliverlaz: yep, you are right. Will update too.

PR #1849, opened 2025-07-08
Wait contain

santhoshvai:
### 💡 Overview

In this PR, we make sure that objectFit doesn't change unnecessarily due to fetching of dimensions. We wait until dimensions are ready to mount the video component first time

### 📝 Implementation notes

- we make sure that floating participant view is always cover, because its always a small view
- and we check if dimensions are not 0
- we also remove usage of track.getSettings() for local participant

### DIFF
--- a/packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
+++ b/packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
@@ -222,6 +222,7 @@ export const CallContent = ({
                 onPressHandler={handleFloatingViewParticipantSwitch}
                 supportedReactions={supportedReactions}
                 {...participantViewProps}
+                objectFit="cover"
               />
             )}
           </View>
--- a/packages/react-native-sdk/src/hooks/useTrackDimensions.ts
+++ b/packages/react-native-sdk/src/hooks/useTrackDimensions.ts
@@ -10,48 +10,25 @@ import { useEffect, useState } from 'react';
  * Note: the `tracktype` is used only for local participants.
  * `tracktype` should be 'videoTrack' for video track and 'screenShareTrack' for screen share track.
  */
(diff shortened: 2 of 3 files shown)


packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
```diff
@@ -222,6 +222,7 @@ export const CallContent = ({
                 onPressHandler={handleFloatingViewParticipantSwitch}
                 supportedReactions={supportedReactions}
                 {...participantViewProps}
+                objectFit="cover"
```
oliverlaz:
```suggestion
                objectFit="cover"
                {...participantViewProps}
```

Why not allow customers override this prop?

packages/react-native-sdk/src/hooks/useTrackDimensions.ts
```diff
@@ -10,48 +10,25 @@ import { useEffect, useState } from 'react';
  * Note: the `tracktype` is used only for local participants.
  * `tracktype` should be 'videoTrack' for video track and 'screenShareTrack' for screen share track.
  */
-export function useTrackDimensions(
-  participant: StreamVideoParticipant,
-  trackType: VideoTrackType,
-) {
+export function useTrackDimensions(participant: StreamVideoParticipant) {
```
oliverlaz, line 10: There is an unused import now `VideoTrackType`.

packages/react-native-sdk/src/components/Call/CallContent/CallContent.tsx
santhoshvai: the floatingview is tiny, only fit for cover usage actually.. future we may need to improve this 

current behaviour is such that it only supports cover, contain is weird when you use it

PR #1850, opened 2025-07-09
feat(react-native): speech detection

kristian-mkd:
### 💡 Overview

Adds speech detection that can be used outside of call (in Lobby components) for Microphone audio checks.
 
### 📝 Implementation notes

🎫 Ticket: https://linear.app/stream/issue/XYZ-249

### DIFF
--- a/packages/react-native-sdk/src/hooks/useSpeechDetection.ts
+++ b/packages/react-native-sdk/src/hooks/useSpeechDetection.ts
@@ -1,0 +1,44 @@
+import { useEffect, useState } from 'react';
+import {
+  type SoundDetectorState,
+  RNSpeechDetector,
+  getLogger,
+} from '@stream-io/video-client';
+
+/**
+ * Hook that provides speech detection info using the RNSpeechDetector.
+ *
+ * @returns An object containing the current audio level (0 - 1) and whether sound is detected.
+ */
+export function useSpeechDetection() {
+  const [audioState, setAudioState] = useState<SoundDetectorState>({
+    isSoundDetected: false,
+    audioLevel: 0,
+  });
+
+  useEffect(() => {
+    const speechDetector = new RNSpeechDetector();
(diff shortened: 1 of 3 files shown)


packages/react-native-sdk/src/hooks/useSpeechDetection.ts
```diff
@@ -0,0 +1,46 @@
...
+    };
+
+    initSpeechDetector();
+
+    return () => {
+      if (unsubscribe) {
+        unsubscribe();
+      }
+    };
+  }, []);
```
oliverlaz: This is racy. Please create init -> deinit -> init promise chain to properly synchronize the flow.
Also, `RNSpeechDetector` should take the current mic's media stream as input.

packages/client/src/helpers/RNSpeechDetector.ts
```diff
@@ -8,16 +8,23 @@ export class RNSpeechDetector {
...
    */
   public async start(onSoundDetectedStateChanged: SoundStateChangeHandler) {
     try {
-      this.cleanupAudioStream();
-      const audioStream = await navigator.mediaDevices.getUserMedia({
-        audio: true,
-      });
-      this.audioStream = audioStream;
+      if (!this.audioStream) {
+        this.cleanupAudioStream();
```
kristian-mkd: @oliverlaz what do you think about this change? If the cleanup is not moved inside this check it cleans the `audioStream` set in the constructor.

oliverlaz: This is buggy. If you don't have an `audioStream`, what would you clean up?
Also, the opposite. If an `audioStream` is obtained from `getUserMedia`, it will not be disposed of and will cause a memory leak.

I think the `cleanup` should be done somewhere else, ideally in the `stop()` function.

PR #1852, opened 2025-07-11
feat: do not detect video dimensions through rtc stats

santhoshvai:
### 💡 Overview

Uses the  new webrtc change for getting dimensions natively without stats. 
This enables to wait for valid dimensions for setting objectFit natively as well.

Depends on: https://github.com/GetStream/react-native-webrtc/pull/14


packages/react-native-sdk/src/components/Participant/ParticipantView/VideoRenderer.tsx
```diff
@@ -25,6 +25,8 @@ const DEFAULT_VIEWPORT_VISIBILITY_STATE: Record<
   screenShareTrack: VisibilityState.UNKNOWN,
 } as const;
 
+console.log('VideoRenderer');
```
kristian-mkd: this can be removed

PR #1853, opened 2025-07-14
fix: more graceful handling of SFU join failures

oliverlaz:
### 💡 Overview

Keeps track of failed SFU join requests.
Once we fail to connect to the same SFU twice, we ask the Coordinator for another one.

📝 Spec: https://www.notion.so/stream-wiki/SFU-WebSocket-client-tracing-19d6a5d7f9f6806d8a55c4ae0d6a415c
🎫 Ticket: https://linear.app/stream/issue/REACT-488/graceful-sfu-join-failure-handling
🔗 Ref: https://github.com/GetStream/video-sfu/pull/1182


packages/client/src/Call.ts
```diff
@@ -870,6 +883,9 @@ export class Call {
 
       await sleep(retryInterval(attempt));
     }
+
+    // remove the migrating_from field from the join data
+    delete joinData.migrating_from;
```
myandrienko: Maybe I'm missing something, but what's the point of this? `joinData` is local to this `join()` method, so why bother cleaning stuff up?

oliverlaz: We re-use `joinData` for reconnects. If we keep the `migrating_from` there, on the next reconnect event, it will "mislead" the coordinator.

PR #1856, opened 2025-07-18
feat(react-native): reject call when busy

kristian-mkd:
### 💡 Overview

Adds support in the react-native SDK for automatically rejecting a call with reason `busy` when the callee is in another call, so that integrators can implement handing of the `callee busy scenario`. 

🎫 Ticket: https://linear.app/stream/issue/RN-232

📑 Docs: https://github.com/GetStream/docs-content/pull/476


sample-apps/react/react-dogfood/components/Ringing/RingingCallNotification.tsx
```diff
@@ -85,6 +88,27 @@ function RingingCallUI() {
...
+  if (isCalleeBusy && ringingCallToReject?.length === 1) {
+    try {
+      ringingCallToReject[0]?.leave({ reject: true, reason: rejectReason });
+    } catch (error) {
+      const logger = getLogger(['RingingCallUI']);
+      logger('error', 'Error rejecting Call when busy', error);
+    }
+    return null;
+  }
+
```
kristian-mkd: Duplicated as this component does not use the `RingingCallControls` but instead inline 
```jsx
            <AcceptCallButton onClick={handleAccept} />
            <CancelCallButton onClick={handleReject} />
```

packages/react-sdk/src/components/RingingCall/RingingCallControls.tsx
```diff
@@ -1,14 +1,42 @@
...
   const callCallingState = useCallCallingState();
 
   if (!call) return null;
 
+  const joinedRingingCalls = calls.filter(
+    (c) => c.ringing && c.state.callingState === CallingState.JOINED,
+  );
+  const alreadyInAnotherRingingCall = joinedRingingCalls.length > 0;
+
+  // TODO: where to add config prop shouldRejectCallWhenBusy?
```
kristian-mkd: In ReactNative we have this in the pushConfigs, not sure where to place this prop in react sdk

sample-apps/react-native/dogfood/src/components/VideoWrapper.tsx
```diff
@@ -75,6 +77,27 @@ export const VideoWrapper = ({ children }: PropsWithChildren<{}>) => {
...
+        await rejectedCall?.getOrCreate();
+
+        const isCalleeBusy =
+          rejectedCall && rejectedCall.isCreatedByMe && event.reason === 'busy';
+
+        if (isCalleeBusy) {
+          InCallManager.stop({ busytone: '_DTMF_' });
+          Alert.alert('Call rejected because user is busy.');
+        }
+      });
```
santhoshvai: I dont get it, why is this on dogfood? 

This should be on our outgoing call component in SDK

packages/react-native-sdk/src/components/Call/CallControls/IncomingCallControls.tsx
```diff
@@ -28,12 +31,38 @@ export const IncomingCallControls = ({
...
+    const isCalleeBusy =
+      alreadyInAnotherRingingCall && shouldRejectCallWhenBusy;
+
+    if (isCalleeBusy) {
+      call?.leave({ reject: true, reason: 'busy' }).catch((err) => {
+        const logger = getLogger(['IncomingCallControls']);
+        logger('error', 'Error rejecting Call when busy', err);
+      });
+    }
+  }, [calls, call, shouldRejectCallWhenBusy]);
```
santhoshvai:
```suggestion
  useEffect(() => {
    if (!shouldRejectCallWhenBusy) return;
    const ringingCallsInProgress = calls.filter(
      (c) => c.ringing && c.state.callingState === CallingState.JOINED,
    );
    const alreadyInAnotherRingingCall = ringingCallsInProgress.length > 0;

    if (alreadyInAnotherRingingCall) {
      call?.leave({ reject: true, reason: 'busy' }).catch((err) => {
        const logger = getLogger(['IncomingCallControls']);
        logger('error', 'Error rejecting Call when busy', err);
      });
    }
  }, [calls, call, shouldRejectCallWhenBusy]);
```

santhoshvai: we can exit early, to prevent the filter

packages/react-native-sdk/src/components/Call/CallControls/IncomingCallControls.tsx
```diff
@@ -28,12 +31,38 @@ export const IncomingCallControls = ({
...
+
   return (
     <View style={[styles.buttonGroup, incomingCall.buttonGroup]}>
       <RejectCallButton
         onPressHandler={onRejectCallHandler}
         size={buttonSizes.md}
-        rejectReason="decline"
+        rejectReason={'decline'}
```
santhoshvai: unnecessary change?

packages/react-native-sdk/src/components/Call/CallControls/IncomingCallControls.tsx
```diff
@@ -28,12 +31,38 @@ export const IncomingCallControls = ({
       variants: { buttonSizes },
     },
   } = useTheme();
+
+  const pushConfig = StreamVideoRN.getConfig().push;
+  const shouldRejectCallWhenBusy =
+    (Platform.OS === 'ios' && pushConfig?.ios?.shouldRejectCallWhenBusy) ||
+    (Platform.OS === 'android' &&
+      pushConfig?.android?.shouldRejectCallWhenBusy);
```
santhoshvai: why check for ios and android here??

sample-apps/react-native/dogfood/src/components/VideoWrapper.tsx
santhoshvai: showoing alert also in outgoing call also.

one of the tasks was to ask design team about what to show in our outgoing all component on busy case.

packages/react-native-sdk/src/components/Call/CallControls/IncomingCallControls.tsx
kristian-mkd: Yes, will revert this, thanks.

packages/react-native-sdk/src/components/Call/CallControls/IncomingCallControls.tsx
kristian-mkd: I am not sure I understand this. Where should this check be moved to?

sample-apps/react-native/dogfood/src/components/VideoWrapper.tsx
kristian-mkd: To keep the solution simpler, in our discussions with @oliverlaz we agreed to move this part out of the SDK. This way we leave up to the integrators how they like to handle this, what kind of UI to show and what sound to be played. Do you think we should absolutely move this to the SDK?

packages/react-native-sdk/src/components/Call/CallControls/IncomingCallControls.tsx
kristian-mkd: As per clarification I will make this prop platform agnostic.

packages/react-native-sdk/src/components/Call/RingingCallContent/RingingCallContent.tsx
```diff
@@ -70,10 +77,56 @@ const RingingCallPanel = ({
...
+    const rejectedCall = client?.call(event.call.type, callId);
+    await rejectedCall?.getOrCreate();
+
+    const isCalleeBusy =
+      rejectedCall && rejectedCall.isCreatedByMe && event.reason === 'busy';
+
+    if (isCalleeBusy) {
+      InCallManager.stop({ busytone: '_DTMF_' });
+    }
+  });
```
oliverlaz: This attaches a new event handler on every re-render. This is a big no-no!
Please move it inside a `useEffect` and have a clean-up function that will detach the event handler on component unmount.

kristian-mkd: Fixed like this: 
```ts
  useEffect(() => {
    if (!client) return;
    return client?.on('call.rejected', async (event) => {
      // Workaround needed for the busy tone:
      // This is because the call was rejected without even starting,
      // before calling the stop method with busy tone we need to start the call first.
      InCallManager.start({ media: 'audio' });

      const callCid = event.call_cid;
      const callId = callCid.split(':')[1];
      if (!callId) return;

      const rejectedCall = client?.call(event.call.type, callId);
      await rejectedCall?.getOrCreate();

      const isCalleeBusy =
        rejectedCall && rejectedCall.isCreatedByMe && event.reason === 'busy';

      if (isCalleeBusy) {
        InCallManager.stop({ busytone: '_DTMF_' });
      }
    });
  }, [client]);
```

santhoshvai: Cant we move it outside of RingingCallContent? 

I have mentioned multiple times not to add inside UI components, most people dont use RingingCallContent directly unfortunately 

You can check our StreamCall implementation, many push listeners are added there 

also please check the push implementation for more ideas where to add stuff.. 
one crucial thing is to not add anything to UI components

sample-apps/react-native/dogfood/src/navigators/Call.tsx
santhoshvai, line 1: Is this change absolutely necessary?

 it seems like we are informing integrators to handle this on their own to not show incoming call component

sample-apps/react-native/dogfood/src/utils/setPushConfig.ts
```diff
@@ -84,6 +88,12 @@ export function setPushConfig() {
     },
   });
 
+  if (Platform.OS === 'ios') {
+    NativeModules.StreamVideoReactNative?.setShouldRejectCallWhenBusy(
+      shouldRejectCallWhenBusy,
+    );
+  }
+
```
santhoshvai: why is this done from dogfood level? this should be unnecessary

the SDK must set this in native side if this config is set to true.

sample-apps/react-native/dogfood/App.tsx
```diff
@@ -188,17 +189,19 @@ const StackNavigator = () => {
...
+    if (!shouldRejectCallWhenBusy && calls.length > 1) {
       const lastCallCreatedBy = calls.at(-1)?.state.createdBy;
       Alert.alert(
         `Incoming call from ${
           lastCallCreatedBy?.name ?? lastCallCreatedBy?.id
         }, only 1 call at a time is supported`,
       );
     }
-  }, [calls]);
+  }, [calls, shouldRejectCallWhenBusy]);
```
santhoshvai: Same comment as https://github.com/GetStream/stream-video-js/pull/1856#discussion_r2256886303

This shouldnt be necessary from integrators side, we must think of a better way

kristian-mkd: I can't see a better way for this check. As this is custom behaviour on dogfood to show alert when there are multiple calls, so I want to keep this dynamic (only checked when we don't have this setting).

packages/client/src/coordinator/connection/client.ts
```diff
@@ -73,6 +73,7 @@ export class StreamClient {
   rejectConnectionId?: Function;
   private connectionIdPromiseSafe?: SafePromise<string | undefined>;
   guestUserCreatePromise?: Promise<CreateGuestResponse>;
+  shouldRejectCallWhenBusy?: boolean;
```
oliverlaz: I don't see a reason why we keep this prop on the `StreamClient` instance.
Can you move it to `StreamVideoClient`?

packages/client/src/StreamVideoClient.ts
```diff
@@ -183,6 +184,11 @@ export class StreamVideoClient {
 
     this.eventHandlersToUnregister.push(
       this.on('call.created', (event) => {
+        if (this.shouldRejectCallWhenBusy()) {
+          this.logger('info', 'Call rejected because user is busy');
+          return;
+        }
```
oliverlaz: I'm not sure if we need this here. The backend will emit `call.created` when you are added as a member. It doesn't necessarily mean the call would be a `ringing`call.

packages/client/src/StreamVideoClient.ts
```diff
@@ -230,6 +236,13 @@ export class StreamVideoClient {
             clientStore: this.writeableStateStore,
             ringing: true,
           });
+
+          if (this.shouldRejectCallWhenBusy()) {
+            await newCallInstance.reject('busy');
+            this.logger('info', 'Call rejected because user is busy');
+            return;
+          }
```
oliverlaz: We should probably have this check in this section too:
```ts
if (theCall) {
  await theCall.updateFromRingingEvent(event);
} else {
```

packages/client/src/StreamVideoClient.ts
```diff
@@ -573,4 +595,13 @@ export class StreamVideoClient {
       this.streamClient.connectAnonymousUser(user, tokenOrProvider),
     );
   };
+
+  private shouldRejectCallWhenBusy = () => {
+    if (!this.streamClient.shouldRejectCallWhenBusy) return false;
+
+    const hasJoinedCall = this.state.calls.some(
+      (c) => c.ringing && c.state.callingState === CallingState.JOINED,
```
oliverlaz: Let's flip this check. Essentially, all states except for `IDLE, LEFT, RECONNECTION_FAILED` represent a joined or joining call.

packages/react-native-sdk/src/providers/StreamCall/RejectCallWhenBusy.tsx
```diff
@@ -0,0 +1,73 @@
...
+      // This is because the call was rejected without even starting,
+      // before calling the stop method with busy tone we need to start the call first.
+      InCallManager.start({ media: 'audio' });
+
+      const callCid = event.call_cid;
+      const callId = callCid.split(':')[1];
+      if (!callId) return;
+
+      const rejectedCall = client?.call(event.call.type, callId);
+      await rejectedCall?.getOrCreate();
```
oliverlaz: The `event` already contains the `call` data. I think we can get rid of this `getOrCreate()`

```ts
event.call.created_by.id === client.state.connectedUser
```

packages/react-native-sdk/src/providers/StreamCall/RejectCallWhenBusy.tsx
```diff
@@ -0,0 +1,73 @@
...
+    client.setShouldRejectCallWhenBusy(shouldRejectCallWhenBusy ?? false);
+  }, [client, shouldRejectCallWhenBusy]);
+
+  useEffect(() => {
+    // android rejection is done in android's firebaseDataHandler
+    if (Platform.OS === 'android') return;
+    if (!shouldRejectCallWhenBusy) return;
+
+    const ringingCallsInProgress = calls.filter(
+      (c) => c.ringing && c.state.callingState === CallingState.JOINED,
```
oliverlaz: Same here, you could potentially have a call in `RECONNECTING`, `JOINING`, or `MIGRATING` state too.

packages/react-native-sdk/src/utils/push/android.ts
```diff
@@ -169,6 +169,36 @@ export const firebaseDataHandler = async (
...
     const receiver_id = data.receiver_id as string;
 
+    const shouldRejectCallWhenBusy = pushConfig.shouldRejectCallWhenBusy;
+
+    const video_client = await pushConfig.createStreamVideoClient();
+    if (video_client && shouldRejectCallWhenBusy) {
+      try {
+        const calls = video_client.state.calls;
+        const ringingCallsInProgress = calls.filter(
+          (c) => c.ringing && c.state.callingState === CallingState.JOINED,
```
oliverlaz: Here as well.

packages/react-native-sdk/src/utils/push/android.ts
```diff
@@ -169,6 +169,36 @@ export const firebaseDataHandler = async (
     const created_by_id = data.created_by_id as string;
     const receiver_id = data.receiver_id as string;
 
+    const shouldRejectCallWhenBusy = pushConfig.shouldRejectCallWhenBusy;
+
+    const video_client = await pushConfig.createStreamVideoClient();
```
oliverlaz, line 172: Are we sure this won't create multiple client instances in case an integrator implements this method with `new StreamVideoClient()`?
Can we somehow shield ourselves from that?

sample-apps/react-native/dogfood/src/components/VideoWrapper.tsx
```diff
@@ -75,6 +76,21 @@ export const VideoWrapper = ({ children }: PropsWithChildren<{}>) => {
...
+        const rejectedCall = _videoClient?.call('default', callId);
+        await rejectedCall?.getOrCreate();
+
+        const isCalleeBusy =
+          rejectedCall && rejectedCall.isCreatedByMe && event.reason === 'busy';
+
+        if (isCalleeBusy) {
+          Alert.alert('Call rejected because user is busy.');
+        }
+      });
```
oliverlaz: We leak here. Please unsubscribe this event handler on component unmount.
Also, the `event` carries the `call` data, so no need for `getOrCreate()` as well.

packages/client/src/StreamVideoClient.ts
kristian-mkd: I had to add also `RINGING` state in the flipped check.

packages/react-native-sdk/src/utils/push/android.ts
kristian-mkd: Added this method that we can use in our SDK to check the validity of the instance. 

```  /**
   * Check if this client instance is properly managed (created via getOrCreateInstance)
   */
  isValid(): void {
    const instanceKey = getInstanceKey(
      this.streamClient.key,
      this.streamClient._user!,
    );

    const managedInstance = StreamVideoClient._instances.get(instanceKey);
    if (!managedInstance) {
      throw new Error(
        'StreamVideoClient: Multiple client instances detected. ' +
          'Please use StreamVideoClient.getOrCreateInstance() in your createStreamVideoClient implementation ' +
          'to avoid potential issues with state management and WebSocket connections.',
      );
    }
  }
```

packages/client/src/StreamVideoClient.ts
```diff
@@ -573,4 +616,19 @@ export class StreamVideoClient {
...
     );
   };
+
+  private shouldRejectCall = () => {
+    if (!this.shouldRejectCallWhenBusy) return false;
+
+    const hasOngoingRingingCall = this.state.calls.some(
+      (c) =>
+        c.ringing &&
+        c.state.callingState !== CallingState.RINGING &&
```
santhoshvai:
```suggestion
```

should not be used.. if the other was incoming and another incoming call came then this would break

kristian-mkd: Every call is automatically rejected if we don't add RINGING in this check as well.

packages/react-native-sdk/src/providers/BusyTonePlayer.tsx
```diff
@@ -0,0 +1,68 @@
...
+                );
+              busyToneTimeout = undefined;
+            }, BUSY_TONE_DURATION_IN_MS);
+          })
+          .catch((error) => {
+            logger('error', 'playBusyTone failed:', error);
+          });
+      }
+      return () => {
+        StreamVideoRN.stopBusyTone();
```
oliverlaz:
```suggestion
        StreamVideoRN.stopBusyTone().catch((err) => logger('error', err));
```

PR #1862, opened 2025-07-22
fix: improved audio and video filter tracing

oliverlaz:
### 💡 Overview

Collects more trace data around enabling/disabling of Audio (Noise Cancellation) and Video (Blur, Image) filters.

### DIFF
--- a/packages/audio-filters-web/src/NoiseCancellation.ts
+++ b/packages/audio-filters-web/src/NoiseCancellation.ts
@@ -180,7 +180,10 @@ export class NoiseCancellation implements INoiseCancellation {
     const { promise: ready, resolve: filterReady } = promiseWithResolvers();
     const filterNode = await sdk.createNoiseFilter(
       this.audioContext,
-      () => filterReady(),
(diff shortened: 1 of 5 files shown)


packages/audio-filters-web/src/NoiseCancellation.ts
```diff
@@ -180,7 +180,10 @@ export class NoiseCancellation implements INoiseCancellation {
     const { promise: ready, resolve: filterReady } = promiseWithResolvers();
     const filterNode = await sdk.createNoiseFilter(
       this.audioContext,
-      () => filterReady(),
+      () => {
+        this.tracer?.trace('noiseCancellation.started', 'true');
```
kristian-mkd, line 184: maybe renamed to `noiseCancellation.enabled` as the mirror action on line 198 is `disabled`

oliverlaz: There are three states:
- enabled but not started
- started
- disabled

PR #1864, opened 2025-07-23
feat: add ability to run a custom promise option for android keep call alive service

santhoshvai:
### 💡 Overview

Currently there is no way to run any code inside our SDK's keep call alive service. But integrators have a desire to listen to call events on the background. This PR gives the ability to do so.

### 📝 Implementation notes

Added a new prop to our Config. 

📑 Docs: https://github.com/GetStream/docs-content/pull/486

### DIFF
--- a/sample-apps/react-native/dogfood/src/utils/setPushConfig.ts
+++ b/sample-apps/react-native/dogfood/src/utils/setPushConfig.ts
@@ -21,7 +21,7 @@ export function setPushConfig() {
         taskToRun: (call: Call) =>
           new Promise(() => {
             console.log(
-              'jumping to foreground service foreground service with call-cid',
+              'jumping to foreground service with call-cid',
               call.cid,
             );
           }),


kristian-mkd (approved):

packages/react-native-sdk/src/utils/StreamVideoRN/types.ts
```diff
@@ -154,6 +155,11 @@ export type StreamVideoConfig = {
         title: string;
         body: string;
       };
+      /**
+       * The task to run in the foreground service
+       * The task must resolve a promise once complete
+       */
+      taskToRun: (call: Call) => Promise<void>;
```
oliverlaz, line 162: Hey @santhoshvai, this one should be made optional. It is a breaking change otherwise

oliverlaz: handled by `deepMerge`.

PR #1886, opened 2025-08-13
fix: apply the crossorigin attribute before applying src

oliverlaz:
### 💡 Overview

Related to: https://github.com/facebook/react/issues/14035
To ensure that CORS headers are properly set, we need to apply the `crossorigin` attribute before we apply the `src` one.


myandrienko (approved):
Wild!

PR #1887, opened 2025-08-14
feat: hi-fi audio

oliverlaz:
### 📝 Implementation notes

### DIFF
--- a/packages/client/src/devices/ScreenShareManager.ts
+++ b/packages/client/src/devices/ScreenShareManager.ts
@@ -6,6 +6,7 @@ import { TrackType } from '../gen/video/sfu/models/models';
 import { getScreenShareStream } from './devices';
 import { ScreenShareSettings } from '../types';
 import { createSubscription } from '../store/rxUtils';
+import { withHifiAudio } from './hifi';
 
 export class ScreenShareManager extends InputMediaDeviceManager<
   ScreenShareState,
@@ -23,6 +24,7 @@ export class ScreenShareManager extends InputMediaDeviceManager<
 
         if (maybeTargetResolution) {
           this.setDefaultConstraints({
+            ...this.state.defaultConstraints,
             video: {
               width: maybeTargetResolution.width,
               height: maybeTargetResolution.height,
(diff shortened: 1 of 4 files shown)


packages/audio-filters-web/src/NoiseCancellation.ts
```diff
@@ -269,8 +269,13 @@ export class NoiseCancellation implements INoiseCancellation {
     if (!this.filterNode || !this.audioContext) {
       throw new Error('NoiseCancellation is not initialized');
     }
+
+    const [audioTrack] = mediaStream.getAudioTracks();
+    if (!audioTrack) throw new Error('No audio track found in the stream');
+
     const source = this.audioContext.createMediaStreamSource(mediaStream);
     const destination = this.audioContext.createMediaStreamDestination();
+    destination.channelCount = audioTrack.getSettings().channelCount || 1;
```
arnautov-anton: Can it be 0 and is that valid? I'd probably do `??` if it's `number | undefined`.

packages/client/src/devices/AudioDeviceManagerState.ts
```diff
@@ -0,0 +1,44 @@
...
+   */
+  protected constructor(
+    disableMode: TrackDisableMode,
+    permission: BrowserPermission | undefined,
+    profile: AudioBitrateProfile,
+  ) {
+    super(disableMode, permission);
+    this.audioBitrateProfileSubject = new BehaviorSubject(profile);
+    this.audioBitrateProfile$ = this.audioBitrateProfileSubject
+      .asObservable()
```
arnautov-anton, line 27:
```suggestion
```
I believe we could omit these, no? Piping should produce new observable.

oliverlaz: Fair point. I'll update.

oliverlaz: Update:
- the return type is `Observable` but the underlying implementation is still a `Subject`, so one can edit the subject data through the obseravable.

So from TS perspective, everything is good, but from data encapsulation perspective, we want `.asObservable()`

packages/audio-filters-web/src/NoiseCancellation.ts
oliverlaz: It should never be 0, but I'll update anyways

packages/client/src/devices/AudioDeviceManagerState.ts
arnautov-anton: Fair enough, thanks!

PR #1897, opened 2025-08-22
fix: handle pre ended calls on ringing push arrival

santhoshvai:
### 💡 Overview

Sometimes, when the push notification for ringing arrives and call was created a WS socket for ended would not arrive if it was ended very fast on an outgoing going by the caller (for example, WS took longer to connect or ended happened before push arrived)

In this PR, we check the ended flag whenever a get() is made for ringing calls to handle this scenario.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -399,7 +399,7 @@ export class Call {
 
     if (this.ringing) {
       // if the call is ringing, we need to register the ringing call effects
-      this.registerRingingCallEffects();
+      this.handleRingingCall();
     } else {
       // if the call is not ringing, we need to register the ringing call subscriptions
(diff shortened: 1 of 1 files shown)


packages/client/src/events/callEventHandlers.ts
```diff
@@ -66,11 +66,6 @@ export const registerEventHandlers = (call: Call, dispatcher: Dispatcher) => {
     handleRemoteSoftMute(call),
   ];
 
-  if (call.ringing) {
-    // these events are only relevant when the call is ringing
-    eventHandlers.push(registerRingingCallEventHandlers(call));
-  }
-
```
santhoshvai, line 73: this part is unnecessary, as it was only creating duplicate ringing subscriptions as we do this now in registerRingingCallEffects()

PR #1899, opened 2025-08-25
fix: capabilities and call grants

arnautov-anton:
### 💡 Overview

The baseline for this issue is; the initial state is incorrect. The user in the call does not have `screenshare` permission but has `send-audio` and `send-video` permissions. If given this "initial state" I call the `Call.updateUserPermissions` with options `{ user_id: "anton", grant_permissions: ["send-video", "send-audio" ]}` - nothing from the user perspective should change (as nothing really did). But I get two events; one from coordinator with new capabilities and one from the SFU with call grants. Logic in the `watchCallGrantsUpdated` suggests that `CallGrants` take precedence once they arrive and capabilities get re-constructed. The capabilities still come with `screenshare` in the array but grants say otherwise. Since there were two handlers that would manipulate capabilities, the logic in the `PermissionNotification` would detect change and display unexpected information.

Now this PR solves the issue only partially - there's two sources of truth where one relies on the other but (to some extent) can function withouth the other. The initial state issue has to be fixed on BE if it's even possible.

**Repro steps:**

(for the notification)
1. join a call with user A
2. revoke permission `screenshare` for that user in that call
3. grant permissions `send-audio` and `send-video` for that user in that call
4. observe that coordinator returns `own_capabilities` which include `screenshare` (that's wrong) but SFU returns call grants which say `canScreenshare: false` - whichever event comes first gets processed first and state gets set accordingly (ignoring the grants precedence if coordinator event comes after SFU event)

(for general initial state issue)
1. join a call with user A
2. revoke permissions `send-audio` and `send-video` permissions for that user in that call
5. re-join the call with user A (refresh the page)
6. buttons for mute/unmute microphone and camera are visible, clicking them results in error: `NegotiationError: participant anton is not authorized to publish track TRACK_TYPE_VIDEO`

Before:

https://github.com/user-attachments/assets/5a7b5163-8df5-4671-a63f-d8b08c9c465f

After:

https://github.com/user-attachments/assets/45023989-6392-44d0-a49b-eb8d9b456a9e

🎫 Ticket: https://linear.app/stream/issue/REACT-542


oliverlaz:
Hey @arnautov-anton, please check this with @thesyncim. The SFU should always have the correct data as this affects the publish grants.

In the SDK, we need to have two sources, where SFU takes precedence.
Of course, both Coordinator and the SFU should have the same data.

oliverlaz (approved):

PR #1902, opened 2025-08-28
Blocked participants list is not visible in CallParticipantsList

jrozbicki:
### 💡 Overview

For some reason toggle for Active/Blocked participants lists is not visible (`display: none;` in `.str-video__participant-list__content-header` class).

[Docs](https://getstream.io/video/docs/react/ui-components/participants/call-participants-list/#debouncesearchinterval) states that CallParticipantsList supports `list blocked participants`.

Currently if you block participant, there is no way of unblocking in this UI component.

### 📝 Implementation notes

Changed to `display: flex;` to make sure it's possible to see list of blocked participants and perhaps unblock them.

Before:
image

After:
image

### DIFF
--- a/packages/styling/src/CallParticipantList/CallParticipantList-layout.scss
+++ b/packages/styling/src/CallParticipantList/CallParticipantList-layout.scss
@@ -25,7 +25,7 @@ $scope-name: 'str-video__participant-list';
   }
 
   .str-video__participant-list__content-header {
-    display: none;
+    display: flex;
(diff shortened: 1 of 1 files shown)


arnautov-anton:
Hey, @jrozbicki, thank you for your contribution. 

Our components have gone through a redesign a while back and this change must've been done by mistake. I believe the change introduced in this PR makes sense so I'm approving it and merging.

PR #1939, opened 2025-09-29
fix: support scenario of accept second call when there is ongoing first call

santhoshvai:
### 💡 Overview

Currently, on RN push, we only support rejecting a incoming call when already there is an ongoing ringing type call. 

This PR makes sure that the opposite is supported. This enables rejecting the ongoing call and accepting the incoming one.


santhoshvai:
released as `@stream-io/video-react-native-sdk@1.21.1-alpha.1` for testing

PR #1944, opened 2025-10-02
chore: add useEffectEvent shim to bindings with react 19.2 dev dep

santhoshvai:
### 💡 Overview

Adds the new useEffectEvent hook to the bindings from React 19.2 and shim for older react


sample-apps/react-native/dogfood/package.json
```diff
@@ -25,7 +25,7 @@
     "@stream-io/video-filters-react-native": "workspace:^",
     "@stream-io/video-react-native-sdk": "workspace:^",
     "axios": "^1.8.1",
-    "react": "19.0.0",
+    "react": "19.2.0",
```
oliverlaz: Won't we run into compatibility issues now that RN 0.79.2 expects React 19.0.0? Same for the other RN sample apps.

PR #1947, opened 2025-10-06
fix: add useIsInPiPMode for ios

santhoshvai:
### 💡 Overview

useIsInPiPMode only worked for Android. But with this change also works for iOS.  Can be used for analytics in iOS. 

### 📝 Implementation notes

Added an `onPiPStateChange` closure to `StreamPictureInPictureController.swift`, which is called when PiP starts or stops, enabling notification of PiP state changes from native code

### DIFF
--- a/packages/react-native-sdk/src/hooks/useIsInPiPMode.tsx
+++ b/packages/react-native-sdk/src/hooks/useIsInPiPMode.tsx
@@ -1,6 +1,10 @@
 import { useEffect, useState } from 'react';
+import { Platform } from 'react-native';
 import { RxUtils } from '@stream-io/video-client';
-import { isInPiPModeAndroid$ } from '../utils/internal/rxSubjects';
+import {
+  isInPiPModeAndroid$,
+  isInPiPModeiOS$,
+} from '../utils/internal/rxSubjects';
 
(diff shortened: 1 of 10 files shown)


packages/react-native-sdk/src/hooks/useIsInPiPMode.tsx
```diff
@@ -1,23 +1,43 @@
 import { useEffect, useState } from 'react';
+import { Platform } from 'react-native';
 import { RxUtils } from '@stream-io/video-client';
-import { isInPiPModeAndroid$ } from '../utils/internal/rxSubjects';
+import {
+  isInPiPModeAndroid$,
+  isInPiPModeiOS$,
+} from '../utils/internal/rxSubjects';
```
oliverlaz: Why do we need two Subjects for the same thing? Can't we handle this universally?

santhoshvai: we can handle this universally.. just seperating here for the different internal implementation.. maybe useful in the future.. right now it isnt 

we can remove it though since its unnecessary right now

oliverlaz (approved):

PR #1948, opened 2025-10-06
fix: fallback to user_id when user.name is missing

oliverlaz:
### 💡 Overview

Fixes an edge case in Gong's integration
```
Could not upload recording for call default:m7mwRh24OfNweGN8G2Lay {
  requestId: '1tg8hsly8vytebs7xlw',
  errors: [
    'parties[4]: non company party must include at least 1 identification information'
  ]
}
```


myandrienko (approved):
Makes sense!

PR #1959, opened 2025-10-13
feat: Migrate logger to js-toolkit logger implementation

jdimovska:
### 💡 Overview

- Migrated usage of `getLogger` to new syntax where the first tag is considered as scope and the subsequent are tags. 

- Introduced new configuration option:  `logOptions?: ConfigureLoggersOptions<string>` which allows the consumers to configure the loggers per scope

- Marked `logger` and `logLevel` as deprecated and in first implementation setting them as default `level` and `level` in `configureLoggers`

### 📝 Implementation notes

📑 Docs: [https://github.com/GetStream/docs-content/pull/673]

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -30,7 +30,7 @@
     "@protobuf-ts/runtime-rpc": "^2.11.1",
     "@protobuf-ts/twirp-transport": "^2.11.1",
     "@stream-io/logger": "^2.0.0",
-    "@stream-io/worker-timer": "^1.2.4",
+    "@stream-io/worker-timer": "^1.2.5",
     "axios": "^1.12.2",
     "rxjs": "~7.8.2",
     "sdp-transform": "^2.15.0",
--- a/packages/video-filters-web/package.json
+++ b/packages/video-filters-web/package.json
@@ -27,7 +27,7 @@
     "CHANGELOG.md"
   ],
   "dependencies": {
-    "@stream-io/worker-timer": "^1.2.4",
+    "@stream-io/worker-timer": "^1.2.5",
     "wasm-feature-detect": "^1.8.0"
   },
   "devDependencies": {
--- a/yarn.lock
+++ b/yarn.lock
@@ -7197,13 +7197,6 @@ __metadata:
   languageName: unknown
   linkType: soft
 
-"@stream-io/logger@npm:^1.1.4":
-  version: 1.1.4
-  resolution: "@stream-io/logger@npm:1.1.4"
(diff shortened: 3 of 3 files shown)


packages/client/src/logger.ts
```diff
@@ -46,26 +36,25 @@ export const logToConsole: Logger = (logLevel, message, ...args) => {
...
-export const setLogLevel = (l: LogLevel) => {
-  level = l;
-};
+/**
+ * @internal
+ */
+export const getLogger = scopedLogger.getLogger<string>;
 
-export const getLogLevel = (): LogLevel => level;
+export type ScopedLogger = scopedLogger.Logger<string>;
```
arnautov-anton:
```suggestion
type AvailableScopes = "scope1" | "scope2" | "...";

export const getLogger = scopedLogger.getLogger<AvailableScopes>;

export type ScopedLogger = scopedLogger.Logger<AvailableScopes>;
```

Can we define scopes available in video or was this an intentional decision? 🤔

packages/client/src/logger.ts
```diff
@@ -46,26 +36,25 @@ export const logToConsole: Logger = (logLevel, message, ...args) => {
...
 };
 
-export const setLogger = (l: Logger, lvl?: LogLevel) => {
-  logger = l;
-  if (lvl) {
-    setLogLevel(lvl);
-  }
+export const setLogger = (
+  logger: Logger,
+  level: LogLevel,
```
arnautov-anton: It'd be good to add a function description which mentions that these two arguments set the default scope and that providing a different value in the config (third argument) might override these two, let's also deprecate the signature in favor of using the one coming from the package.

packages/client/src/StreamVideoClient.ts
```diff
@@ -94,9 +94,14 @@ export class StreamVideoClient {
...
 
     const rootLogger = clientOptions?.logger || logToConsole;
-    setLogger(rootLogger, clientOptions?.logLevel || 'warn');
 
-    this.logger = getLogger(['client']);
+    setLogger(
+      rootLogger,
+      clientOptions?.logLevel || 'warn',
+      clientOptions?.logOptions,
+    );
```
arnautov-anton: We can probably use the new signature here since it's our code.

packages/react-sdk/src/components/BackgroundFilters/BackgroundFilters.tsx
```diff
@@ -198,9 +198,8 @@ export const BackgroundFiltersProvider = (
 
   const handleError = useCallback(
     (error: any) => {
-      getLogger(['filters'])(
-        'warn',
-        'Filter encountered an error and will be disabled',
+      console.warn(
+        '[filters] Filter encountered an error and will be disabled',
       );
```
arnautov-anton, line 203: Why are we sinking logs to console here? 🤔

PR #1975, opened 2025-10-30
fix: add allowBluetoothA2DP option for audio routing too

santhoshvai:
Aligns with Swift SDK to allow A2DP also 

The behaviour will be as follows, taken from apple docs

> If this option and the allowBluetooth option are both set, when a single device supports both the Hands-Free Profile (HFP) and A2DP, the system gives hands-free ports a higher priority for routing.

so if there is only A2DP device, it will be played in mono currently with this added.. and we wont eliminate a A2DP only BT headset


santhoshvai:
Deciding against adding this.. as this wont enable mic

we will add this only to stereo output and no mic input scenarios

PR #1995, opened 2025-11-13
feat: improve performance of android video filters RN-308

santhoshvai:
### 💡 Overview

30 fps on a midrange android device (pixel 7) on 720p vp9 at ~1500kbps 🎉

previously was ~17kbps unstable in the same device for blur filter.

Also the blur quality is improved as we dont do heavy downscaling as before.

https://github.com/user-attachments/assets/a456679c-12b8-46a4-b5b5-fc515a874191

### 📝 Implementation notes

Changes that had maximum improvement:

* offloading ml model invoking to background thread
* throttling ml mode invoking as suggested in the android segmented selfie lib documentation
* removing a complex transformation matrix canvas draw in the virtual background filter

Changes with small improvements: 

* walk the mask buffer with a float array that reads the entire row at once
* minimise GC's per frame with creating reusable arrays (256x256)

🎫 Ticket: https://linear.app/stream/issue/RN-308

### DIFF
--- a/packages/video-filters-react-native/android/src/main/java/com/streamio/videofiltersreactnative/factories/BackgroundBlurFactory.kt
+++ b/packages/video-filters-react-native/android/src/main/java/com/streamio/videofiltersreactnative/factories/BackgroundBlurFactory.kt
@@ -59,7 +59,7 @@ private class BlurredBackgroundVideoFilter(
     private val isProcessing = AtomicBoolean(false)
     private var foregroundThreshold: Double = foregroundThreshold.coerceIn(0.0, 1.0)
     private var scaleBetweenSourceAndMask: Pair<Float, Float>? = null
-    private var scaleMatrix: Matrix? = null
+    private var scaleBetweenSourceAndMaskMatrix: Matrix? = null
 
     // bitmap to hold the background values, 1 for background 0 for foreground, raw mask sized
     private var backgroundMaskBitmap: Bitmap? = null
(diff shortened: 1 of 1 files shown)


packages/video-filters-react-native/android/src/main/java/com/streamio/videofiltersreactnative/common/FilterUtils.kt
```diff
@@ -1,67 +1,86 @@
...
+                    }
+                index++
+            }
+        }
+    } else {
+        for (y in 0 until segmentationMask.height) {
+            maskProbabilities.get(lineBuffer)
+            for (confidence in lineBuffer) {
+                maskArray[index] =
+                    if (confidence < confidenceThreshold) {
```
greenfrvr, line 55: @santhoshvai it looks that segment affects only condition for including/excluding pixel to the mask. does it make sense to move segment check inside lineBuffer for loop? as the rest array iterations are the same for both segment cases.

santhoshvai: i guess you mean about `segment == Segment.FOREGROUND`

I didnt move into the loop deliberately.. its an o(1) operation done 256*256 times.. so any saving is a saving here.. because this operation is done per frame.. even though its small, slightly makes sense to avoid the branching here

PR #2007, opened 2025-11-19
feat(react): Video Call moderation for React SDK

jdimovska:
### 💡 Overview

Adds automatic full-screen blur when a moderation event is received, with optional auto-removal after a set duration. Also adds a lightweight warning notification when moderation warnings are triggered.

### 📝 Implementation notes

- Introduced `useModeration` hook which applies full-screen blur when moderation event is submitted
- If full-screen blur is unsupported on the device fallback is turning off the camera  
- Implement ModerationNotification when warning event is emitted

🎫 Ticket: https://linear.app/stream/issue/REACT-625/video-call-moderation-for-react

📑 Docs: https://github.com/GetStream/docs-content/pull/773

### DIFF
--- a/packages/video-filters-web/src/BaseVideoProcessor.ts
+++ b/packages/video-filters-web/src/BaseVideoProcessor.ts
@@ -83,7 +83,7 @@ export abstract class BaseVideoProcessor {
       .pipeThrough(transformStream, { signal: this.abortController.signal })
       .pipeTo(writable, { signal: this.abortController.signal })
       .catch((e) => {
-        if (e.name !== 'AbortError') {
+        if (e.name !== 'AbortError' && e.name !== 'InvalidStateError') {
           console.error(`[${this.processorName}] Error processing track:`, e);
           this.hooks.onError?.(e);
         }


packages/react-sdk/src/hooks/useModeration.ts
```diff
@@ -0,0 +1,190 @@
...
+    return false;
+  }
+};
+
+export interface ModerationOptions {
+  /**
+   * How long the moderation effect should stay active before being disabled.
+   * Set to `0` to keep it active indefinitely.
+   */
+  moderationDuration?: number;
```
oliverlaz: Let's simplify this to `duration` only

packages/react-sdk/src/hooks/useModeration.ts
```diff
@@ -0,0 +1,190 @@
...
+    });
+  } catch {
+    return false;
+  }
+};
+
+export interface ModerationOptions {
+  /**
+   * How long the moderation effect should stay active before being disabled.
+   * Set to `0` to keep it active indefinitely.
```
oliverlaz:
```suggestion
   * Set to `0` to keep it active indefinitely. Defaults to 5000ms.
```

packages/react-sdk/src/hooks/useModeration.ts
```diff
@@ -0,0 +1,190 @@
...
+
+              await unregisterRef.current?.();
+              unregisterRef.current = null;
+              processorRef.current = null;
+
+              await handleFallback();
+              return;
+            }
+
+            processor
```
oliverlaz: Lets move `processor.start()` into the upper `try/catch` block so we don't duplicate the error handling

PR #2046, opened 2025-12-11
feat(react): Retryable call watching

jdimovska:
### 💡 Overview
Upon reconnect, the client reattempts watching active calls, but currently a failed queryCalls stops the process entirely. This PR introduces up to 3 retries with exponential backoff to ensure rewatching succeeds despite failures.

🎫 Ticket: https://linear.app/stream/issue/REACT-553/retryable-call-watching


packages/client/src/StreamVideoClient.ts
```diff
@@ -252,6 +248,45 @@ export class StreamVideoClient {
...
+    for (let attempt = 0; attempt < maxRetries; attempt++) {
+      try {
+        this.logger.info(
+          `Rewatching calls ${callsToReWatch.join(', ')} attempt ${attempt + 1}`,
+        );
+
+        await this.queryCalls({
+          watch: true,
+          filter_conditions: { cid: { $in: callsToReWatch } },
+          sort: [{ field: 'cid', direction: 1 }],
```
oliverlaz: Maybe we can get rid of `sort` here, as I don't think it is necessary for the sake of "rewatching"

PR #2056, opened 2025-12-15
fix: slow rampup on vp9/h264 codec

santhoshvai:
### 💡 Overview

Adds an optional way to  increase the start bitrate of the av1/vp9/h264 codec video encoder

### 📝 Testing

1. Enable webrtc logs in iOS 

```
    let options = WebRTCModuleOptions.sharedInstance()
    options.loggingSeverity = .info
```

2. Create call with 

```
        call.updatePublishOptions({
          dangerouslySetStartBitrateFactor: 0.7,
        });
        call.getOrCreate()
```

3. Join a call on pronto web 

4. Filter iOS logs in xcode with `SetStartBitrate` and you should see 
```(video_stream_encoder.cc:914): SetStartBitrate 1050000000```
    (assuming pronto web uses vp9 1500kbps) 

## Summary by CodeRabbit

* **New Features**
  * Added an optional configuration to control the initial video publish bitrate for supported codecs.
  * Added logic to apply a computed start-bitrate value to outgoing video offers when the option is set.

* **Tests**
  * Added unit tests covering insertion, clamping, scoping, and preservation behavior of the start-bitrate handling in SDP.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/client/src/rtc/Publisher.ts
+++ b/packages/client/src/rtc/Publisher.ts
@@ -382,22 +382,28 @@ export class Publisher extends BasePeerConnection {
         await this.pc.setLocalDescription(offer);
 
         const { sdp: baseSdp = '' } = offer;
-        const { dangerouslyForceCodec, fmtpLine } =
-          this.clientPublishOptions || {};
+        const {
+          dangerouslyForceCodec,
+          dangerouslySetStartBitrateFactor,
+          fmtpLine,
+        } = this.clientPublishOptions || {};
         let sdp = dangerouslyForceCodec
           ? removeCodecsExcept(baseSdp, dangerouslyForceCodec, fmtpLine)
           : baseSdp;
-        if (this.clientPublishOptions?.dangerouslySetStartBitrateFactor) {
-          const startBitrateFactor =
-            this.clientPublishOptions.dangerouslySetStartBitrateFactor;
+        if (dangerouslySetStartBitrateFactor) {
           this.transceiverCache.items().forEach((t) => {
             if (t.publishOption.trackType !== TrackType.VIDEO) return;
-            const maxBitrate = t.publishOption.bitrate;
+            const maxBitrateBps = t.publishOption.bitrate;
             const trackId = t.transceiver.sender.track?.id;
             if (!trackId) return;
             const mid = tracks.find((x) => x.trackId === trackId)?.mid;
             if (!mid) return;
-            sdp = setStartBitrate(sdp, maxBitrate, startBitrateFactor, mid);
+            sdp = setStartBitrate(
+              sdp,
+              maxBitrateBps / 1000, // convert to kbps
+              dangerouslySetStartBitrateFactor,
+              mid,
+            );
           });
         }
         const { response } = await this.sfuClient.setPublisher({ sdp, tracks });

### CODE REVIEW

File: packages/client/src/rtc/helpers/sdp.ts [hunk 1]
```diff
@@ -29,6 +29,42 @@ export const extractMid = (
...
+
+    for (const rtp of media.rtp) {
+      if (!targetCodecs.includes(rtp.codec.toLowerCase())) continue;
+
+      // Find or create fmtp entry for this codec payload
+      let fmtp = media.fmtp.find((f) => f.payload === rtp.payload);
+      if (!fmtp) {
+        fmtp = { payload: rtp.payload, config: '' };
+        media.fmtp.push(fmtp);
+      }
```
oliverlaz: I don't think we should create `fmtp` payloads, as that's essentially cheating the SFU that we support a codec that the client doesn't.

packages/client/src/rtc/Publisher.ts
```diff
@@ -384,9 +384,10 @@ export class Publisher extends BasePeerConnection {
         const { sdp: baseSdp = '' } = offer;
         const { dangerouslyForceCodec, fmtpLine } =
           this.clientPublishOptions || {};
-        const sdp = dangerouslyForceCodec
+        let sdp = dangerouslyForceCodec
           ? removeCodecsExcept(baseSdp, dangerouslyForceCodec, fmtpLine)
           : baseSdp;
+        sdp = setStartBitrate(sdp, 1000);
```
oliverlaz: Ideally, the `startBitrate` parameter should be a fraction of the `target_bitrate` defined in the `publishOption`. This will prevent us from running into situations where:
- target_bitrate: 500kb, start bitrate: 1000kb
- target_bitrate: 5000kb, start bitrate: 1000kb;

Maybe start with 60/70% of the target bitrate?

packages/client/src/rtc/helpers/sdp.ts
```diff
@@ -29,6 +29,42 @@ export const extractMid = (
...
+ *
+ * @param offerSdp the offer SDP to modify.
+ * @param startBitrate the start bitrate to set.
+ */
+export const setStartBitrate = (
+  offerSdp: string,
+  startBitrate: number = 1000,
+): string => {
+  const parsedSdp = parse(offerSdp);
+  const targetCodecs = ['vp9', 'h264'];
```
oliverlaz: What about av1?

packages/client/src/rtc/helpers/sdp.ts
```diff
@@ -29,6 +29,42 @@ export const extractMid = (
   return String(transceiverInitIndex);
 };
 
+/*
+ * Sets the start bitrate for the VP9 and H264 codecs in the SDP.
+ *
+ * @param offerSdp the offer SDP to modify.
+ * @param startBitrate the start bitrate to set.
+ */
+export const setStartBitrate = (
```
oliverlaz, line 38: Also, can you add a test case for this utility?

packages/client/src/rtc/Publisher.ts
```diff
@@ -384,9 +384,18 @@ export class Publisher extends BasePeerConnection {
...
           : baseSdp;
+        if (this.clientPublishOptions?.dangerouslySetStartBitrateFactor) {
+          const startBitrateFactor =
+            this.clientPublishOptions.dangerouslySetStartBitrateFactor;
+          this.transceiverCache.items().forEach((t) => {
+            if (t.publishOption.trackType !== TrackType.VIDEO) return;
+            const maxBitrate = t.publishOption.bitrate;
+            sdp = setStartBitrate(sdp, maxBitrate, startBitrateFactor);
+          });
+        }
```
oliverlaz: This is potentially wrong. `setStartBitrate` updates the `start-bitrate` of every `video` track.
This includes Screen Share video tracks as well. I think you should also provide the `mid` parameter, so you can munge only the relevant `camera video track` transceiver.

PR #2070, opened 2025-12-22
chore(react): add codemod to make useCallStateHooks React Compiler compatible

jdimovska:
### 💡 Overview
This PR introduces a new `@stream-io/codemod `package containing codemods to assist with migrating Stream Video SDK integrations across versions.
The initial codemod `use-call-state-hooks`, updates useCallStateHooks() usage to be compatible with the React Compiler.

What the codemod does:

- Renames useCallStateHooks to getCallStateHooks
- Hoists the call to module scope

### 📝 Implementation notes

🎫 Ticket: https://linear.app/stream/issue/REACT-675/enable-react-compiler

📑 Docs: https://github.com/GetStream/docs-content/pull/869

### DIFF
--- a/packages/codemod/package.json
+++ b/packages/codemod/package.json
@@ -1,6 +1,7 @@
 {
   "name": "@stream-io/video-codemod",
   "version": "0.0.1",
+  "description": "Stream Video SDK Codemods",
   "license": "See license in LICENSE",
   "bin": "./bin/cli.js",
   "scripts": {
--- a/packages/codemod/project.json
+++ b/packages/codemod/project.json
@@ -6,6 +6,7 @@
       "executor": "@jscutlery/semver:version",
       "options": {
         "dryRun": false,
+        "baseBranch": "main",
         "preset": {
           "name": "conventionalcommits",
           "preMajor": false,
@@ -13,7 +14,7 @@
             { "type": "feat", "section": "Features" },
             { "type": "fix", "section": "Bug Fixes" },
             { "type": "chore", "hidden": true },
-            { "type": "docs", "section": "Documentation" },
+            { "type": "docs", "hidden": true },
             { "type": "style", "hidden": true },
             { "type": "refactor", "hidden": true },
             { "type": "perf", "section": "Features" },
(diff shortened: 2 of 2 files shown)


packages/codemod/README.md
```diff
@@ -0,0 +1,56 @@
...
+### Usage
+
+Please carefully review all generated changes before committing them. Certain edge cases cannot be handled automatically
+and will be reported in the log output.
+
+```bash
+npx jscodeshift ./path/to/src/ \
+  --extensions=ts,tsx \
+  --parser=tsx \
+  --transform=./node_modules/@stream-io/codemod/transforms/use-call-state-hooks-transform.js
```
oliverlaz: Maybe it would be better if we ship a `bin` to abstract away this. And have something like this:
```bash
npx @stream-io/codemod use-call-state-hooks \
  --src=./path/to/src \
  --extensions=ts,tsx \
  --parser=tsx
```

Ref: https://docs.npmjs.com/cli/v11/configuring-npm/package-json#bin

packages/codemod/package.json
```diff
@@ -0,0 +1,25 @@
+{
+  "name": "@stream-io/codemod",
```
oliverlaz: The name is too generic and could clash with future codemods. 
Perhaps consider renaming this package to `@stream-io/video-codemod`?

packages/codemod/transforms/use-call-state-hooks-transform.js
```diff
@@ -0,0 +1,210 @@
...
+ *   import { useCallStateHooks as getCallStateHooks } from "@stream-io/video-react-sdk";
+ *   const { useCallCallingState, useParticipants } = getCallStateHooks();
+ *   const MyComponent = () => {
+ *     ...
+ *   };
+ */
+
+const SDK_PACKAGES = [
+  '@stream-io/video-react-sdk',
+  '@stream-io/video-react-bindings',
```
oliverlaz, line 21: Should we also include `@stream-io/video-react-native-sdk`?

oliverlaz (approved):

PR #2073, opened 2025-12-29
feat(react-native): expose useModeration hook

oliverlaz:
### 💡 Overview

Introduces new `useModeration` hook that handles the `call.moderation_blur` event.
Follow-up of: #1822

🎫 Ticket: https://linear.app/stream/issue/RN-329/moderation-video-blurring
📑 Docs: https://github.com/GetStream/docs-content/pull/373

### DIFF
--- a/packages/react-native-sdk/src/hooks/useModeration.ts
+++ b/packages/react-native-sdk/src/hooks/useModeration.ts
@@ -62,7 +62,6 @@ export const useModeration = (options?: ModerationOptions) => {
     });
     return () => {
       unsubscribe();
-      restoreRef.current?.then(() => clearTimeout(blurTimeoutRef.current));
     };
   }, [
     applyBackgroundBlurFilter,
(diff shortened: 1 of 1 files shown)


jdimovska (approved):

packages/react-native-sdk/src/hooks/useModeration.ts
```diff
@@ -0,0 +1,77 @@
...
+      const turnCameraOff = () =>
+        call.camera.disable().catch((err) => {
+          console.error(`Failed to disable camera`, err);
+        });
+
+      // not scheduling a timeout to enable the camera
+      clearTimeout(blurTimeoutRef.current);
+      if (!isSupported) return turnCameraOff();
+
+      restoreRef.current = (restoreRef.current || Promise.resolve()).then(() =>
```
greenfrvr, line 42: why not to pass `Promise.resolve()` in `userRef()` as init value?

oliverlaz: I don't want to create junk on every re-render. `useRef(Promise.resolve())` will create many dangling promises that should be garbage collected afterwards.

PR #2076, opened 2025-12-31
feat(react): Add Grid View During PIP

jdimovska:
### 💡 Overview

Adds PipLayout.Grid, a grid layout for PiP mode with built-in pagination and automatic column adjustment based on the current participant count.

### 📝 Implementation notes

🎫 Ticket: https://linear.app/stream/issue/REACT-702/improve-pip-while-presenting-add-grid-view-during-screenshare

### DIFF
--- a/packages/react-sdk/src/core/components/CallLayout/PipLayout/Grid.tsx
+++ b/packages/react-sdk/src/core/components/CallLayout/PipLayout/Grid.tsx
@@ -80,11 +80,9 @@ export const Grid = (props: PipLayoutGridProps) => {
 
   const pageCount = participantGroups.length;
 
-  useEffect(() => {
-    if (page > pageCount - 1) {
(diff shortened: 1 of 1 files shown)


packages/react-sdk/src/core/components/CallLayout/PipLayout/Grid.tsx
```diff
@@ -0,0 +1,160 @@
...
+    [participants, groupSize],
+  );
+
+  const pageCount = participantGroups.length;
+
+  useEffect(() => {
+    if (page > pageCount - 1) {
+      setPage(Math.max(0, pageCount - 1));
+    }
+  }, [page, pageCount]);
```
oliverlaz: Do we need this effect? https://react.dev/learn/you-might-not-need-an-effect

PR #2088, opened 2026-01-20
fix: add start bitrate even if there is no existing fmtp line

santhoshvai:
## Summary by CodeRabbit

* **Improvements**
  * Smarter start-bitrate calculation with clamping, min/max enforcement, and correct handling of zero-valued factors.
  * Now applies start-bitrate to additional video codecs and will create missing bitrate configuration entries when needed.
  * Added runtime validations and clearer warnings for bitrate factor inputs.

* **Tests**
  * Expanded tests covering multiple codecs, mid-scoping, and varied bitrate-factor scenarios.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/packages/client/src/rtc/Publisher.ts
+++ b/packages/client/src/rtc/Publisher.ts
@@ -391,15 +391,6 @@ export class Publisher extends BasePeerConnection {
           ? removeCodecsExcept(baseSdp, dangerouslyForceCodec, fmtpLine)
           : baseSdp;
         if (dangerouslySetStartBitrateFactor !== undefined) {
-          if (
-            dangerouslySetStartBitrateFactor < 0 ||
-            dangerouslySetStartBitrateFactor > 1
-          ) {
-            this.logger.warn(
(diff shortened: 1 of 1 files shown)


packages/client/src/rtc/Publisher.ts
```diff
@@ -390,7 +390,16 @@ export class Publisher extends BasePeerConnection {
...
-        if (dangerouslySetStartBitrateFactor) {
+        if (dangerouslySetStartBitrateFactor !== undefined) {
+          if (
+            dangerouslySetStartBitrateFactor < 0 ||
+            dangerouslySetStartBitrateFactor > 1
+          ) {
+            this.logger.warn(
+              'dangerouslySetStartBitrateFactor should be between 0 and 1, got:',
+              dangerouslySetStartBitrateFactor,
+            );
```
oliverlaz: I think this should be throwing an error.

PR #2105, opened 2026-01-30
fix: eagerly start callingx on call.join

santhoshvai:


packages/client/src/Call.ts
```diff
@@ -900,6 +900,14 @@ export class Call {
...
       throw new Error(`Illegal State: call.join() shall be called only once`);
     }
+    const callingX = globalThis.streamRNVideoSDK?.callingX;
+    if (callingX) {
+      // for Android/iOS, we need to start the call in the callingx library as soon as possible
+      if (data.ring) {
+        this.ringingSubject.next(true);
+      }
+      callingX.startCall(this);
+    }
```
oliverlaz, line 910: How would this recover from join failures?

greenfrvr (approved):

PR #2120, opened 2026-02-11
fix: allow interrupting join retries

oliverlaz:
### 💡 Overview

Allows cancelling join retries through `call.leave()`.

Up until now, `call.leave()` was waiting for any ongoing `call.join()` operations to finish successfully before disposing allocated resources.
In cases where `join()` couldn't successfully finish, and retries kick in, this meant that disposing of the `call` instance would be delayed until the retries are either exhausted or join succeeds.
While this behavior makes sense to some extent, it clashes with the device's back navigation, as there was no way of interrupting the join flow.
This PR aims to solve that by allowing us to stop the retries at any time by invoking `call.leave()`. `leave()` still waits for any ongoing `join()` process to finish or reject.

## Summary by CodeRabbit

* **New Features**
  * A new JoinCanceledError is now part of the public API.

* **Bug Fixes**
  * Join operations are now cancelable: leave() reliably cancels in-flight joins and surfaces JoinCanceledError.
  * Improved robustness of concurrent join/leave flows and safer error logging when leave fails.

* **Tests**
  * Comprehensive unit tests added for join/leave interactions, retries, and cancellation scenarios.


jdimovska (approved):

santhoshvai (approved):

oliverlaz:
Closing stale PR.

PR #2121, opened 2026-02-11
fix: improve the handling of join errors and prevent cross-socket event leaking

oliverlaz:
### 💡 Overview

Some SFU errors are preventing the `join` flow. The common case is `SFU_FULL`, which technically shouldn't be retried, as the SFU will likely reject the second attempt too. In this case, we immediately ask the coordinator for a brand new SFU, hopefully with enough capacity.

Also, we noticed that while MIGRATING, events delivered through the shutting-down client were leaking into the newly created client/peer connections. To fix that, now all events are scoped to a single tag.

🎫 Ticket: https://linear.app/stream/issue/REACT-764/improve-the-handling-of-join-errors-sfu-full

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SFU join, recovery and migration logic to reduce failed joins, interruptions, and stale migration state.

* **Refactor**
  * Event system enhanced with tag-specific and wildcard routing; event listener registration/unregistration now respects tag/wildcard semantics.

* **API**
  * Added a dedicated SFU join error type surfaced via the public errors export for clearer error classification.

* **Tests**
  * Added/updated tests validating tag-aware event routing and join/migration behaviors.

### DIFF
--- a/packages/client/src/rtc/Dispatcher.ts
+++ b/packages/client/src/rtc/Dispatcher.ts
@@ -86,6 +86,7 @@ export class Dispatcher {
 
   /**
    * Emit an event to a list of listeners.
+   *
    * @param payload the event payload to emit.
    * @param listeners the list of listeners to emit the event to.
    */


packages/client/src/StreamSfuClient.ts
```diff
@@ -541,23 +539,29 @@ export class StreamSfuClient {
...
     let timeoutId: NodeJS.Timeout | undefined = undefined;
-    const unsubscribeJoinErrorEvents = this.dispatcher.on('error', (event) => {
-      const { error, reconnectStrategy } = event;
-      if (!error) return;
-      if (reconnectStrategy === WebsocketReconnectStrategy.DISCONNECT) {
+    const unsubscribeJoinErrorEvents = this.dispatcher.on(
+      'error',
+      this.tag,
+      (event) => {
+        if (SfuJoinError.isJoinErrorCode(event)) {
```
brunoribc, line 558: Shouldn't we keep the 
`event.reconnectStrategy === WebsocketReconnectStrategy.DISCONNECT`
condition?

packages/client/src/errors/SfuJoinError.ts
```diff
@@ -0,0 +1,26 @@
...
+    super(event.error?.message || 'Join Error');
+    this.errorEvent = event;
+    this.unrecoverable =
+      event.reconnectStrategy === WebsocketReconnectStrategy.DISCONNECT;
+  }
+
+  static isJoinErrorCode(event: SfuErrorEvent): boolean {
+    const code = event.error?.code;
+    return (
+      code === ErrorCode.SFU_FULL ||
```
brunoribc, line 21: Is this a robust way of knowing it was an error from a join response? I fear that we add some other new code on the future and triggers a different functionality. For the purpose of testing I added a different error code for sfu full, and following migration path the call was not on a loop but video would not render and I would still have the older & new sub peer connection.
So two things to verify: 1. Migration (old path) works, 2. Maybe have a better way of knowing it's join rejection (if migration doesn't work well enough)

packages/client/src/StreamSfuClient.ts
oliverlaz: I removed it, as now the logic is based on the code itself. In the future, we need to make this more precise, as the current approach is a bit heuristic

packages/client/src/errors/SfuJoinError.ts
oliverlaz: No, this isn't robust at all, but at this moment, we don't have any mechanism to tell us that something went wrong.
The WebSocket connects, and we get an `error` event through it to tell us that connecting isn't allowed.
Ideally, the WS connection should be rejected in this case.

brunoribc: Hey Oliver, SFU does close the WS after sending the error event

brunoribc: Ahh I think I know what u mean, if you mean not even accepting the WS, it could be possible, but we lose context of the reason and also there is some rejections to the joinRequest that happen when reading data from it (f.e. token) so we would have two same paths

PR #2128, opened 2026-02-16
feat: callingx legacy arch support

greenfrvr:
### 💡 Overview

Added legacy arch support for callingx package.


packages/react-native-callingx/src/CallingxModule.ts
```diff
@@ -64,6 +64,7 @@ class CallingxModule implements ICallingxModule {
       return;
     }
 
+    console.log('[callingx] isTurboModuleEnabled:', isTurboModuleEnabled);
```
santhoshvai: Can we just change this to callingx logger or completely remove it maybe

greenfrvr: removed it

PR #2137, opened 2026-02-20
feat: add callingx endcall reasons from client level

santhoshvai:
### 💡 Overview

Sending proper reasons to callkit/telecom would show the appropriate states in the native view

For example, in callkit.. "call failed" - "call ended" - "call missed" are appropirately shown based on what we pass

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -942,6 +942,9 @@ export class Call {
     }
 
     await this.setup();
+
+    this.joinResponseTimeout = joinResponseTimeout;
+    this.rpcRequestTimeout = rpcRequestTimeout;
     // we will count the number of join failures per SFU.
     // once the number of failures reaches 2, we will piggyback on the `migrating_from`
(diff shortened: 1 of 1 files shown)


packages/client/src/Call.ts
```diff
@@ -631,15 +636,30 @@ export class Call {
...
+            busy: 'busy',
+            decline: 'rejected',
+          } as const;
+          const rejectReason = reason ?? 'decline';
+          const endCallReason =
+            reasonToEndCallReason[
+              rejectReason as keyof typeof reasonToEndCallReason
+            ] ?? 'rejected';
+          globalThis.streamRNVideoSDK?.callingX?.endCall(this, endCallReason);
+          await this.reject(rejectReason);
```
greenfrvr, line 649: @santhoshvai on line 635 we're already invoking `endCall` all invocation below will be ignored. I think we in that case we need to remove 635 line

santhoshvai: yeah, forgot to remove line 635, fixed now

PR #2147, opened 2026-03-02
fix(client): don't overwrite call members on call.update

jdimovska:
### 💡 Overview

This PR fixes an issue where call.update overwrites the local members state with an empty array from the backend response. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where call members were unexpectedly cleared when updating custom call properties. Members are now properly preserved during these operations.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -2415,9 +2415,8 @@ export class Call {
       UpdateCallRequest
     >(`${this.streamClientBasePath}`, updates);
 
-    const { call, members, own_capabilities } = response;
+    const { call, own_capabilities } = response;
     this.state.updateFromCallResponse(call);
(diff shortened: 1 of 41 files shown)


packages/client/src/Call.ts
```diff
@@ -2415,9 +2415,8 @@ export class Call {
       UpdateCallRequest
     >(`${this.streamClientBasePath}`, updates);
 
-    const { call, members, own_capabilities } = response;
+    const { call, own_capabilities } = response;
     this.state.updateFromCallResponse(call);
-    this.state.setMembers(members);
```
oliverlaz, line 2441: I'm not sure about this @jdimovska. This seems to be a backend bug; the SDK is implemented according to the OpenAPI contract.

PR #2148, opened 2026-03-02
fix: adjusted CXProvider delegate initialization

greenfrvr:
### 💡 Overview

This PR solves an issue with quickly answered CallKit call. The problem was that CXProvider delegate was not set by the moment CallKit answer event was omitted and was not added to delayed events store. Current implementation ensures that delegate will be set up before we report the call, so that all consequent events will be handled.

### DIFF
--- a/packages/react-native-callingx/ios/CallingxImpl.swift
+++ b/packages/react-native-callingx/ios/CallingxImpl.swift
@@ -337,6 +337,7 @@ import stream_react_native_webrtc
         
         // This is mostly needed for very first setup, as we need to override the default
         // provider configuration which is set in the constructor.
+        // IMPORTANT: We override CXProvider instance only if there is no registered call, otherwise we may lose corrsponding call state/events from CallKit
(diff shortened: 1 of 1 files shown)


packages/react-native-callingx/ios/CallingxImpl.swift
```diff
@@ -334,19 +335,29 @@ import stream_react_native_webrtc
...
         Settings.setSettings(options)
         
-        CallingxImpl.initializeIfNeeded()
-        
-        callKeepProvider = CallingxImpl.sharedProvider
-        callKeepProvider?.setDelegate(self, queue: nil)
-        
+        // This is mostly needed for very first setup, as we need to override the default
+        // provider configuration which is set in the constructor.
+        // IMPORTANT: We override CXProvider instance only if there is no registered call, otherwise we may lose corrsponding call state/events from CallKit
```
santhoshvai, line 340:
```suggestion
// we should not override provider settings when there is already a registered call
// because that would remove the provider that was associated with that call
```

PR #2152, opened 2026-03-04
fix: fgs management improvements

greenfrvr:
### 💡 Overview

This PR canges the way how ringing flow is initialized via push notification. 
We intercept firebase notification service, start FGS and register incoming call, before JS is fully loaded to match Android 10 seconds window for starting FGS from background.

Also PR includes some refactoring: 
* added global broadcast receiver which listens for call related events and emit them with event bus. This allows to handle events for cases when native module is not loaded (previously there was local broadcast receiver, which depend on native module instance).
* improvements for FGS lifecycle. previously headless task commands were invoked as `startForegroundService` which could lead to error. 
* redundant stop headless task calls are removed as for end call paths fgs will be killed with corresponding headless tasks 
* added singleton for tracked call ids tracking which allow to store data with no dependency to native module lifecycle (important for initiating call register from native pn service)
* added promise resolution for `startCall` the same manner it was done for `displayIncomingCall`. this fixes behavior of `startCallingxCall`, as we want to execute `setCurrentCallActive` synchronously after `startCall` invocation
* added pending actions management for disconnect/answer/mute actions, which fixes race condition for cases when action was requested before/during Telecom API processes call registration
* moved `rejectCallWhenBusy` and `canPostNotifications` check to native side 

## Summary by CodeRabbit

* **New Features**
  * Added Firebase Cloud Messaging integration for push notifications on incoming calls
  * Introduced stopService API to allow users to manually stop the calling service when needed

* **Bug Fixes**
  * Improved call event handling architecture with new event bus system
  * Enhanced service lifecycle management and notification delivery reliability
  * Fixed call ring notification handling to work correctly with Firebase integration

### DIFF
--- a/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
+++ b/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
@@ -475,7 +475,6 @@ class CallService : Service(), CallRepository.Listener {
             } else {
                 startForeground(CallNotificationManager.NOTIFICATION_ID, notification)
             }
-            debugLog(TAG, "[service] startForegroundSafely")
             isInForeground = true
         } catch (e: Exception) {
             // If starting the foreground service fails (for example due to background start


packages/react-native-callingx/android/src/firebase/java/io/getstream/rn/callingx/StreamMessagingService.kt
```diff
@@ -0,0 +1,50 @@
...
+ */
+class StreamMessagingService : ReactNativeFirebaseMessagingService() {
+
+  companion object {
+    const val TAG = "[Callingx] StreamMessagingService"
+  }
+
+  override fun onNewToken(token: String) {
+    super.onNewToken(token)
+  }
```
santhoshvai: can be removed?

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/StreamMessagingService.kt
```diff
@@ -0,0 +1,50 @@
...
+  override fun onNewToken(token: String) {
+    super.onNewToken(token)
+  }
+
+  override fun onMessageReceived(remoteMessage: RemoteMessage) {
+    val data = remoteMessage.data
+    debugLog(TAG, "onMessageReceived data = $data")
+
+    val isSupportedStreamVideoCallRing =
+      data["sender"] == "stream.video" && data["type"] == "call.ring"
```
santhoshvai, line 28: what happens if data had no property of "sender" here? i mean someother notification.. this is not JS so just want to check for null pointer exception

greenfrvr: it will be a check ```null == "stream.video"``` which will just return false. data is non-null, so there should be no NPE here

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallingxModuleImpl.kt
```diff
@@ -152,13 +157,30 @@ class CallingxModuleImpl(
...
-        // event and clear eat immidiate after getting initial events
+        CallEventBus.drainPendingEvents().forEach { onCallEvent(it) }
+
+        // NOTE: writable native array can be consumed only once, think of getting rid from clear
+        // event and clear it immediately after getting initial events
         val events = delayedEvents
         debugLog(TAG, "[module] getInitialEvents: Getting initial events: $events")
         delayedEvents = WritableNativeArray()
         canSendEvents = true
+        CallEventBus.markJsReady()
```
santhoshvai, line 174: we can do markJsReady on init right? or is that wrong.. this seems late to me

packages/react-native-callingx/android/src/firebase/java/io/getstream/rn/callingx/StreamMessagingService.kt
greenfrvr: removed

packages/react-native-callingx/android/build.gradle
```diff
@@ -87,4 +92,14 @@ dependencies {
...
+  // Compile-time only dependency so StreamMessagingService can reference RemoteMessage.
+  // The consuming app (via @react-native-firebase/messaging) must provide the actual runtime version.
+  compileOnly "com.google.firebase:firebase-messaging-ktx:24.1.2"
+  // Optional: use Firebase native code when the app has react-native-firebase installed (peer deps)
+  if (rootProject.findProject(':react-native-firebase_app') != null) {
+    implementation project(':react-native-firebase_app')
+  }
+  if (rootProject.findProject(':react-native-firebase_messaging') != null) {
+    implementation project(':react-native-firebase_messaging')
+  }
```
santhoshvai: we dont need the if here.. it should always be true otherwise our compilation should fail.. callingx should not be working without these two peer deps

packages/react-native-callingx/package.json
```diff
@@ -70,10 +70,20 @@
...
     "react-native": "*"
   },
+  "peerDependenciesMeta": {
+    "@react-native-firebase/app": {
+      "optional": true
+    },
+    "@react-native-firebase/messaging": {
+      "optional": true
+    }
+  },
```
santhoshvai: these are definitely not optional with callingx, it must be installed

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallingxModuleImpl.kt
greenfrvr: the idea is to get those events only when all subscriptions are setup. otherwise we may lose some events

packages/react-native-callingx/package.json
```diff
@@ -70,10 +70,20 @@
     "typescript": "^5.9.2"
   },
   "peerDependencies": {
+    "@react-native-firebase/app": "~23.0.0",
+    "@react-native-firebase/messaging": "~23.0.0",
```
santhoshvai:
```suggestion
    "@react-native-firebase/app": ">=23.0.0",
    "@react-native-firebase/messaging": ">=23.0.0",
```

we dont want to fail when v24 releases and callingx is not updated

packages/react-native-callingx/android/src/main/AndroidManifest.xml
```diff
@@ -25,5 +26,33 @@
...
+      <intent-filter>
+        <action android:name="call_registered" />
+        <action android:name="call_registered_incoming" />
+        <action android:name="call_answered" />
+        <action android:name="call_inactive" />
+        <action android:name="call_active" />
+        <action android:name="call_muted" />
+        <action android:name="call_endpoint_changed" />
+        <action android:name="call_end" />
+        <action android:name="call_registration_failed" />
```
santhoshvai: can we unify this to one place, instead of hardcoding here.. useful for future maintenace

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
```diff
@@ -52,13 +58,29 @@ class CallService : Service(), CallRepository.Listener {
...
+                    Intent(context, CallService::class.java).apply {
+                        action = ACTION_START_CALL_SERVICE
+                        putExtra(EXTRA_CALL_ID, data["call_cid"])
+                        putExtra(EXTRA_URI, data["call_cid"]?.toUri())
+                        putExtra(EXTRA_NAME, data["created_by_display_name"])
+                        putExtra(EXTRA_IS_VIDEO, data["video"] == "true")
+                    }
+
+            ContextCompat.startForegroundService(context, intent)
+        }
```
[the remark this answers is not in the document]
greenfrvr: fixed

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallingxModuleImpl.kt
```diff
@@ -152,13 +157,30 @@ class CallingxModuleImpl(
...
+        try {
+            Intent(reactApplicationContext, CallService::class.java)
+                    .apply { action = CallService.ACTION_STOP_SERVICE }
+                    .also { ContextCompat.startForegroundService(reactApplicationContext, it) }
+
+            promise.resolve(true)
+        } catch (e: Exception) {
+            Log.e(TAG, "[module] stopService: Failed to stop service: ${e.message}", e)
+            promise.reject("STOP_SERVICE_ERROR", e.message, e)
+        }
```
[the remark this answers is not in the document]
greenfrvr: fixed

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallingxModuleImpl.kt
```diff
@@ -100,15 +100,20 @@ class CallingxModuleImpl(
             }
 
     init {
+        CallEventBus.subscribe(this)
+
```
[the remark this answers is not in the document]
greenfrvr: fixed

packages/react-native-callingx/package.json
greenfrvr: fixed

packages/react-native-callingx/android/build.gradle
greenfrvr: fixed

PR #2153, opened 2026-03-04
fix(react): await initialization before enabling / disabling noise cancellation

jdimovska:
### 💡 Overview
This PR fixes a race condition where calling `setEnabled` before noise cancellation initialization completed could silently no-op because filterNode was still undefined. 

### 📝 Implementation notes

🎫 Ticket: https://linear.app/stream/issue/REACT-833/timing-issue-with-noise-cancellation-implementation-on-web
📑 Docs: https://github.com/GetStream/docs-content/pull/1065

## Summary by CodeRabbit

* **Refactor**
  * Improved noise cancellation initialization timing for more reliable enable/disable behavior.

* **New Features**
  * Exposes an "isReady" readiness state for noise cancellation so components can react to initialization.

* **Bug Fixes / UX**
  * Added a toggle for noise cancellation in the lobby UI.
  * Toggle shows a loading indicator and is disabled while noise cancellation is initializing, and reflects active/inactive state once ready.

### DIFF
--- a/packages/audio-filters-web/src/NoiseCancellation.ts
+++ b/packages/audio-filters-web/src/NoiseCancellation.ts
@@ -209,9 +209,9 @@ export class NoiseCancellation implements INoiseCancellation {
    * Enables the noise cancellation.
    */
   enable = async () => {
+    if (!this.filterNode) return;
     await this.initializing;
 
-    if (!this.filterNode) return;
     this.filterNode.enable();
     this.dispatch('change', true);
   };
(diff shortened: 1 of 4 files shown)


packages/audio-filters-web/src/NoiseCancellation.ts
```diff
@@ -209,8 +209,9 @@ export class NoiseCancellation implements INoiseCancellation {
    * Enables the noise cancellation.
    */
   enable = async () => {
-    if (!this.filterNode) return;
     await this.initializing;
+
+    if (!this.filterNode) return;
```
oliverlaz: Coderabbit's comment makes sense. Although chances are low, we can introduce a deadlock.

Instead, we can try the following:
```ts
private initializing: Promise<void> | undefined; // make it optional

init() {
  this.initializing = promiseWithResolvers();
  // ...
}
```

This way, `await this.initializing` will be a no-op in cases where `init` isn't invoked

PR #2156, opened 2026-03-09
chore: added optimistic notification updates

greenfrvr:
### 💡 Overview

This PR contains optimistic notification updates. The aim is to improve notifications interaction UX. 
Case: when push notification is shown and the app is in the killed state, user still can interact with notification, but JS logic will be executed only when JS bundle is loaded. Depending on device it make take some time after user press some action and notification changes its state. With notification optimistic updates for reject/accept actions notification will change it's state immediately, with no waiting for corresponding JS logic execution.


packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/notifications/CallNotificationManager.kt
santhoshvai, line 1: add documentation to the functions and branches in this whole file please.. this has gown in comp;exity hard to understand by intuition now

PR #2209, opened 2026-04-20
chore: ongoing calls adjustments

greenfrvr:
### 💡 Overview

Added CallKit support for non-ringing call cases for iOS to use background mode `voip` to keep call alive while the app is in background.
Excluded `livestream` type calls from being registered in CallKit. Livestream call alive is using separate FGS independent from Android Telecom. 

🎫 Ticket: https://linear.app/stream/issue/RN-379/call-alive-tweaks

📑 Docs: [https://github.com/GetStream/docs-content/pull/<id>](https://github.com/GetStream/docs-content/pull/1218)

## Release Notes

* **New Features**
  * Added iOS keep-call-alive configuration option to help maintain call stability on iOS.

* **Configuration Changes**
  * Ongoing calls feature is now configured separately per platform (iOS and Android) instead of globally.
  * Push provider configuration is now more flexible and can be omitted if not required.

* **Improvements**
  * Enhanced robustness of push configuration handling throughout the SDK with better safety checks.

### DIFF
--- a/packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
+++ b/packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
@@ -53,7 +53,7 @@ async function startForegroundService(call_cid: string) {
   const foregroundServiceConfig = videoConfig.foregroundService;
   const notificationTexts = foregroundServiceConfig.android.notificationTexts;
   const channel = foregroundServiceConfig.android.channel;
-  const smallIconName = videoConfig.push?.android.smallIcon;
+  const smallIconName = videoConfig.push?.android?.smallIcon;
 
   // NOTE: we use requestAnimationFrame to ensure that the foreground service is started after all the current UI operations are done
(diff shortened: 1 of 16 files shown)


packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
```diff
@@ -105,7 +104,10 @@ export const useAndroidKeepCallAliveEffect = () => {
     const callingx = getCallingxLibIfAvailable();
     if (
       callingx?.isSetup &&
-      (isRingingCall || (!isRingingCall && callingx?.isOngoingCallsEnabled))
+      (isRingingCall ||
+        (!isRingingCall &&
+          callingx?.isOngoingCallsEnabled &&
+          !isLivestreamCall))
```
santhoshvai:
```
          !isLivestreamCall
```

why do we have this check? I dont get it.. `!isRingingCall &&
          callingx?.isOngoingCallsEnabled` should cover all the condition, that we should not use SDK foreground service when callingx is there I think

santhoshvai (approved):

PR #2235, opened 2026-05-12
feat: pre-call testing

greenfrvr:
### 💡 Overview

This PR contains pre-call testing functionality for React Native SDK and dogfood app example how it can be used by end user. The idea of pre-call testing is to receive loopback tracks and record them to a local video file to be able to check the call from the callee's perspective.

### 📝 Implementation notes

To subscribe to their own published tracks, the caller should send `selfSubAudioVideo` during publishing. From public API `allowOwnTracksLoopback` should be passed during `join` invocation. 
When connected, loopback streams are routed to `localParticipant` (instead of default camera/microphone tracks). 

All pre-call testing functionality is provided with a single hook `useLoopbackRecording`. It allows to start/stop recording of loopback video, getting the loopback tracks refs, check local recordings directory and clear previously recorded videos.

🎫 Ticket: https://linear.app/stream/issue/RN-389/pre-call-testing

📑 Docs: https://github.com/GetStream/docs-content/pull/1325

## Summary

* **New Features**
  * Added loopback recording support (audio and optional video) via a new `useLoopbackRecording` hook.
  * Introduced native track-recording bridges on iOS and Android, exposing start/stop/clear/list recording APIs.
  * Added a dogfood “Test Recording” flow with recording, playback/results, and sharing screens.

* **Bug Fixes**
  * Improved self-subscription loopback behavior to prevent echo by muting loopback audio and tracking loopback streams.
  * Enhanced join flow options so loopback recording is applied correctly and reset on leave.


packages/client/src/helpers/client-details.ts
```diff
@@ -192,6 +192,6 @@ export const getClientDetails = async (): Promise<ClientDetails> => {
         .join(' '),
       version: '',
     },
-    webrtcVersion: browserVersion,
+    webrtcVersion: webRtcInfo?.version || '',
```
oliverlaz: Please fallback to `browserVersion` here, React Native has a different code path.

packages/client/src/rtc/Subscriber.ts
```diff
@@ -14,6 +14,14 @@ import { enableStereo, removeCodecsExcept } from './helpers/sdp';
  * @internal
  */
 export class Subscriber extends BasePeerConnection {
+  /**
+   * Remote streams received from the SFU. For a self-sub case
+   * we need to be able to distinguish between the local capture stream.
+   * The map will never contain local streams so we can safely use it to
+   * check if the stream is remote and dispose it when needed.
+   */
+  private trackedStreams: WeakSet<MediaStream> = new WeakSet();
```
oliverlaz: perf: Maybe we should lazy-init this `WeakSet`, to not waste resources when the users run the normal call flow.

packages/client/src/Call.ts
```diff
@@ -1042,11 +1074,13 @@ export class Call {
     maxJoinRetries = 3,
     joinResponseTimeout,
     rpcRequestTimeout,
+    selfSubEnabled = false,
```
oliverlaz: Shall we name this field somewhat better? I think the current name is confusing.
Maybe `allowOwnTracksLoopback`?

packages/client/src/helpers/client-details.ts
greenfrvr: done

packages/client/src/rtc/Subscriber.ts
greenfrvr: done

packages/client/src/Call.ts
greenfrvr: renamed public property to `allowOwnTracksLoopback `

oliverlaz (approved):

PR #2244, opened 2026-05-19
feat: made messaging service extendable

greenfrvr:
### 💡 Overview

When Android consumer app contains custom FirebaseMessagingService implementation (either from 3rd party SDK or own) there is a conflict as only single service can be active in runtime. This PR contains callingx package adjustments, so that several FirebaseMessagingService overrides issue can be resolved.

### 📝 Implementation notes

Added a public API on @stream-io/react-native-callingx that lets consumer apps inject ringing push notifications handling logic into custom FirebaseMessagingService implementations. 
Made StreamMessagingService extendable as an alternative for resolving conflicting messaging services issue.

🎫 Ticket: https://linear.app/stream/issue/RN-388/firebase-messaging-service-overrides-issue

📑 Docs: [https://github.com/GetStream/docs-content/pull/<id>](https://github.com/GetStream/docs-content/pull/1294)

### DIFF
--- a/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/StreamMessagingHelper.kt
+++ b/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/StreamMessagingHelper.kt
@@ -16,9 +16,17 @@ import com.google.firebase.messaging.RemoteMessage
  * ```
  * class AppMessagingService : ReactNativeFirebaseMessagingService() {
  *   override fun onMessageReceived(remoteMessage: RemoteMessage) {
- *     StreamMessagingHelper.handleMessage(applicationContext, remoteMessage)
- *     // ...forward to other push SDKs here...
- *     super.onMessageReceived(remoteMessage) // keeps RN-Firebase JS handler running
+ *     // pass remote message to React Native Firebase JS background handler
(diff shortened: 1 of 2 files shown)


packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/StreamMessagingHelper.kt
```diff
@@ -0,0 +1,59 @@
...
+ * ```
+ * class AppMessagingService : ReactNativeFirebaseMessagingService() {
+ *   override fun onMessageReceived(remoteMessage: RemoteMessage) {
+ *     StreamMessagingHelper.handleMessage(applicationContext, remoteMessage)
+ *     // ...forward to other push SDKs here...
+ *     super.onMessageReceived(remoteMessage) // keeps RN-Firebase JS handler running
+ *   }
+ * }
+ * ```
+ */
```
santhoshvai, line 33: it is not using isStreamCallRing in the example..

santhoshvai: I mean we need if else example here.. why handle in other SDKs if alerady handled by stream

santhoshvai (approved):

PR #2274, opened 2026-06-08
perf(react-native-sdk): reduce re-renders, fix listener cleanup, prune dead code

oliverlaz:
### 💡 Overview

A batch of React Native SDK performance and cleanup changes: re-render reductions across the participant and livestream render chain, an event-listener memory-leak fix, an unmount-safety guard, and dead-icon cleanup. No public API changes.

### 📝 Implementation notes

**Fixes**
- `BusyTonePlayer`: the `call.rejected` disposer was returned from the event handler, not the effect, so it never ran. Moved it to the effect body so the listener, timeout, and tone are cleaned up on unmount.
- `AppStateListener`: added a `cancelled` guard so the mount effect's `isInPiPMode().then` cannot write to the shared subject after unmount.

**Re-render / allocation reductions**
- `ParticipantView` and `VideoRenderer` memoized; `VideoRenderer`'s visibility-effect deps narrowed to the two primitives it reads.
- `CallParticipantsList`: memoized the participant prop-bag and corrected `renderItem` deps (dropped a stale-closure `eslint-disable`).
- `CallContent`: stores a remote-participant count bucket plus the first participant instead of the full array, re-rendering only on bucket-boundary crossings.
- `LivestreamLayout`: derives `objectFit` during render instead of via a renderless child effect.
- `Lobby` and `FloatingParticipantView`: memoized per-render literals; hoisted a constant lookup map.
- `useAutoEnterPiPEffect`: merged two duplicate `[disablePictureInPicture]` effects.
- `CallContent`, `CallParticipantsGrid`, and `CallParticipantsSpotlight`: `landscapeStyles` converted to `StyleSheet` constants; memoized the spotlight tile's `style`.
- `LivestreamEnded`: extracted a memoized `RecordingRow`; `useCallback`'d `renderItem`, `keyExtractor`, and `openUrl`.
- `getCallDisplayName`: collapsed `filter`/`map`/`filter` into a single `flatMap`.
- `DeviceStats`: constructs `NativeEventEmitter` inside the gated effect instead of at module load.

**Dead-code cleanup**
- Removed 5 unused icons (`LeaveStreamIcon`, `Settings`, `ShieldBadge`, `Spotlight`, `ThreeDots`) and their barrel entries.
- Moved `TopViewBackground` into the dogfood app (its only consumer, via a deep `src/icons` import) and repointed the import.

**Verification:** SDK `tsc`, `eslint`, jest (5/5 suites, 13/13), and `NODE_ENV=production yarn build` all green; dogfood `tsc` green.

🎫 Ticket: _TBD_

📑 Docs: N/A

## Release Notes

* **Bug Fixes**
  * Improved Picture-in-Picture mode handling on Android, preventing late initialization after cleanup and disabling auto-enter when requested.
  * Improved call reliability for busy tone playback and device statistics event handling.

* **Performance**
  * Reduced call UI re-renders with smarter floating-participant switching and memoized participant rendering.
  * Simplified livestream video rendering to rely on internal sizing logic.
  * Improved recordings list rendering responsiveness.

* **Chores**
  * Removed unused/deprecated icons from the SDK exports and updated the sample app to match.


packages/react-native-sdk/src/providers/BusyTonePlayer.tsx
santhoshvai, line 1: very good catch 🫡

greenfrvr (approved):

PR #2275, opened 2026-06-09
feat(client): show connecting indicator while video track is connecting

jdimovska:
### 💡 Overview
The SFU now publishes a video track right after `setPublisher` (before the first frame), so there's a window where a participant has a video track that's still connecting. This PR shows a spinner during that window - mirroring the existing
audio-connecting behaviour which clears once the track fires `unmute`. Relies on GetStream/react-native-webrtc#32 (remote audio and video tracks start muted and emit `unmute` on the first frame/buffer).

### 📝 Implementation notes

🎫 Ticket: https://linear.app/stream/issue/REACT-1009/track-muted-loading-indicator-for-video

📑 Docs: https://github.com/GetStream/docs-content/pull/1368

## Release Notes

* **New Features**
  * Added visual indicators for video and screen share connection states
  * Improved participant labels to display both audio and video connecting status during active calls


packages/react-bindings/src/hooks/callUtilHooks.ts
```diff
@@ -84,7 +87,35 @@ export const useIsAudioConnecting = (
...
+ * This hook checks if the participant has a video track and if the video track is unmuted.
+ *
+ * @param participant the participant to check.
+ * @returns true if the video track is connecting, false otherwise.
+ */
+export const useIsVideoConnecting = (
+  participant: StreamVideoParticipant,
+): boolean => {
+  return useIsTrackConnecting(participant.videoStream, hasVideo(participant));
 };
```
oliverlaz, line 159: It would be good to have helpers for all track kinds

packages/react-sdk/src/core/components/ParticipantView/DefaultParticipantViewUI.tsx
```diff
@@ -155,6 +157,12 @@ export const ParticipantDetails = ({
               tooltip={t('Audio is connecting...')}
             />
           )}
+          {indicatorsVisible && isVideoConnecting && (
+            <LoadingIndicator
+              className="str-video__participant-details__name--video-connecting"
+              tooltip={t('Video is connecting...')}
+            />
+          )}
```
oliverlaz, line 168: Should this block go after the next one? So we can have:
- spinner, spinner
- mic, spinner
- spinner, camera
- mic, camera

oliverlaz (approved):

PR #2286, opened 2026-06-16
feat(client): allow disabling client event reporter

jdimovska:
### 💡 Overview
This PR introduces a `disableClientEventReporting` client option that turns off the client-side event reporter (call-lifecycle telemetry) for a given `StreamVideoClient` instance. It defaults to `false`, so existing behavior is unchanged for all callers.
The egress composite sample app now opts out via this flag.

### 📝 Implementation notes

🎫 Ticket: https://linear.app/stream/issue/REACT-1011/call-event-reporting-option-for-disabling-reporting-for-clients

## Summary

* **New Features**
  * Added `clientEventsReportingEnabled` option to control client-side call event reporting/telemetry. When set to `false`, the client suppresses event transmission; it remains enabled by default.

* **Bug Fixes / Changes in Behavior**
  * Event delivery is now skipped when reporting is disabled.

* **Tests**
  * Added coverage to verify no event requests are sent when reporting is turned off (including an egress composite sample setup).

### DIFF
--- a/packages/client/src/coordinator/connection/types.ts
+++ b/packages/client/src/coordinator/connection/types.ts
@@ -214,6 +214,13 @@ export type StreamClientOptions = Partial<AxiosRequestConfig> & {
   browser?: boolean;
 
   /**
+   * Disables the client-side event reporter (call lifecycle telemetry).
+   * Useful for non-interactive sessions such as egress/recording where the
+   * telemetry adds no value. Defaults to `false`.
+   */
+  disableClientEventReporting?: boolean;
+
+  /**
    *  @deprecated Use `logOptions` instead.
    *  Custom logger instance used to handle log messages.
    *  Will be removed in a future release.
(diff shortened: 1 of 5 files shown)


packages/client/src/coordinator/connection/types.ts
```diff
@@ -213,6 +213,13 @@ export type StreamClientOptions = Partial<AxiosRequestConfig> & {
   baseURL?: string;
   browser?: boolean;
 
+  /**
+   * Disables the client-side event reporter (call lifecycle telemetry).
+   * Useful for non-interactive sessions such as egress/recording where the
+   * telemetry adds no value. Defaults to `false`.
+   */
+  disableClientEventReporting?: boolean;
```
oliverlaz: Let's not have "negative" settings. `!disable` reads weird :)

oliverlaz: maybe `clientEventsReportingEnabled`?

oliverlaz (approved):

PR #2310, opened 2026-06-30
fix: callingx background handling

greenfrvr:
### 💡 Overview

Fixes two Android bugs where callingx calls break when the app is backgrounded (locked screen cases) — e.g. answering a ringing call on the lock screen and then tabbing out:

  1. Audio and video drop a few seconds after backgrounding with locked screen.
  2. Call drops 2 minutes after backgrounding with locked screen. 

  Both only affect the **callingx-managed** path (ringing calls, or non-ringing calls with `enableOngoingCalls`).

### 📝 Implementation notes

Two independent root causes, two fixes:

  **1. Background mic/camera behavior**

  `CallService` ran as a `phoneCall`-only foreground service, which on Android 14+ doesn't grant microphone/camera access in the background — so the mic went silent and the camera was cut shortly after
  backgrounding.

  Fix: declare and use the `microphone`/`camera` FGS types. Since those types can't be activated from the background (the service starts from a push), `CallService` starts as `phoneCall` and is **re-promoted** to
  include mic/camera during the foreground answer window; the types then persist into the background.

  **2. JS timers suspended in background**

  Telecom + FGS with `phoneCall` keep the process alive, but without a background keep-alive task, React Native suspended JS timers while backgrounded. The client's healthcheck pings stopped, and the server dropped the participant.

  Fix: run a headless JS task on callingx's existing `CallService` (no extra notification) to keep RN/timers alive. It's shared by the ringing-push handler and the keep-call-alive hook, and stops once the call ends. `useAndroidKeepCallAliveEffect` and the push flow now use this instead of bailing out, with guards so only one keep-alive mechanism runs per call.

🎫 Ticket: https://linear.app/stream/issue/RN-404/callingx-background-with-locked-screen-issues

📑 Docs: https://github.com/GetStream/docs-content/pull/1396

## Summary by CodeRabbit

* **New Features**
  * Added Android ref-counted background keep-alive via `acquireBackgroundTask(owner)` / `releaseBackgroundTask(owner)`.
  * Foreground service now declares microphone/camera types when available, with automatic upgrading when the app returns to the foreground.
* **Bug Fixes**
  * Improved foreground/background transition handling to prevent unintended call shutdowns.
  * Enhanced cleanup for ended calls, including ringing calls triggered from push events.
* **Documentation**
  * Updated public API docs to reflect the new background keep-alive workflow.


packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
```diff
@@ -578,6 +584,76 @@ class CallService : Service(), CallRepository.Listener {
...
+     *  - the corresponding runtime permission is granted, AND
+     *  - the app currently holds while-in-use access (foreground).
+     */
+    private fun computeForegroundServiceType(): Int {
+        var type = ServiceInfo.FOREGROUND_SERVICE_TYPE_PHONE_CALL
+
+        // microphone/camera FGS types require API 30 (R) — keep phoneCall-only below it.
+        if (Build.VERSION.SDK_INT < Build.VERSION_CODES.R) {
+            return type
+        }
```
santhoshvai, line 611: this logic seems wrong.. PhoneCall is also. API 34+ https://developer.android.com/reference/android/Manifest.permission#FOREGROUND_SERVICE_PHONE_CALL

I think 33 and below, no type is needed

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
```diff
@@ -578,6 +584,76 @@ class CallService : Service(), CallRepository.Listener {
...
+        if (!LifecycleListener.isInForeground) {
+            return type
+        }
```
santhoshvai, line 614: also this if is unnecessary and wrong.. if the app is foregroud or not, the type must be same because the app can go to background after being in foreground

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
santhoshvai: This function should be Q+ only in fact

packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
```diff
@@ -101,15 +102,46 @@ export const useAndroidKeepCallAliveEffect = () => {
...
+      (isRingingCall || (!isRingingCall && callingx?.isOngoingCallsEnabled));
+
+    const isCallEnded =
+      callingState === CallingState.IDLE || callingState === CallingState.LEFT;
+
+    // Release the callingx keep-alive task when the call ends.
+    if (callingxKeepAliveOwnerRef.current && isCallEnded) {
+      const currentOwner = callingxKeepAliveOwnerRef.current;
+      callingxKeepAliveOwnerRef.current = undefined;
+      callingx?.releaseBackgroundTask(currentOwner).catch(() => {});
```
santhoshvai, line 116:
```
      callingx?.releaseBackgroundTask(currentOwner).catch(() => {});
```

we should also release on unmount of this hook.. callstate may not change always to LEFT but just component using the hook can unmount on integrator end..

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
greenfrvr: there is a difference between ServiceInfo values which are used here and Permissions values which are used in manifest. permissions were added in 34+. Service type values, were added before fgs permissions phoneCall -  29, microphone/camera - 30

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
greenfrvr: mircophone/camera types can be requested only in foreground, that's why we use re-promotion approach here. there will be crash, if we request it in background 

https://developer.android.com/develop/background-work/services/fgs/service-types

packages/react-native-sdk/src/hooks/useAndroidKeepCallAliveEffect.ts
greenfrvr: yes, it is covered on lines 195-200

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
greenfrvr: the logic is:
- if the app is in foreground we start FGS with full set of types (phoneCall/microphone/camera). 
- if the app is in background we start FGS with phoneCall type, and after app is moved to foreground we update service type to full set

santhoshvai (approved):

PR #2324, opened 2026-07-07
fix: android callingx calls should handle audio through through telecom

### LINKED ISSUE #2320: Ringing audio call always uses speakerphone and audio route can't be changed

**Describe the bug**

On Android, an **audio-only** ringing call always comes up on the **loudspeaker** on the callee, and the audio output route **cannot be switched** during the call.

This happens even though the call is configured for earpiece and the SDK computes the correct route:

- Call type: `default_device: "earpiece"`, `speaker_default_on: false`, `camera_default_on: false`
- JS SDK log: `SpeakerManager: setting default device { defaultDevice: 'earpiece' }`
- Native log: `StreamInCallManager:AudioDeviceManager Decided newAudioDevice: Earpiece`

…but the physical route stays on **speaker** for the whole call.

**To Reproduce**

1. Android app with Stream Video + `@stream-io/react-native-callingx` ringing, `enableOngoingCalls: true`.
2. Audio-only call type (`default_device: "earpiece"`, `speaker_default_on: false`).
3. Ring an Android callee and answer.
4. Audio plays on the **loudspeaker**; switching the route has no effect.

**Expected behavior**

Audio-only ringing call starts on the **earpiece**, and the user can switch the output route (earpiece / speaker / wired / Bluetooth) at runtime.

**Screenshots**

N/A — this is an audio routing issue. Relevant device logs below.

**Version**

- `@stream-io/video-react-native-sdk`: **1.40.1**
- `@stream-io/react-native-callingx`: **0.6.1**
- `@stream-io/react-native-webrtc`: **145.0.0**

**Env:**

**React Native / Expo**

- OS: Android 14 (SDK 36) — **reproduced here**; **iOS not tested**
- Device or emulator: Xiaomi `23076PC4BI` (real device, MIUI)
- React Native: **0.86.0** (New Architecture **enabled**)
- Expo: **57.0.1** (dev client)
- Ringing: `StreamVideoRN.setPushConfig` + FCM, `enableOngoingCalls: true`

**Additional context**

### Device logs (single answered audio call, callee)

```
18:18:27.501 [Callingx] displayIncomingCall: ... hasVideo=false
18:18:27.761 Telecom    CallAttributes.AUDIO_CALL --> VideoProfile.STATE_AUDIO_ONLY
18:18:28.166 Telecom    AUDIO_ROUTE, Entering state ActiveEarpieceRoute        # correct

18:18:28.717 Telecom    VideoProfile.STATE_TX_ENABLED --> CallAttributes.VIDEO_CALL   # audio call becomes "video"?
18:18:28.740 Telecom    CallAudioRouteStateMachine: USER_SWITCH_SPEAKER -> ActiveSpeakerRoute

18:18:32.020 ReactNativeJS         SpeakerManager: setting default device { defaultDevice: 'earpiece' }
18:18:35.229 AudioDeviceManager    Decided newAudioDevice: Earpiece
18:18:35.233 AudioService          setCommunicationDevice(earpiece) from uid:10334 (app)   # app, non-privileged
18:18:35.242 AudioDeviceBroker     preferredCommunicationDevice: type:speaker  mode owner pid:<telecom>  mIsPrivileged:true
             # -> app's earpiece request is ignored; route stays SPEAKER until hang-up
```

### Observations (not a confirmed diagnosis)

- The call is registered as audio (`hasVideo=false`), but Telecom logs a `VideoProfile.STATE_TX_ENABLED --> VIDEO_CALL` transition right before it switches to speaker.
- The system Telecom route owner is **privileged** (`mIsPrivileged: true`); the app-side `setCommunicationDevice(earpiece)` runs at the app uid (`mIsPrivileged: false`) and is overridden.
- On iOS this seems avoided by `shouldBypassForCallKit()`, which is iOS-only; on Android both Telecom (via callingx) and `StreamInCallManager` appear to manage audio at the same time.

We could not fix this from the app side — `callManager.android.selectAudioDevice(...)` had no effect on the route. Sharing the logs in case they help pinpoint where the speaker route is being forced.

### What we're seeing

In our app, when a listener answers an **incoming voice call** on Android, the call does **not** behave like a normal phone call:

- Audio comes out of the **loudspeaker** (speakerphone) instead of the **earpiece** (holding the phone to your ear).
- We want voice-only calls to start on the earpiece and only use the speaker if the user chooses it.
- **Changing the audio route does nothing** — if the user taps earpiece / speaker / Bluetooth in the UI, the sound stays on the loudspeaker for the whole call.

We've set our call type to use the earpiece by default (`default_device: "earpiece"`, speaker off). The SDK logs show it's *trying* to use the earpiece, but on the device the audio still plays through the speaker. This makes voice calls feel broken and too public for users who expect a private earpiece conversation.

**Reproduced on Android only. Not tested on iOS.**

santhoshvai:
### 💡 Overview

**Primarily this PR implements the advice below from Android core-telecom docs**

image

Other changes: 

* Fix: incoming calls were always added as Audio type calls, now properly reads from push data
* Feat: setPushConfig support for defaultAudioDeviceType like iOS for Android

Fixes #2320  

### 📝 Implementation notes

Incallmanager delegates Audio routing to callingx. But still supports reading state through it and calling its methods. 

## Summary by CodeRabbit

* **New Features**
  * Added Android Telecom-managed audio routing with endpoint snapshot updates via `didChangeAudioEndpoints`, including listing available endpoints and switching endpoints.
  * Introduced `defaultDeviceEndpointType` to set the starting/default audio endpoint for Telecom-managed calls; added native APIs to check Telecom mode, get registered call IDs, and manage endpoints.
* **Bug Fixes**
  * Normalized endpoint event payloads for more reliable parsing, and reduced duplicate/conflicting notifications in Telecom-managed mode; cleared stored snapshots when calls end.
  * Removed iOS route-change notifications (no Telecom audio routing on iOS).
* **Tests**
  * Added Jest coverage for Telecom-managed endpoint status and `didChangeAudioEndpoints`.

### DIFF
--- a/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallingxModuleImpl.kt
+++ b/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallingxModuleImpl.kt
@@ -1,7 +1,9 @@
 package io.getstream.rn.callingx
 
 import android.content.Intent
+import android.os.Build
 import android.os.Bundle
+import android.os.ParcelUuid
 import android.telecom.DisconnectCause
 import android.util.Log
 import androidx.core.content.ContextCompat
@@ -34,6 +36,7 @@ class CallingxModuleImpl(
         const val EXTRA_ON_HOLD = "hold"
         const val EXTRA_DISCONNECT_CAUSE = "disconnect_cause"
         const val EXTRA_AUDIO_ENDPOINT = "audio_endpoint"
+        const val EXTRA_AUDIO_ENDPOINTS_SNAPSHOT = "audio_endpoints_snapshot"
         const val EXTRA_SOURCE = "source"
         const val EXTRA_ACTION = "action_name"
 
@@ -45,6 +48,7 @@ class CallingxModuleImpl(
         const val CALL_ACTIVE_ACTION = "io.getstream.CALL_ACTIVE"
         const val CALL_MUTED_ACTION = "io.getstream.CALL_MUTED"
         const val CALL_ENDPOINT_CHANGED_ACTION = "io.getstream.CALL_ENDPOINT_CHANGED"
+        const val CALL_AUDIO_ENDPOINTS_CHANGED_ACTION = "io.getstream.CALL_AUDIO_ENDPOINTS_CHANGED"
         const val CALL_END_ACTION = "io.getstream.CALL_END"
         const val CALL_REGISTRATION_FAILED_ACTION = "io.getstream.CALL_REGISTRATION_FAILED"
         const val CALL_OPTIMISTIC_ACCEPT_ACTION = "io.getstream.ACCEPT_CALL_OPTIMISTIC"
@@ -72,6 +76,7 @@ class CallingxModuleImpl(
 
         LifecycleListener.unregister()
         CallRegistrationStore.clearAll()
+        AudioEndpointStore.clearAll()
         CallEventBus.unsubscribe(this)
 
         isModuleInitialized = false
@@ -192,12 +197,7 @@ class CallingxModuleImpl(
 
     fun answerIncomingCall(callId: String, promise: Promise) {
         debugLog(TAG, "[module] answerIncomingCall: Answering call: $callId")
-        // TODO: get the call type from the call attributes
-        val isAudioCall = true // TODO: get the call type from the call attributes
-        // registeredCall.callAttributes.callType ==
-        //         CallAttributesCompat.CALL_TYPE_AUDIO_CALL
-        // currentCall?.processAction(TelecomCallAction.Answer(isAudioCall))
-        executeServiceAction(callId, CallAction.Answer(isAudioCall), promise)
+        executeServiceAction(callId, CallAction.Answer, promise)
     }
 
     fun startCall(
@@ -289,6 +289,35 @@ class CallingxModuleImpl(
 
     fun hasRegisteredCall(): Boolean {
         return CallRegistrationStore.hasRegisteredCall()
+    }
+
+    /** Android backs Telecom audio routing only when the Jetpack Telecom repository is used (API 26+). */
+    fun isTelecomBacked(): Boolean = Build.VERSION.SDK_INT >= Build.VERSION_CODES.O
+
+    fun getAvailableAudioEndpoints(callId: String, promise: Promise) {
+        promise.resolve(AudioEndpointStore.getSnapshot(callId))
+    }
+
+    fun requestAudioEndpointChange(callId: String, endpointId: String, promise: Promise) {
+        debugLog(TAG, "[module] requestAudioEndpointChange: $callId -> $endpointId")
+        val parcelUuid =
+                try {
+                    ParcelUuid.fromString(endpointId)
+                } catch (e: IllegalArgumentException) {
+                    promise.reject("INVALID_ENDPOINT_ID", "Invalid endpoint id: $endpointId", e)
+                    return
+                }
+        executeServiceAction(callId, CallAction.SwitchAudioEndpoint(parcelUuid), promise)
+    }
+
+    fun setDefaultAudioDeviceEndpointType(endpointType: String?) {
+        debugLog(TAG, "[module] setDefaultAudioDeviceEndpointType: $endpointType")
+        AudioEndpointStore.setDefaultEndpointPref(endpointType)
+        SettingsStore.setDefaultDeviceEndpointType(reactApplicationContext, endpointType)
+    }
+
+    fun getRegisteredCallIds(): WritableArray {
+        return Arguments.fromList(CallRegistrationStore.getTrackedCallIds())
     }
 
     fun setMutedCall(callId: String, isMuted: Boolean, promise: Promise) {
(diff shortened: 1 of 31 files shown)


packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/repo/CallRepository.kt
```diff
@@ -37,6 +38,11 @@ abstract class CallRepository(protected val context: Context) {
     fun onCallRegistered(callId: String, incoming: Boolean)
     fun onMuteCallChanged(callId: String, isMuted: Boolean)
     fun onCallEndpointChanged(callId: String, endpoint: String)
+
+    /**
+     * Fired when the current endpoint or the set of available endpoints changes.
+     */
+    fun onCallAudioEndpointsChanged(callId: String) {}
```
greenfrvr, line 44: why can't we use existing callback `onCallEndpointChanged`? it is not really used anywhere, it was implemented to mirror ios implemention. we could reuse it for new real logic, or we can just remove it, if we have a new path with real logic

packages/react-native-sdk/src/modules/call-manager/CallManager.ts
```diff
@@ -23,6 +75,14 @@ class AndroidCallManager {
    */
   getAudioDeviceStatus = async (): Promise<AudioDeviceStatus> => {
     invariant(Platform.OS === 'android', 'Supported only on Android');
+    if (isAndroidTelecomManaged()) {
+      const callId = getTelecomCallId();
+      if (callId && CallingxModule) {
```
greenfrvr: `CallingxModule` check is redundant here and below. if it is `false` `isAndroidTelecomManaged` will guard that path

packages/react-native-sdk/src/utils/internal/callingx/callingx.ts
```diff
@@ -15,6 +15,47 @@ import { CallingState, videoLoggerSystem } from '@stream-io/video-client';
...
+    return;
+  }
+  const logger = videoLoggerSystem.getLogger('callingx');
+  logger.warn(
+    'Telecom registration failed; falling back to classic StreamInCallManager audio',
+  );
+  try {
+    StreamInCallManagerNativeModule.stop();
+    StreamInCallManagerNativeModule.setTelecomManagedMode(false);
+    StreamInCallManagerNativeModule.setup();
```
greenfrvr: `setup` invocation is redundant here. `start` invokes it internally.

packages/react-native-sdk/android/src/main/java/com/streamvideo/reactnative/audio/AudioDeviceManager.kt
```diff
@@ -114,12 +122,14 @@ class AudioDeviceManager(
...
+        if (!telecomManagedMode) {
+            if (callAudioRole == CallAudioRole.Communicator) {
+                mAudioManager.mode = AudioManager.MODE_IN_COMMUNICATION
+            } else {
+                // Audio routing is handled automatically by the system in normal media mode
+                // and bluetooth microphones may not work on some devices.
+                mAudioManager.mode = AudioManager.MODE_NORMAL
+            }
         }
         audioFocusUtil.setup(callAudioRole, mReactContext)
```
greenfrvr, line 134: in `start` method this line is guarded by `!telecomManagedMode`. looks inconsistent

packages/react-native-sdk/src/modules/call-manager/CallManager.ts
santhoshvai: redundant yes.. but was there for TS to be happy

now I moved it to a common helper

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/repo/CallRepository.kt
santhoshvai: removed the older code

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallingxModuleImpl.kt
```diff
@@ -291,6 +291,35 @@ class CallingxModuleImpl(
...
+
+    fun setDefaultAudioDeviceEndpointType(endpointType: String?) {
+        debugLog(TAG, "[module] setDefaultAudioDeviceEndpointType: $endpointType")
+        AudioEndpointStore.setDefaultEndpointPref(endpointType)
+        SettingsStore.setDefaultDeviceEndpointType(reactApplicationContext, endpointType)
+    }
+
+    fun getRegisteredCallIds(): WritableArray {
+        return Arguments.fromList(CallRegistrationStore.getTrackedCallIds())
+    }
```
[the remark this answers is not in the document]
santhoshvai: fixed it, now we remove the call from the store at the single terminal state every end path passes through — the Call.None/Unregistered branch in CallService.kt:351 — rather than patching the source-specific handler. That covers app, system, and remote ends uniformly (removeTrackedCall is idempotent, so the existing source == "app" removal remains harmless).

greenfrvr (approved):

PR #2332, opened 2026-07-13
fix: use speakerLayoutSortPreset in RN spotlight view

santhoshvai:
### 💡 Overview

the RN SDK spotlight view, did not prioritise the dominant speaker. Which is not expected in spotlight view. This PR fixes.

### 📝 Implementation notes

Port the useSpeakerLayoutSortPreset from React SDK

## Summary by CodeRabbit

* **New Features**
  * Improved participant ordering behavior for calls, including special handling for one-on-one layouts.
  * Added participant sorting support for paginated layouts.
  * Exposed a way to read the currently active participant-sorting configuration.
* **Bug Fixes**
  * Restores the prior participant ordering after temporary layout-specific sorting is no longer needed.


packages/react-native-sdk/src/hooks/useSpeakerLayoutSortPreset.ts
```diff
@@ -0,0 +1,34 @@
...
+  useEffect(() => {
+    if (!call) return;
+    // always show the remote participant in the spotlight
+    if (isOneOnOneCall) {
+      call.setSortParticipantsBy(combineComparators(screenSharing, loggedIn));
+    } else {
+      call.setSortParticipantsBy(speakerLayoutSortPreset);
+    }
+    return () => {
+      resetSortPreset(call);
```
oliverlaz: Maybe capture the possibly integration-side overridden comparator? We don't have guarantees that integrators set it to the `CallTypes` registry

packages/react-sdk/src/core/components/CallLayout/hooks.ts
```diff
@@ -100,14 +100,15 @@ export const useSpeakerLayoutSortPreset = (
 ) => {
   useEffect(() => {
     if (!call) return;
+    const previousSortPreset = call.getSortParticipantsBy();
     // always show the remote participant in the spotlight
     if (isOneOnOneCall) {
       call.setSortParticipantsBy(combineComparators(screenSharing, loggedIn));
     } else {
       call.setSortParticipantsBy(speakerLayoutSortPreset);
```
oliverlaz:
```suggestion
      call.setSortParticipantsBy(previousSortPreset);
```

Shouldn't we reset to the predefined preset here?

santhoshvai: no we should not.. it should be speakerLayoutSortPreset here

oliverlaz (approved):

PR #2334, opened 2026-07-15
feat: add iOS audio output selection methods

santhoshvai:
### 💡 Overview

Adds a single, cross-platform audio output device picker for the React Native SDK.
One API — `callManager.audioDevices` and the `useAudioDeviceStatus()` hook

### 📝 Implementation notes

**Changed method names and behaviour:**
- `callManager.android` → `callManager.audioDevices`.
- `getAudioDeviceStatus()` → `getStatus()`; `selectAudioDevice(name)` → `select(deviceId)`;
  `addAudioDeviceChangeListener()` → `addChangeListener()`.
- `AudioDeviceStatus` type removed → use `AudioDevicesState` + `AudioDevice`.
- Selection is now by stable **id**, not display **name**.

📑 Docs: https://github.com/GetStream/docs-content/pull/1442

## Summary by CodeRabbit

* **New Features**
  * Added a cross-platform audio device picker and real-time device state hook via `useAudioDeviceStatus`.
  * Introduced the `didChangeAudioRoute` event for audio-route updates.
* **Bug Fixes**
  * Improved Bluetooth audio routing/selection by using stable endpoint/device IDs instead of names.
  * Prevented stale route monitoring by managing the iOS route observer lifecycle more reliably.
* **API / Behavior Updates**
  * Audio device APIs/state now use `AudioDevicesState`/`AudioDevice` with `selectedDeviceId`.
  * Native/bridge methods now select endpoints by device ID, plus added `reapplyAudioRoute`.
  * Removed the prior “audio endpoints snapshot” event payload behavior; route events are now used instead.
* **Tests**
  * Updated Android Telecom-related tests for the new `AudioDevicesState` shape.


packages/react-native-callingx/src/types.ts
```diff
@@ -356,6 +356,7 @@ export type EventName =
   | 'didDisplayIncomingCall'
   | 'didToggleHoldCallAction'
   | 'didChangeAudioEndpoints'
+  | 'didChangeAudioRoute'
```
greenfrvr, line 358: what is the difference between `didChangeAudioEndpoints` and `didChangeAudioRoute` events? do we really need separate events for each platform here?

packages/react-native-sdk/src/modules/call-manager/types.ts
```diff
@@ -6,10 +6,35 @@ export type AudioDeviceEndpointType =
...
+
+/**
+ * The current audio-output device state. Returned by
+ * `callManager.audioDevices.getStatus()` and the `useAudioDeviceStatus()` hook.
+ */
+export type AudioDevicesState = {
+  /** All currently available output devices. */
+  devices: AudioDevice[];
+  /** The id of the active device, matching one of `devices[].id` (when known). */
+  selectedDeviceId?: string;
```
greenfrvr, line 37: is there a difference between selected device and current endpoint? it may have sense to use consistent naming here

packages/react-native-sdk/src/hooks/useAudioDeviceStatus.ts
```diff
@@ -0,0 +1,39 @@
...
+ */
+export const useAudioDeviceStatus = (): AudioDevicesState | undefined => {
+  const [state, setState] = useState<AudioDevicesState>();
+
+  useEffect(() => {
+    let active = true;
+    callManager.audioDevices.getStatus().then((s) => {
+      if (active) setState(s);
+    });
+    const unsubscribe = callManager.audioDevices.addChangeListener(setState);
```
greenfrvr: it makes sense to do some memoization here. `onChange` callback might be called several times during single route change which will lead to redundant re-renders

santhoshvai: thanks, fixed this.. now introduced an explicit check for setstate duplication

packages/react-native-callingx/src/types.ts
santhoshvai: it existed because telecom had a different event and shape of the data.. now I have simplified this by just unifying to one event and changing the shape in JS side

fixed!

packages/react-native-sdk/src/modules/call-manager/types.ts
santhoshvai: yes, its intended for bluetooth devices.. like this 

```
selectedDeviceId: "1234",  -- for showing which device is selected 
currentEndpointType: "Bluetooth" -- for displaying icon 
```

there can be multiple bluetooth devices.. so id matters there.. for other types there can be only one, like one wired, one speaker

packages/react-native-sdk/src/modules/call-manager/CallManager.ts
```diff
@@ -71,95 +71,147 @@ const endpointTypeToDisplayName = (
...
+   */
+  private ensureInterruptionReassert = () => {
+    if (this.interruptionReassertSetup) return; // subscibed once per app lifetime
+    if (Platform.OS !== 'ios' || !CallingxModule) return;
+    this.interruptionReassertSetup = true;
+    CallingxModule.addEventListener('didAudioInterruption', (event) => {
+      if (event.phase === 'ended') {
+        NativeManager.reapplyAudioRoute();
+      }
+    });
```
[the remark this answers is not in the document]
santhoshvai: @coderabbitai review

greenfrvr (approved):

PR #2344, opened 2026-07-26
fix(client): fix stale mid in SetPublisher after PeerConnection rollback

jdimovska:
### 💡 Overview
This PR fixes the publisher sending the SFU a `mid` that doesn't exist in the SDP it was sent with. The announced tracks are read before `setLocalDescription()`. At that point `transceiver.mid` is `null`, so `extractMid` falls back to the offer SDP: it looks for the m-line whose `a=msid` carries the track id, and if there's no match it returns the transceiver's index instead. After a failed negotiation is rolled back neither holds `a=msid` still names the track the sender was created with, not the clone `replaceTrack()` swapped in, and the browser's `mid` counter has moved past the transceiver index.

### 📝 Implementation notes

🎫 Ticket: https://linear.app/stream/issue/REACT-1048/stale-mid-in-setpublisher-after-peerconnection-rollbacks

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved WebRTC negotiation error handling and rollback behavior during retries.
  * Ensured media track announcements use the browser-assigned identifier when negotiation is retried.
  * Enhanced recovery when negotiation fails or when no media tracks are detected, preventing stale negotiation state.

### DIFF
--- a/packages/client/src/rtc/Publisher.ts
+++ b/packages/client/src/rtc/Publisher.ts
@@ -467,9 +467,10 @@ export class Publisher extends BasePeerConnection {
         // doesn't match the m-section carrying the track stops the SFU from
         // correlating the two: it then answers with every offered codec instead
         // of the announced one, and we publish the first codec of that list.
-        // See https://www.w3.org/TR/webrtc/#set-description - applying a local
-        // offer sets `transceiver.[[Mid]]` (step 10.1.2), and a rollback resets
-        // it to null for a transceiver that wasn't associated yet (step 11.2.1).
+        // Per https://www.w3.org/TR/webrtc/#set-description, applying a local
+        // offer associates the transceiver with its m-section and sets
+        // `[[Mid]]` to `[[JsepMid]]`; a rollback disassociates a transceiver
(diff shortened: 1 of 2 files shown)


packages/client/src/rtc/Publisher.ts
```diff
@@ -456,12 +456,12 @@ export class Publisher extends BasePeerConnection {
   private negotiate = async (options?: RTCOfferOptions): Promise<void> => {
     return withoutConcurrency(`publisher.negotiate.${this.lock}`, async () => {
       const offer = await this.pc.createOffer(options);
-      const tracks = this.getAnnouncedTracks(offer.sdp);
-      if (!tracks.length) throw new Error(`Can't negotiate without any tracks`);
-
       try {
         this.isIceRestarting = options?.iceRestart ?? false;
         await this.pc.setLocalDescription(offer);
+        const tracks = this.getAnnouncedTracks(offer.sdp);
```
oliverlaz, line 474: Good to have a comment here on why it is important to compute `tracks` after setting the local description.

oliverlaz (approved):

PR #2379, opened 2026-08-19
fix(client): cancel in-flight joins when the user leaves

santhoshvai:
### 💡 Overview

`leave()` now cancels a join completely: no further retries, no coordinator response applied, no peer connections or reporters created, and reconnect callers skip post-join work on a left call.

### 📝 Implementation notes

Reuses the existing supersededByLeave pattern.

Now also checked in the retry loop and at each await in doJoin, and surfaced as a 'joined' | 'superseded' result for reconnect callers.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved call lifecycle handling when leaving during an ongoing join or reconnection.
  - Prevented stale connection attempts from restoring call state or creating unnecessary media connections.
  - Ensured resources are released after cancelled joins, migrations, and reconnections.
  - Improved cancellation during retries and peer-connection setup.
  - Ensured calls remain correctly marked as left after a join is superseded.
  - Improved lifecycle reporting by removing outdated session identifiers.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -1182,6 +1182,12 @@ export class Call {
               delete joinData.migrating_from_list;
               return;
             } catch (err) {
+              // the user left mid-attempt: the failure is a consequence of that,
+              // so resolve quietly instead of rejecting `join()`
+              if (supersededByLeave()) {
+                this.logger.debug('Join superseded by leave; discarding error');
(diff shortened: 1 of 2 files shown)


santhoshvai:
closing this PR in place of a different implementation planned for soon

the scope should be reduced and a simpler change will be done in a follow up

Issue #2144, opened 2026-02-27
Calling call.get() seems to make me join the call


Jorundur:
**Which package/packages do you use?**

- [x] `@stream-io/video-react-sdk`
- [ ] `@stream-io/video-react-native-sdk`
- [ ] `@stream-io/video-client`

**Describe the bug**
I have a messenger-like application, and I've implemented the ringing call flow.

In a chat channel (1-1 or group), there's a special "CallMessage" I've implemented each time a user calls. That CallMessage component is supposed to track the current state of the relevant call (for showing information such as who is currently in the call, etc.).

This is my component

```tsx
const LatestCallMessageWithFeatureFlag = (
  props: Omit<LatestCallMessageCoreProps, "call">,
) => {
  const { client: videoClient } = useVideoClient();
  const [call, setCall] = useState<Call>();

  const attachment = props.message.attachments?.[0];
  const callId = attachment?.type === "call" ? attachment.call_id : undefined;

  useEffect(() => {
    if (!videoClient || !callId) {
      return;
    }

    const callInstance = videoClient.call("default", callId);
    callInstance
      .get()
      .then(() => setCall(callInstance))
      .catch((error) => {
        console.error("Failed to get call instance:", error);
      });

    return () => {
      callInstance.leave().catch((error) => {
        console.error("Failed to clean up call instance", error);
      });
      setCall(undefined);
    };
  }, [videoClient, callId]);

  return <LatestCallMessageCore {...props} call={call} />;
};
```

Simply by having this useEffect (even when no calls ongoing! i.e. this call is old), I get the video/audio activated as if I'm joining the call and the console renders a bunch of these:

```
{
    "type": "mic.capture_report",
    "call_cid": "default:the-call-uuid...",
    "capturesAudio": false,
    "deviceId": "71b9222c4468f59e1....",
    "label": "My Bose NC 700 (Bluetooth)"
}
```

If I comment out this useEffect, this does not happen

**To Reproduce**
See above

**Expected behavior**
I expected to be able to do call.get() to get latest events for the call.

**Screenshots**

**Version**

- SDK version 1.33.0

**Env:**

Fill out what's applicable

**React / JS client**

- Chrome

Jorundur:
Hi @oliverlaz — following up, because I think there's still a similar issue with `queryCalls`.

[...]

PR #2382, opened 2026-08-20
fix(client): keep devices disabled when querying calls

oliverlaz:
### 💡 Overview

Prevent `queryCalls()` from activating a camera or microphone while hydrating calls for read-only listings.

### 📝 Implementation notes

- default queried calls to disabled camera and microphone devices, including when an empty options object is provided
- prevent persisted enabled device preferences from overriding that default
- add regression coverage for query options and persisted preferences

🎫 Ticket: https://github.com/GetStream/stream-video-js/issues/2144

📑 Docs: N/A

## Summary by CodeRabbit

* **New Features**
  * Added optional control over whether camera and microphone devices are disabled when calls are queried or configured.
  * Device selection preferences are preserved without automatically re-enabling disabled devices.
  * Speaker handling now uses platform-appropriate defaults when no override is provided.

* **Bug Fixes**
  * Improved consistency when applying device settings across call setup and persisted preferences.

* **Tests**
  * Added coverage for disabled-device behavior and preference preservation.

### DIFF
--- a/packages/client/src/Call.ts
+++ b/packages/client/src/Call.ts
@@ -936,11 +936,7 @@ export class Call {
     // const calls = useCalls().filter((c) => c.ringing);
     const calls = this.clientStore.calls.filter((c) => c.cid !== this.cid);
     this.clientStore.setCalls([this, ...calls]);
-    const skipSpeakerApply = isReactNative();
-    await this.applyDeviceConfig(settings, {
-      publish: false,
-      skipSpeakerApply,
-    });
+    await this.applyDeviceConfig(settings, { publish: false });
   };
 
   /**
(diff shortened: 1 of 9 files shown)


packages/client/src/StreamVideoClient.ts
```diff
@@ -471,7 +475,10 @@ export class StreamVideoClient {
         clientStore: this.writeableStateStore,
       });
       call.state.updateFromCallResponse(c.call);
-      await call.applyDeviceConfig(c.call.settings, false, isReactNative());
+      await call.applyDeviceConfig(c.call.settings, {
+        publish: false,
+        withDisabledDevices,
+      });
```
oliverlaz, line 481: It starts here :)

jdimovska (approved):

packages/client/src/Call.ts
```diff
@@ -3372,20 +3362,29 @@ export class Call {
...
-    skipSpeakerApply: boolean,
+    opts: {
+      publish: boolean;
+      skipSpeakerApply?: boolean;
+      withDisabledDevices?: boolean;
+    },
   ) => {
+    const {
+      publish,
+      // Skip speaker setup on RN if ringing was requested or the call is already ringing
```
santhoshvai:
```suggestion
 // Skip speaker setup on RN until joining (explicitly passed as false during join())
```

santhoshvai (approved):

PR #2391, opened 2026-08-26
chore: call service adjustments

greenfrvr:
### 💡 Overview

Small tightening in CallService. Behavior is not changed, just guarding some potential scenarios. 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when updating call displays, including clearer diagnostics when updates fail.
  * Prevented background call services from restarting unnecessarily when no active calls are present.
  * Improved handling of repeated call registrations and service restarts.
  * Prevented duplicate call notifications and reduced repeated notification alerts when call details update.

### DIFF
--- a/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
+++ b/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
@@ -281,11 +281,9 @@ class CallService : Service(), CallRepository.Listener {
             }
             ACTION_START_BACKGROUND_TASK -> {
                 startBackgroundTask(intent)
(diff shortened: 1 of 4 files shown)


packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
santhoshvai, line 1: also we should stopSelf() on null intent

greenfrvr: done

santhoshvai:
closing as its merged to https://github.com/GetStream/stream-video-js/pull/2394

PR #2394, opened 2026-08-27
fix(callingx): don't tear down concurrent calls on stop service

santhoshvai:
### 💡 Overview

ACTION_STOP_SERVICE unconditionally cancelled all notifications and called start-id blind stopSelf(), whose onDestroy disconnects every call. Its only caller is the createStreamVideoClient() failure path, so a ringing push landing mid-await got torn down.

also has changes from https://github.com/GetStream/stream-video-js/pull/2397 

### 📝 Implementation notes

The stop is now a request, gated on hasAnyCalls() || hasRegisteredCall() and stopSelfResult(startId) — each covers an ordering the other can't see. The push handler ends the abandoned call explicitly. Three tracked-id leak paths closed.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Android call service shutdown behavior to avoid interrupting active calls.
  - Ensured failed incoming call notifications are fully ended and cleaned up when setup fails.
  - Improved reliability after app reloads or runtime shutdowns.
  - Avoided unnecessary service startup when stopping an inactive service.
  - Improved notification tracking and foreground call handling.
  - Added error reporting for failed display updates.

- **Documentation**
  - Clarified that `stopService()` requests service shutdown and does not end individual calls.
  - Documented using `endCallWithReason()` to terminate specific calls.

### DIFF
--- a/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
+++ b/packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
@@ -245,10 +245,7 @@ class CallService : Service(), CallRepository.Listener {
 
         unregisterReceiver(optimisticNotificationReceiver)
 
-        if (isInForeground) {
-            stopForeground(STOP_FOREGROUND_REMOVE)
-            isInForeground = false
-        }
+        demoteForeground()
 
         notificationManager.cancelAllNotifications()
         notificationManager.stopRingtone()
@@ -295,10 +292,7 @@ class CallService : Service(), CallRepository.Listener {
                 processAction(intent)
             }
             ACTION_STOP_SERVICE -> {
-                if (isInForeground) {
-                    stopForeground(STOP_FOREGROUND_REMOVE)
-                    isInForeground = false
-                }
+                demoteForeground()
                 notificationManager.cancelAllNotifications()
                 notificationManager.stopRingtone()
                 stopSelf()
@@ -371,10 +365,7 @@ class CallService : Service(), CallRepository.Listener {
                             TAG,
                             "[service] onCallStateChanged[$callId]: No more calls, stopping service"
                     )
-                    if (isInForeground) {
-                        stopForeground(STOP_FOREGROUND_REMOVE)
-                        isInForeground = false
-                    }
+                    demoteForeground()
                     stopSelf()
                 }
             }
@@ -543,10 +534,7 @@ class CallService : Service(), CallRepository.Listener {
 
                 // Only stop foreground/service when no other calls remain
                 if (!callRepository.hasAnyCalls()) {
-                    if (isInForeground) {
-                        stopForeground(STOP_FOREGROUND_REMOVE)
-                        isInForeground = false
-                    }
+                    demoteForeground()
                     notificationManager.stopRingtone()
                     stopSelf()
                 }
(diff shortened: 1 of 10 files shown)


packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
```diff
@@ -578,6 +609,55 @@ class CallService : Service(), CallRepository.Listener {
...
+            debugLog(
+                    TAG,
+                    "[service] stopServiceIfIdle: Calls still present (tracked=${CallRegistrationStore.getTrackedCallIds()}), keeping service alive"
+            )
+            return
+        }
+
+        if (!stopSelfIfNoPendingStart(startId)) return
+
+        notificationManager.cancelAllNotifications()
```
greenfrvr: I assume those lines are redundant? if we stop service, we do the cleanup  in `onDestroy`

packages/react-native-sdk/src/utils/push/internal/android.ts
```diff
@@ -79,6 +79,25 @@ export const onRingNotificationReceived = async (
...
+  const abandonPush = async () => {
+    if (asForegroundService) {
+      finishBackgroundTask();
+    }
+    try {
+      await callingx.endCallWithReason(call_cid, 'error');
+    } catch (error) {
+      nativeLog(`Failed to end call ${call_cid}: ${error}`, 'error');
+    }
+    await callingx.stopService();
```
greenfrvr: maybe in this case stop service is redundant in general? by the time `stopService` is invoked here, call can still pass the `stopServiceIfIdle` check for being in repo or tracking store, so it will be ignored. but `endCallWithReason` invocation on the other hand will stop the service in the end (call state moves to unregistered -> if there are not calls in repo/store we do call `stopSelfIfNoPendingStart`). does it make sense?

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
santhoshvai: removed

packages/react-native-sdk/src/utils/push/internal/android.ts
santhoshvai: addressed in our slack huddle

packages/react-native-callingx/android/src/main/java/io/getstream/rn/callingx/CallService.kt
```diff
@@ -578,6 +598,51 @@ class CallService : Service(), CallRepository.Listener {
...
+     * it may only go down when nothing owns it. Two independent signals are checked because they
+     * cover opposite interleavings of "tear this call down" against "a new call is arriving":
+     * - [CallRegistrationStore.hasRegisteredCall] covers a call already tracked when we read state.
+     *   Tracking happens synchronously *before* `startForegroundService`, so it is set even while
+     *   the new call's intent is still queued and the repository is empty.
+     * - [stopSelfIfNoPendingStart] covers a call whose start reached ActivityManager *after* we
+     *   read state, which the check above cannot see.
+     */
+    private fun stopServiceIfIdle(startId: Int) {
+        if (callRepository.hasAnyCalls() || CallRegistrationStore.hasRegisteredCall()) {
```
greenfrvr: I still have some concerns on using `CallRegistrationStore.hasRegisteredCall()` it is global - it will be shared across all service instances and in case we have some stale id from previous service all subsequent services will behave incorrectly. `callRepository.hasAnyCalls()` is safer as repo's lifetime is matching service itself.

greenfrvr (approved):

PR #2400, opened 2026-08-31
feat(client)!: replace RxJS with @stream-io/state-store

oliverlaz:
### 💡 Overview

Replaces RxJS with [`@stream-io/state-store`](https://www.npmjs.com/package/@stream-io/state-store) across `video-client`, `video-react-bindings` and `video-react-native-sdk`. `rxjs` is no longer a dependency of any of them.

`CallState` becomes a single store, so a server payload lands as **one atomic update** instead of 15+ separate subject emissions. Measured over 1000 `call.updated` payloads, a subscriber reading a second field from inside its callback observed a half-applied payload **1000 times before and 0 times after**. That is the reason for the change; everything else is a consequence or a side effect.

Full consumer-facing write-up in `MIGRATION-rxjs-to-state-store.md`.

**This is a breaking major.** Opened as a draft — see Open questions below before this is ready for review.

### 📝 Implementation notes

**What replaces what.** `foo$` members are now `Subscribable<T>`: `subscribe()` keeps both the callback and observer forms, and the returned subscription is still callable *and* has `.unsubscribe()`. `.pipe()` is gone. Derived state (`localParticipant`, `remoteParticipants`, `pinnedParticipants`, `dominantSpeaker`, `hasOngoingScreenShare`, `ownCapabilities`, and a new `participantsBySessionId` index) is maintained eagerly by a store preprocessor guarded on input identity, so synchronous reads are field accesses rather than subscribe/read/teardown round-trips.

**Blast radius is narrower than it looks.** `video-react-sdk` needed **zero** changes. All 45 react-bindings hooks keep their exact signatures. Across `ts-quickstart`, `e2ee-demo` and `react-dogfood`, every consumer call site used `.subscribe()` and **none used `.pipe()`**.

**An honest note on the performance numbers.** Reads are ~290x faster and subscribe/unsubscribe 7x, but most of that is *not* attributable to the substrate. The old `RxUtils.getCurrentValue()` built a `combineLatest`, subscribed, took one value and tore it down on **every** synchronous read (869 ns vs ~3 ns) — a bug that was equally fixable under RxJS. Same for the `participantsBySessionId` index (2.1x at 100 tiles, 4.6x at 400). `claude/rxjs-perf-optimizations` demonstrates this: on the current RxJS codebase, with no public API change, it reaches ~290x reads, 8.5x subscribe/unsubscribe (better than this branch) and 1.7-1.8x element binding — but **cannot** fix torn reads, and gets no bundle reduction. Reviewers deciding whether the breaking change is warranted should weigh that branch, not just this one.

**What only this branch delivers:** atomicity (torn reads 1000 -> 0), removal of the runtime dependency, cross-SDK alignment with Chat and Feeds, tear-free bindings on `useSyncExternalStore`, and -6.9 kB gzip in an application bundle (the client's own `dist` grows ~7 kB gzipped; the app still comes out ahead by not shipping RxJS).

**Writes cost more, and one row deserves attention.** Scalar writes are 3-5x slower in ratio and 60-75 ns in absolute terms — 0.007 ms of CPU per second at 100 writes/s, i.e. noise. Participant churn is the real cost and it scales with the call: parity at ~10 participants, 1.1x slower at 50, 1.9x at 400 (21.9 µs per write, ~2.2 ms CPU/s at 100 writes/s, about 0.22% of one core). Worth watching if very large calls become a target. The derivation is a single pass over the roster rather than six, which recovered roughly 20% of it.

**Behaviour changes worth reading:** `hasBrowserPermission` is now `false` until the permission state is known rather than not emitting at all; `setSortParticipantsBy` re-renders immediately instead of on the next participant change; reading `callStatsReport` off the store (rather than via `callStatsReport$` or `useCallStatsReport`) now requires an explicit `call.state.observeCallStatsReport()` registration, because a store subscription cannot declare which fields it reads.

### ✅ Verification

- `yarn lint:ci:all` clean
- `packages/client`: 78 files / 1102 tests pass
- `yarn test:react-native:sdk`: 29 + 97 pass
- `NODE_ENV=production yarn build:all` green
- `Call.test.ts`, `StreamVideoClient.test.ts` and `StreamVideoClient.ringing.test.ts` require `STREAM_API_KEY`/`STREAM_SECRET` and were **not run**
- **No end-to-end verification on real media yet** (ts-quickstart, react-dogfood, RN dogfood). This is the main gap.

### ❓ Open questions before review

- End-to-end testing on real media is outstanding — headless has no media, so this needs a manual pass on the sample apps.
- The benchmark harness backing every number here is deliberately **not** in this PR; it lands separately.
- Worth an explicit decision: is the atomicity + dependency removal worth a breaking major, given `claude/rxjs-perf-optimizations` captures most of the performance win without one?

🎫 Ticket: _none — spike/experiment_

📑 Docs: _pending; `MIGRATION-rxjs-to-state-store.md` in this PR is the draft migration guide_


oliverlaz:
Not feasible at this moment. The impact is quite large, with not extra benefits, and it requires thorough testing that nothing regressed.

PR #2418, opened 2026-09-07
perf(client): one declaration site per CallState field, faster state reads

oliverlaz:
### 💡 Overview

`CallState` declared every piece of state three times: the `BehaviorSubject`, a bare `x$: Observable<T>` placeholder carrying the JSDoc, and an assignment in the constructor. Understanding one field meant jumping between three regions ~200 lines apart, and adding one meant touching all three. This collapses it to one subject/observable pair per field, then takes the two performance wins that fall out of it.

### 📝 Implementation notes

- Three module-local helpers keep each declaration to a single line. `subject()` is overloaded so the seeded and empty cases stay honest at the type level: `subject<T>(initial)` yields `BehaviorSubject<T>`, `subject<T>()` yields `BehaviorSubject<T | undefined>`. `duc()` and `shared()` replace the constructor-local `distinctUntilChanged` closure and the six hand-spelled `shareReplay({ bufferSize: 1, refCount: true })` calls.
- Every observable is now a field initializer, which leaves the constructor holding only the event handler map (215 lines -> 98). The map has to stay in the constructor: it references private arrow-function fields declared further down, so as a field initializer it would capture `undefined` and those events would silently stop updating state.
- The 18 `undefined` entries in `eventHandlers` existed purely as a compile-time guard. They are replaced by an `UnhandledEventType` union in the new `src/store/types.ts`, enforced with `satisfies CallStateEventHandlers`. This keeps the "a newly introduced event fails to compile until it is triaged" guarantee and adds two more: a stale name in the ignore list, and an event that appears in both lists, now also fail to compile. All three are verified by perturbing the source, not assumed.
- `RxUtils.getCurrentValue` no longer wraps its argument in `combineLatest`. That wrapper landed in #1095 to replace `pipe(take(1))`, whose problem was resolving with the *first* rather than the *last* synchronous value; a bare `subscribe` already has that property. It now also reads a `BehaviorSubject`'s value directly instead of subscribing.
- The 27 getters backed by a plain `asObservable()`/`duc()` pair read their subject so they take that fast path. The eight backed by transforming pipelines (`participants`, `rawParticipants`, `localParticipant`, `remoteParticipants`, `dominantSpeaker`, `pinnedParticipants`, `hasOngoingScreenShare`, `ownCapabilities`) still read their observable, since those pipelines change the value.
- `call.state.getCurrentValue()` and `call.state.setCurrentValue()` are removed in favour of calling `RxUtils` directly, which is what `stateStore`, `DeviceManagerState`, `SpeakerState` and `Call.ts` already do.

### 📈 Benchmarks

Measured with the existing `bench/` harness against this branch's merge base, median of 7, three full runs agreeing within ~2%.

| exercise | before | after | result |
| --- | --- | --- | --- |
| read a scalar field (`state.callingState`) | 248 ns | 12 ns | **20.2x faster** |
| six setters, one of which reads a getter | 366 ns | 128 ns | **2.9x faster** |
| `updateFromCallResponse` | 900 ns | 633 ns | **1.42x faster** |
| participant patch, 10-participant call | 1.1 µs | 791 ns | 1.36x faster |
| participant patch, 50-participant call | 2.2 µs | 1.8 µs | 1.21x faster |
| read `state.participants` | 919 ns | 754 ns | 1.17x faster |
| `updateFromSfuCallState`, fan-out, subscribe churn | | | flat |
| bundle, minified / gzip | 456.6 kB / 133.2 kB | 453.7 kB / 132.8 kB | -2.9 kB / -0.3 kB |

Two caveats measured rather than waved away. Scalar writes look ~1.1x slower in the harness, but those totals are sub-millisecond and therefore display granularity; at 2M writes across 9 interleaved rounds the real delta is 14.31 -> 14.47 ns, i.e. flat. Heap per `CallState` instance is genuinely ~250 bytes (~1%) larger; the harness reports +8%, but that number is cold-start attribution, and repeating it with the order swapped gives a stable ~0.9%.

### ⚠️ Breaking change

`call.state.getCurrentValue()` and `call.state.setCurrentValue()` are removed. Use `RxUtils.getCurrentValue()` and `RxUtils.setCurrentValue()`, both already exported from `@stream-io/video-client`. Nothing in this repo outside `CallState` itself used them and no docs reference them.

One further intentional behavior change: `RxUtils.getCurrentValue` on a **completed** `BehaviorSubject` now returns its last value instead of `undefined`, because it reads the held value rather than subscribing (RxJS does not replay to subscribers of a stopped subject). Nothing in the SDK completes a `BehaviorSubject` (the only `.complete()` in the monorepo is `IceTrickleBuffer.live`, a plain `Subject`), and there is a test pinning the new behavior. Every other shape was verified identical: `ReplaySubject` at buffer size 1 and 2, finite `windowTime`, completed and errored subjects, plain `Subject`, `AsyncSubject`, `of(1,2,3)`, `EMPTY` and `throwError`.

### ✅ Verification

`yarn lint:ci:all`, `yarn test:ci:client` (82 files, 1168 passed, 1 skipped), `yarn test:react-native:sdk` (22 suites, 130 passed), `yarn build:client`. `react-bindings`, `react-sdk` and `react-native-sdk` all typecheck against the rebuilt client. The generated `CallState.d.ts` is unchanged apart from the two removed members, so all 35 observables keep their exact public types.

🎫 Ticket: https://linear.app/stream/issue/REACT-1106/callstate-clientstate

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved retrieval of current observable values across completed, errored, unsubscribed, and multi-emission states.
  - Improved consistency when reading screen-sharing settings and call state values.

- **Refactor**
  - Streamlined call-state initialization and observable handling.
  - Improved change detection for capability updates.
  - Added stronger validation for video event handling to improve reliability.

### DIFF
--- a/packages/client/src/devices/__tests__/ScreenShareManager.test.ts
+++ b/packages/client/src/devices/__tests__/ScreenShareManager.test.ts
@@ -99,7 +99,7 @@ describe('ScreenShareManager', () => {
 
   it('should use call settings to set up constraints', async () => {
     const call = manager['call'];
-    call.state.setCurrentValue(call.state['settingsSubject'], {
+    RxUtils.setCurrentValue(call.state['settingsSubject'], {
       // @ts-expect-error partial data
       screensharing: {
         target_resolution: {
--- a/packages/client/src/store/CallState.ts
+++ b/packages/client/src/store/CallState.ts
@@ -8,8 +8,13 @@ import {
   shareReplay,
   startWith,
 } from 'rxjs';
-import type { Patch } from './rxUtils';
-import * as RxUtils from './rxUtils';
+import {
+  getCurrentValue,
+  isShallowArrayEqual,
+  type Patch,
+  setCurrentValue,
+  updateValue,
+} from './rxUtils';
 import { CallingState } from './CallingState';
 import {
   type CallRecordingType,
@@ -59,10 +64,9 @@ import { Comparator, defaultSortPreset } from '../sorting';
 import { ensureExhausted } from '../helpers/ensureExhausted';
 import { hasScreenShare } from '../helpers/participantUtils';
 import { videoLoggerSystem } from '../logger';
+import type { AllEventHandlers, CallStateEventHandlers } from './types';
 
-/**
- * Returns the default egress object - when no egress data is available.
- */
+/** Returns the default egress object - when no egress data is available. */
 const defaultEgress: EgressResponse = {
   broadcasting: false,
   hls: { playlist_url: '', status: '' },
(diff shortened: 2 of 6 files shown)


oliverlaz:
Superseded by #2419 - same commit, branch renamed to `react-1106-callstate-declarations`. Renaming the head branch closed this PR rather than repointing it.

PR #2452, opened 2026-09-21
fix(react-native-callingx): use sip scheme for Telecom call handles

Sujeet76:
### 💡 Overview

ColorOS/Realme/Oppo can reboot the whole device when CallingX registers a Telecom call whose address is a Stream user id.

CallingX currently does `streamUserId.toUri()`. Opaque ids parse with `scheme = null`. ColorOS then NPEs in `PhoneNumberUtilsExtImpl.getNumberFromIntent` and `UserCallIntentProcessor.blockAndLaunchSystemDialer` crashes `system_server`.

AOSP `PhoneNumberUtils.getNumberFromIntent` already handles `tel` and `sip`. Core-Telecom documents `sip:` for VoIP handles that are not PSTN numbers.

### 📝 Implementation notes

- New `toTelecomAddress()`: empty → `sip:unknown`; keep an existing scheme; otherwise `Uri.fromParts("sip", handle, null)`.
- Used for incoming extras in `CallService` and outgoing start/update extras in `CallingxModuleImpl`.
- Do not use `tel:` — OEM dialer may try to place a real call.

Reproduced on Realme/ColorOS with `enableOngoingCalls: true`. Samsung unaffected.

🎫 Ticket

📑 Docs: noneandle scheme only


santhoshvai:
Hey @Sujeet76 thanks for  your PR and this bug report on ColorOS issue. I have cherry picked your commit to #2453 and adjusted it to using package name as the scheme instead of using SIP. But I have preserved your commit history by cherry picking. 

Closing this PR in favor of #2453

PR #2454, opened 2026-09-21
feat(client): integrate OpenAPI generated client

jdimovska:
### 💡 Overview

Replaces the hand-written coordinator calls with a client generated from the v2 OpenAPI spec, using the in-house `chat-manager` generator — the same one stream-chat-js and stream-feeds-js use.

Before, every endpoint was a hand-written `streamClient.post('/call/...', body)`. Now `Call` and `StreamVideoClient` each hold a generated API object, and the methods you already use forward to it:

```ts
call.goLive()            // curated, unchanged
call.api.goLive()        // the raw generated operation
client.api.queryCalls()  // anything not promoted onto the client
```

`Call` and `StreamVideoClient` *hold* the generated client rather than extending it, so their public surface stays what it was — same names, same signatures, same bound methods — while everything the spec offers is still reachable through`.api`.

### 📝 Implementation notes

**The spec is generated locally, video-only.** `generate-openapi.sh` builds it from a `chat` checkout with `-products video` instead of reading the published `video-clientside-api.yaml`. Including `common` would add 45 operations (polls,uploads, block lists, user groups…) and 106 models for the three the SDK uses.`VideoApi` ends up at 59 operations rather than 103. The three `common` routes we do need — devices, guest, WebSocket auth — are hand-written, with their types in `src/gen/shims.ts` (84 lines).

**Dates are now numbers.** `response_dates_as_number` types every server-sent date as the unix-nanosecond number the wire already carries, which removes the whole decoder layer. `new Date(ns)` is out of range and silently yields `Invalid Date`, so `src/helpers/time.ts` provides the conversions — `convertTimestampToDate`, `nsToDate`, `nowNs`. Request dates stay `Date`; `JSON.stringify` emits the RFC3339 the server expects.

**Bundle:** +1.76% minified, +1.16% minified+gzip. Under the 2% bar in AGENTS.md.

### ⚠️ Breaking changes

- **Base URL loses its `/video` suffix.** The generated paths are absolute (`/api/v2/video/...`), so an explicit `baseURL` ending in `/video` now produces  `/video/api/v2/video/...`. Drop the suffix.
- **Response dates are `number` (unix nanoseconds), not `string`.** Anything doing `new Date(call.created_at)` must use `convertTimestampToDate`. This is silent —  it yields `Invalid Date`, not an error.
- **42 `*Enum` exports removed** (`NoiseCancellationSettingsModeEnum`, `RequestPermissionRequestPermissionsEnum`, …). The fields are literal unions now,  so `mode: 'auto-on'` replaces `mode: NoiseCancellationSettingsModeEnum.AUTO_ON`.
- **Responses carry a `metadata` envelope** — `StreamResponse<T>` adds rate-limit and request-id info alongside the body.
- **`Call.getCallStats()` removed.** It was deprecated in favour of `call.getCallReport()` and called by nothing.
- **Dropped by the v2 spec:** `goLive`'s `notify` option, `listRecordings`'  session argument, and `getDevices` / `removeDevice`'s server-side `userID`.
- **`streamClient.doAxiosRequest` returns `AxiosResponse<T>`** rather than the body, so the metadata envelope can be built from the headers.

🎫 Ticket: https://linear.app/stream/issue/REACT-1092/openapi-v2-migration

📑 Docs: https://github.com/GetStream/docs-content/pull/1614

## Summary by CodeRabbit

* **New Features**
  * Added coordinator response metadata, including request IDs, status, and rate-limit details.
  * Added time conversion utilities and expanded exported client and response types.

* **Improvements**
  * Updated requests to API v2 endpoints and improved query serialization.
  * Standardized timestamps as Unix nanoseconds with safer invalid-date handling.
  * Improved call operations and recording retrieval.

* **Breaking Changes**
  * Updated device, call-stat, and permission method signatures.
  * Replaced several generated enum types with string-based values.
  * Removed deprecated recording query methods and the alternate development API generation workflow.


packages/client/src/coordinator/connection/api-client.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.streamClient = streamClient;
+  }
+
+  sendRequest = async <T>(
+    method: HttpMethod,
+    pathTemplate: string,
+    pathParams?: Record<string, string>,
+    queryParams?: Record<string, unknown>,
+    body?: unknown,
+    contentType?: string,
```
oliverlaz: It would be best to make `application/json` the default value so we don't have to send it as an argument everywhere.

packages/client/src/Call.ts
```diff
@@ -3278,72 +2977,12 @@ export class Call {
...
-  deleteTranscription = async (
-    callSessionId: string,
-    filename: string,
-  ): Promise<DeleteTranscriptionResponse> => {
-    return this.streamClient.delete<DeleteTranscriptionResponse>(
-      `${this.streamClientBasePath}/${encodeURIComponent(
-        callSessionId,
-      )}/transcriptions/${encodeURIComponent(filename)}`,
-    );
+  queryRecordings = async () => {
```
oliverlaz: I think we can remove the legacy `query*` API, as `list*` is what our openAPI and server-side SDKs define.

packages/client/src/gen/coordinator/video/VideoApi.ts
```diff
@@ -0,0 +1,926 @@
...
+  UpdateCallMembersRequest,
+  UpdateCallMembersResponse,
+  UpdateCallRequest,
+  UpdateCallResponse,
+  UpdateUserPermissionsRequest,
+  UpdateUserPermissionsResponse,
+} from '../models';
+
+export class VideoApi {
+  constructor(public readonly apiClient: ApiClient) {}
```
oliverlaz, line 103: Since we create a new `videoApi` instance for every `call` instance, maybe we can optimize the `'/api/v2/video/call/{type}/{id}'` processing, as we don't have to replace the path variables on every invocation. Also, we'll save some bytes on the bundle size. Does this need adjustments in the generator?

